PR #16, opened 2019-09-25
added auth

### LINKED ISSUE #13: create auth routes
Should include middleware and routes for registration/login. Optionally password reset (requires mail module).

sumeet-bansal:
Closes #13. Includes basic API routes (registration and login) and middleware.


app/api/v1/auth/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+const router = express.Router();
+
+const TOKEN_EXPIRES = 86400; // 1 day in seconds
+
+/**
+ * Middleware function that determines if a user is authenticated and assigns req.user to their user info from the db.
+ * Auth header should be included in the 'Authorization' request header in the form of 'Bearer <TOKEN>'.
+ */
+const authenticated = (req, res, next) => {
```
jadenpadua: fix indentation here

app/api/v1/auth/index.js
```diff
@@ -0,0 +1,90 @@
...
+      req.user = user;
+    }).then(next).catch(next);
+  });
+};
+
+/**
+ * Login route. POST body should be in the format of { email, password } and returns an auth token on success.
+ */
+router.post('/login', (req, res, next) => {
+  if (!req.body.email || req.body.email.length < 1) return next(new error.BadRequest('Email must be provided'));
```
jadenpadua: is this to account for cases where your email is just an empty '' string? If so should the same logic also apply to the User in the following functions?

app/api/v1/auth/index.js
sumeet-bansal: indentation is correct?

app/api/v1/auth/index.js
sumeet-bansal: Not sure I understand. The route accepts an email and password, and each must be some non-empty string. Then, [`User::findByEmail`](https://github.com/acmucsd/membership-portal/blob/master/app/db/models/user.js#L182) queries the database and retrieves the `User` object attached to the given email. Then if the user is in good standing, [`User.prototype::verifyPassword`](https://github.com/acmucsd/membership-portal/blob/master/app/db/models/user.js#L241) checks if the given password matches the hash associated with that `User` object. Then, a JWT will be signed by the app and returned in the response JSON. We validate every piece of data before writing it to the database, so therefore we know that the `User` object we retrieve from the database is well-formed.

PR #20, opened 2019-09-27
finished API routes

sumeet-bansal:
Includes `attendance`, `event`, and `user` routes.


sumeet-bansal:
> wait can you enable cors on development

will do

PR #24, opened 2019-09-28
Fixed CORS to allow OPTIONS

ronakdev:
Closes issue where preflight requests could not be made from frontend properly


sumeet-bansal:
This is a pretty small change so I went ahead and merged a PR without the `package-lock` stuff bc it seemed important for frontend development.

PR #69, opened 2020-03-22
Create API user/permissions for editing account type of users

### LINKED ISSUE #45: add an admin route for changing user permissions
Add an admin route (e.g. `/api/v1/user/permissions`) to change user permissions (e.g. `STANDARD` &rarr; `STAFF`).

nalinbhardwaj:
Modelled on user/bonus API.

Note that this allows users to edit their own access type as well (so it is potentially dangerous to use directly). Maybe we should add a check to disallow admins from being downgraded.

Closes #45

### DIFF
--- a/app/api/v1/user/index.js
+++ b/app/api/v1/user/index.js
@@ -164,4 +164,43 @@ router.route('/bonus')
     }).catch((err) => next(err));
   });
 
+router.route('/permissions')
+
+  /**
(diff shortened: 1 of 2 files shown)


sumeet-bansal:
Closing this because of the security issues mentioned in the PR body and because we're moving to a different permissions model soon.

PR #71, opened 2020-03-23
implemented email verification

### LINKED ISSUE #56: implement email verification
Users should have to verify their emails on account creation.

StoneT2000:
Closes #56 
Automatically sends an email to users to verify their email
Default access type is now RESTRICTED
RESTRICTED users are no longer returned in the leaderboard API

# TODO
Make sure RESTRICTED users cannot access store api
Rate limit new email routes here, which include `/register` and `/emailVerification`, latter of which is used to resend email verification emails.


app/api/v1/auth/index.js
```diff
@@ -82,16 +83,51 @@ router.post('/register', (req, res, next) => {
...
     newUser.hash = hash;
-    return User.create(newUser);
-  }).then((user) => {
-    log.info('user authentication (registration)', { request_id: req.id, user_uuid: user.uuid });
-    res.json({ error: null, user: user.getPublicProfile() });
-    Activity.accountCreated(user.uuid);
+    // require that we create a user succesfully and send email succesfully
+    db.transaction({
+      isolationLevel: Sequelize.Transaction.ISOLATION_LEVELS.REPEATABLE_READ,
+    }, (transaction) => Promise.all([
```
sumeet-bansal, line 94: This `Promise.all` only contains one promise element inside so we don't need it&mdash;we can just return the promise element instead.

app/api/v1/auth/index.js
```diff
@@ -82,16 +83,51 @@ router.post('/register', (req, res, next) => {
...
+      User.create(newUser).then((user) => {
+        User.generateAccessCode().then((code) => {
+          user.accessCode = code;
+          email.sendEmailVerification(user.email, user.firstName, code).then(() => {
+            // save the access code if email was sent
+            user.save();
```
sumeet-bansal, line 100: This writes to the same user twice, and I get the rationale behind saving the access code once the email's sent, but it doesn't change anything functionally if we do this double write and or if we create the user including the access code (i.e. User::generateAccessCode &rarr; User::create &rarr; Email::sendEmailVerification) besides internally writing an access code that gets overwritten once the user re-requests email verification. Also the `save` returns a Promise, so if that fails, the transaction rolls back the entire user object but still sends an email, so if we want to only commit the transaction when the email is sent, `sendEmailVerification` should be the last Promise in the chain.

Although the `save` isn't currently being returned so if it failed the transaction would still go through, minus the access code, and the email would still be sent, so the email itself would be invalid since there's no access code written to the user object.

Marginal, basically negligible difference so I think we might as well simplify where we can and do a single database operation.

sumeet-bansal, line 100: [`Model.prototype.save`](https://sequelize.org/master/class/lib/model.js~Model.html#instance-method-save) returns a Promise so make sure to handle that (or just write to the database once).

app/api/v1/auth/index.js
```diff
@@ -82,16 +83,51 @@ router.post('/register', (req, res, next) => {
...
+});
+
+/**
+ * Emails the user a email verification link
+ */
+// TODO: RATE LIMIT THIS
+router.post('/emailVerification', (req, res, next) => {
+  if (!req.body.email) return next(new error.BadRequest('Email must be provided!'));
+  User.findByEmail(req.body.email).then((user) => {
+    if (!user) return next(new error.BadRequest('Invalid email'));
```
sumeet-bansal, line 123: Throw errors inside Promises so the `catch` block passes it to `next` (I've run into some weird errors otherwise).

app/api/v1/auth/index.js
```diff
@@ -175,4 +212,25 @@ router.post('/verification', (req, res, next) => {
   });
 });
 
+/**
+ * Verifies a user's email based on a code sent out to the user's provided email
+ * Responds with a JSON object with fields 'error', 'verified' (if a user is succesfully verified)
+ */
+router.post('/verifyEmail', (req, res, next) => {
```
sumeet-bansal, line 219: Use the same `emailVerification` route and match password reset:

`GET /resetPassword/:email` + `POST /resetPassword/:accessCode` &rarr;
`GET /emailVerification/:email` + `POST /emailVerification/:accessCode`

app/api/v1/auth/index.js
```diff
@@ -175,4 +212,25 @@ router.post('/verification', (req, res, next) => {
...
+  if (!req.body.code || !req.body.email) {
+    return next(new error.BadRequest('Email and Email verification code must be provided'));
```
sumeet-bansal, line 221: why require email?

app/api/v1/auth/index.js
```diff
@@ -175,4 +212,25 @@ router.post('/verification', (req, res, next) => {
...
+  }
+  User.findByEmail(req.body.email.toLowerCase()).then(async (user) => {
+    if (!user) return res.json({ error: 'Invalid user email', verified: false });
```
sumeet-bansal, line 224: The error handler middleware takes care of status codes and error messages, so throw an error instead of responding.

app/api/v1/user/index.js
```diff
@@ -76,6 +76,13 @@ router.route('/')
     }
   });
 
+router.get('/:id', (req, res, next) => {
+  User.findByUUID(req.params.id).then((user) => {
+    if (req.params.id !== req.user.uuid) user = user.getPublicProfile();
+    res.json({ error: null, user });
+  }).catch(next);
+});
+
```
sumeet-bansal, line 85: what's this for?

app/db/models/user.js
```diff
@@ -42,13 +42,13 @@ module.exports = (Sequelize, db) => {
     },
 
     // account type
-    //   RESTRICTED - not used currently
+    //   RESTRICTED - not used currently / email not verified yet
```
sumeet-bansal, line 45: now it's used so you can replace "not used currently" with "an unverified user/member" or something

app/db/models/user.js
```diff
@@ -228,13 +243,16 @@ module.exports = (Sequelize, db) => {
       firstName: this.getDataValue('firstName'),
       lastName: this.getDataValue('lastName'),
       profilePicture: this.getDataValue('profilePicture'),
+      graduationYear: this.getDataValue('graduationYear'),
+      major: this.getDataValue('major'),
```
sumeet-bansal, line 247: I think this and the other route might be caused by merge conflicts since `master`'s been pushed a bunch since this branch was originally created.

app/email/index.js
```diff
@@ -27,4 +27,24 @@ const sendPasswordReset = (email, firstName, code) => {
...
+    }),
+  };
+  sendEmail(data).then(() => {
+    log.info(`Sent verification email to ${email}`);
+    resolve();
+  }).catch((error) => {
+    log.warn(`Failed to send email verification email to ${config.email.user}: ${error}`);
+    reject();
+  });
+});
```
sumeet-bansal, line 48: Instead of wrapping the whole thing in a Promise, can you return the `sendEmail` promise? And maybe throw an error in the `catch` block since the database transaction depends on it.

app/api/v1/auth/index.js
```diff
@@ -141,6 +177,8 @@ router.post('/resetPassword/:accessCode', (req, res, next) => {
     return User.generateHash(req.body.user.newPassword).then((hash) => {
       user.hash = hash;
       user.state = 'ACTIVE';
+      // as user reset password, it means their email must be valid
+      user.validateEmail();
```
sumeet-bansal, line 181: this and the next line are both Promises so handle them accordingly

app/api/v1/auth/index.js
```diff
@@ -152,8 +190,7 @@ router.post('/resetPassword/:accessCode', (req, res, next) => {
 
 /**
  * Verifies a user's auth token, given the token in the 'Authorization' request header in the form of 'Bearer <TOKEN>'.
- * Responds with a JSON object with fields 'error', 'authenticated' (if a user's token is valid), and 'admin' (if the
- * user is an admin).
+ * Responds with a JSON object with fields 'error', 'authenticated' (if a user's token is valid)
```
sumeet-bansal, line 193: can you check if this was a change you made or another merge thing?

app/db/models/user.js
```diff
@@ -228,13 +243,16 @@ module.exports = (Sequelize, db) => {
...
       points: this.getDataValue('points'),
     };
   };
 
   User.prototype.getUserProfile = function () {
     return {
       uuid: this.getDataValue('uuid'),
+      accountType: this.getDataValue('accessType'),
```
sumeet-bansal, line 255: same here with merge conflicts

app/api/v1/auth/index.js
StoneT2000: Makes a lot of sense actually, returning `next` inside a passed in function is kinda wierd because `next` is associated with the enclosing function

app/api/v1/auth/index.js
StoneT2000: wrote just once instead, applied same idea to error found in `emailVerification` route

app/api/v1/auth/index.js
StoneT2000: My change, the comment was old and we only return authenticated now anyway

app/db/models/user.js
StoneT2000: Merged master and it worked fine

app/db/models/user.js
StoneT2000: I added? Or you added? I don't see any merge conflicts with master, but this is needed so front end knows what access a user has

app/api/v1/auth/index.js
StoneT2000: did single

app/db/models/user.js
sumeet-bansal: looks like this is from `master` (see [here](https://github.com/acmucsd/membership-portal/blob/master/app/db/models/user.js#L240))

app/api/v1/auth/index.js
StoneT2000: Actually not sure, might have been originally me trying to use the email to generate the access code, I'll switch over to the findAcessCode method instead

app/api/v1/user/index.js
StoneT2000: old master

app/api/v1/auth/index.js
```diff
@@ -82,16 +83,69 @@ router.post('/register', (req, res, next) => {
...
+// TODO: RATE LIMIT THIS
+router.get('/emailVerification/:email', (req, res, next) => {
+  if (!req.params.email) return next(new error.BadRequest('Email must be provided!'));
+  User.findByEmail(req.params.email).then((user) => {
+    if (!user) throw new error.BadRequest('Invalid email');
+    User.generateAccessCode().then(async (code) => {
+      user.accessCode = code;
+      // wait for user to save first before sending email
+      await user.save();
+      email.sendEmailVerification(user.email, user.firstName, code).then(() => {
```
sumeet-bansal, line 124: nit: the rest of the repo uses `then`/`catch` instead of `async`/`await`

app/api/v1/auth/index.js
```diff
@@ -82,16 +83,69 @@ router.post('/register', (req, res, next) => {
...
+ * Verifies a user's email based on a code sent out to the user's provided email
+ * Responds with a JSON object with fields 'error', 'verified' (if a user is succesfully verified)
+ */
+router.post('/emailVerification/:accessCode', (req, res, next) => {
+  if (!req.params.accessCode) {
+    return next(new error.BadRequest('Email verification code must be provided'));
+  }
+  User.findByAccessCode(req.params.accessCode).then(async (user) => {
+    if (!user) throw new error.BadRequest('Code is invalid');
+    if (user.accessCode === req.params.accessCode) {
```
sumeet-bansal, line 141: Isn't this implied by `User.findByAccessCode`? Like, `user` exists only if the access code is a match so this conditional isn't doing anything.

app/api/v1/auth/index.js
```diff
@@ -82,16 +83,69 @@ router.post('/register', (req, res, next) => {
...
+
+/**
+ * Verifies a user's email based on a code sent out to the user's provided email
+ * Responds with a JSON object with fields 'error', 'verified' (if a user is succesfully verified)
+ */
+router.post('/emailVerification/:accessCode', (req, res, next) => {
+  if (!req.params.accessCode) {
+    return next(new error.BadRequest('Email verification code must be provided'));
+  }
+  User.findByAccessCode(req.params.accessCode).then(async (user) => {
```
sumeet-bansal, line 139: async nit

app/api/v1/auth/index.js
```diff
@@ -138,9 +192,11 @@ router.post('/resetPassword/:accessCode', (req, res, next) => {
   User.findByAccessCode(req.params.accessCode).then((user) => {
     // if no such user was found, the access code must be invalid/non-existent
     if (!user || !user.requestedPasswordReset()) throw new error.BadRequest('Invalid access code');
-    return User.generateHash(req.body.user.newPassword).then((hash) => {
+    return User.generateHash(req.body.user.newPassword).then(async (hash) => {
       user.hash = hash;
       user.state = 'ACTIVE';
+      // as user reset password, it means their email must be valid
+      await user.validateEmail();
```
sumeet-bansal, line 199: async nit

app/db/models/user.js
sumeet-bansal, line 247: Can you manually delete these? These lines are already in `master` and idk what'd happen if they get merged in but I'd rather not find out.

app/db/models/user.js
sumeet-bansal, line 255: delete this too

app/db/models/user.js
```diff
@@ -42,13 +42,13 @@ module.exports = (Sequelize, db) => {
...
     // account type
-    //   RESTRICTED - not used currently
+    //   RESTRICTED - unverified user / member
     //   STANDARD   - a regular ACM member
     //   STAFF      - a Diamond Staff member
     //   ADMIN      - admin type user
     accessType: {
       type: Sequelize.ENUM('RESTRICTED', 'STANDARD', 'STAFF', 'ADMIN'),
-      defaultValue: 'STANDARD',
+      defaultValue: 'RESTRICTED',
```
sumeet-bansal, line 51: Just realized, you'll need to write a migration for this.

app/api/v1/auth/index.js
```diff
@@ -82,16 +83,69 @@ router.post('/register', (req, res, next) => {
...
   const newUser = User.sanitize(req.body.user);
   newUser.state = 'ACTIVE';
   User.generateHash(req.body.user.password).then((hash) => {
     newUser.hash = hash;
-    return User.create(newUser);
-  }).then((user) => {
-    log.info('user authentication (registration)', { request_id: req.id, user_uuid: user.uuid });
-    res.json({ error: null, user: user.getPublicProfile() });
-    Activity.accountCreated(user.uuid);
+    // require that we create a user succesfully and send email succesfully
```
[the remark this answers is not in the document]
StoneT2000: nice catch!

app/db/models/user.js
```diff
@@ -204,11 +204,26 @@ module.exports = (Sequelize, db) => {
     });
   };
 
+  // Verifies user's email by setting their account type to STANDARD if it is RESTRICTED
+  User.prototype.validateEmail = function () {
```
sumeet-bansal, line 208: update to set `state ACTIVE`

app/db/models/user.js
```diff
@@ -204,11 +204,26 @@ module.exports = (Sequelize, db) => {
...
+
   User.getLeaderboard = function (offset, limit) {
     if (!offset || offset < 0) offset = 0;
     if (!limit || limit < 0) limit = undefined;
     return this.findAll({
-      where: { accessType: { [Sequelize.Op.not]: 'ADMIN' } },
+      where: {
+        [Sequelize.Op.and]: [
+          { accessType: { [Sequelize.Op.not]: 'ADMIN' } },
+          { accessType: { [Sequelize.Op.not]: 'RESTRICTED' } },
```
sumeet-bansal, line 224: update to `state ACTIVE`

app/email/index.js
```diff
@@ -27,4 +27,18 @@ const sendPasswordReset = (email, firstName, code) => {
...
-module.exports = { sendPasswordReset };
+const emailVerificationTemplate = readTemplate('emailVerification.ejs');
+const sendEmailVerification = async (email, firstName, code) => {
+  const data = {
+    to: email,
+    from: config.email.user,
+    subject: 'ACM UCSD Membership Portal Email Verification',
+    html: ejs.render(emailVerificationTemplate, {
+      firstName,
+      link: `${config.client}/verifyEmail/${email}/${code}`,
```
sumeet-bansal, line 38: remove `email`

app/db/models/user.js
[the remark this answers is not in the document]
StoneT2000: using default STANDARD now

app/db/models/user.js
StoneT2000: now blocking for `accessType ADMIN`, `state PENDING` and `state BLOCKED`

PR #72, opened 2020-03-25
updated CORS to allow main website and subdomains

paulpan05:

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,7 @@ const server = express();
 
 // enable CORS
 server.use((req, res, next) => {
-  const regex = '.*acmucsd.com';
+  const regex = '.*\\.acmucsd\\.com';
   let allowedOrigin = 'https://acmucsd.com';
   if (app.config.isDevelopment) {
     allowedOrigin = '*';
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -9,7 +9,14 @@ const server = express();
 
 // enable CORS
 server.use((req, res, next) => {
-  res.header('Access-Control-Allow-Origin', app.config.isDevelopment ? '*' : 'https://members.acmucsd.com');
+  const regex = '.*acmucsd.com';
```
sumeet-bansal: should it be `*.acmucsd.com` (asterisk then dot) instead?

paulpan05: I'm using regex matching so .* matches any number of characters before "acmucsd.com"

sumeet-bansal: Couldn't that match a non-ACM domain like "fakeacmucsd.com"? Also, we need to escape the period since that's a special regex character. If we're restricting to sub-domains of "acmucsd.com" e.g. "cyber.acmucsd.com" then our regex should be something like `.*\.acmucsd\.com` (asterisk then escape the dot).

paulpan05: Got it, I changed it based on your suggestion and put in double backslashes since JavaScript ignores the first backslash.

PR #74, opened 2020-03-26
implemented the merch store

### LINKED ISSUE #53: implement the merch store
#### /api/v1/store/item
+ [user] GET all items
+ [admin] POST a new item
+ [admin] DELETE some item(s)
+ [admin] PATCH some item(s)

#### /api/v1/store/order
+ [user] GET past orders for the current user
+ [user] POST new order to order some item(s)
+ [admin] GET past orders and unfulfilled items for a given user
+ [admin] PATCH unfulfilled order items

#### MERCHANDISE
+ id: incrementing int
+ uuid: uuid
+ item_name: string
+ price: int
+ quantity: int
+ description: string
+ discountPercentage: int

#### ORDERS
+ id: incrementing id
+ uuid: uuid
+ user: uuid
+ total_cost: int
+ ordered_at: timestamp

#### ORDER_ITEMS
+ id: incrementing id
+ order: uuid
+ item: uuid
+ fulfilled: boolean
+ fulfilled_at: timestamp

#### Misc
+ split user points into total points and spendable credits (diamonds)
+ add activity type

sumeet-bansal:
Closes #53.


src/email/templates/orderConfirmation.ejs
```diff
@@ -0,0 +1,166 @@
...
+                                                        </tr>
+
+                                                        <% for(var i=0; i < order.items.length; i++) { %>
+                                                            <tr>
+                                                                <td><%= order.items[i].itemName %></td>
+                                                                <td><%= order.items[i].price %></td>
+                                                                <td><%= order.items[i].quantityRequested %></td>
+                                                                <td><%= order.items[i].total %></td>
+                                                            </tr>
+                                                        <% } %>
```
shravanhariharan2, line 128: The `price` and `total` don't reflect discounted items, they reflect the actual price of the items even though the discounted price is subtracted from the user's credits. I think you should probably display discounted price here rather than raw price

sumeet-bansal: Good catch, will do.

sumeet-bansal:
> If two or more errors are encountered, I think it would be useful for users to see all errors that they encountered, such as "error: monthly limit reached for x, lifetime limit reached for y, out of stock for z" instead of just listing "monthly limit reached for x" because of its priority. This way, users ordering multiple items can fix all their ordering issues at once instead of fixing each one individually and running into the next one when ordering again.

> Then again, I feel like it would be beneficial to warn the user of these errors before they even select the erroneous items and try to place their order (like in a shopping list in the FE, gray-out the "Add item" button if monthly/lifetime limits are reached). Although, this would probably result in a bunch of extra API calls to load all that info, so the current solution may be the most performant.

The FE could preload the user's purchase history (`GET /store/order`) and use that to determine if an order exceeds some monthly/lifetime limit, and the FE already has all the data it needs (credits, merch prices and discounts) to determine if the user has enough credits, so those are UX improvements that the FE can implement without any further changes to the store controller.

> Are discounts permanent/static? Or do they last a certain amount days and have to be manually updated by an admin once to start/end the discount? Since there was no timestamp attached to a Merchandise item, I'm assuming its static but I think it would be cool to have temporary discounts for items, perhaps for celebration of some event/activity (like the Camp Snoopy Snapback going on sale for 1 week to celebrate some Sixth College event)

Discounts can be updated, and I debated keeping track of price/discount changes or at what price an item gets bought, so `Order.totalCost` kind of logs that (and there could be an associated `OrderItem.cost`), but I didn't really put a lot of thought into it so I'm open to suggestions. Maybe `OrderItem.cost` and `OrderItem.discount`? So we can determine the original price, discount, and discounted price at which any item is purchased at.

> Is there any functional/monetary benefit of collections besides the fact that they are a group of similar items? I saw you mentioned perhaps having discounts over collections, and I think that would be super cool (like Hack School collection having all items 20% off for a week to celebrate the last week of Hack School).

In our initial store planning meetings, we wanted to do limited "drops" and collections provide a way to logically group together similar items. Collections as a whole can be discounted, archived, etc., so it's easier for admins to make changes to an entire group of merch items as well (although the PR doesn't currently have a way for collections to be discounted or archived&mdash;I'll add that). I'm thinking some collection-wide operations just write to all the collection's items instead of adding attributes to `Collection` itself (e.g. changing discount percentage), `Collections` gets an `archived` attribute, and we delete `GET /store/merchandise` since it basically does the same thing as `GET /store/collection` but without the collection info.

src/api/v1/store/index.js
```diff
@@ -0,0 +1,337 @@
...
+        const collection = await MerchandiseCollection.findByUUID(req.params.uuid).then((c) => (c ? c.get() : c));
+        if (!collection) return next(new error.NotFound('No such merchandise collection found'));
+        const merchandise = await Merchandise.findAllByCollection(collection.uuid, req.user.isAdmin());
+        collection.merchandise = req.user.isAdmin() ? merchandise : merchandise.map((m) => m.getPublicItem());
+        res.json({ error: null, collection });
+        return;
+      }
+      const findCollectionsMinusMerch = req.user.isAdmin()
+        ? MerchandiseCollection.getAllCollections() : MerchandiseCollection.getAllUnarchivedCollections();
+      const collectionsMinusMerch = (await findCollectionsMinusMerch).then((cs) => cs.map((c) => c.get()));
```
shravanhariharan2, line 23: Does this line cause an error for you? I'm getting a `TypeError: (intermediate value).then is not a function` on the `.then` when calling GET `/api/v1/store/collection/` as an admin (errors out as non-admin as well). I'm on the latest pull from this branch too.

sumeet-bansal: Fixed.

src/db/models/merchandise.js
```diff
@@ -0,0 +1,183 @@
...
+
+  Merchandise.findAllByCollection = function (collection, isAdmin) {
+    const where = isAdmin ? { collection } : { collection, hidden: false };
+    return this.findAll({ where });
+  };
+
+  Merchandise.getDiscountedItems = function (offset, limit) {
+    if (!offset || offset < 0) offset = 0;
+    if (!limit || limit < 0) limit = undefined;
+    return this.findAll({
```
steets250, line 129: Do you also need to check for `hidden: false` here for non-admin users?

src/api/v1/store/index.js
```diff
@@ -0,0 +1,337 @@
...
+  })
+
+  /**
+   * Creates an item, given a 'merchandise' object in the request body, and returns the created item
+   * upon success. Required fields: itemName, price, description. Optional fields: picture, quantity,
+   * discountPercentage, hidden. All other fields will be ignored.
+   */
+  .post(async (req, res, next) => {
+    if (!req.body.merchandise) return next(new error.BadRequest('Merchandise item must be provided'));
+
```
shravanhariharan2, line 124: POST `/store/merchandise/` throws an error if no collection uuid is provided or if collection is null in request body. Not all items need to belong to a collection though, right?

src/db/models/merchandise.js
sumeet-bansal: The check could be elsewhere, e.g. the API layer, but I don't think this function is being used so I'm just gonna delete it.

src/api/v1/store/index.js
sumeet-bansal: Items must belong to a collection; it makes implementation (and merch taxonomy) much simpler and design team signed off on it.

sumeet-bansal:
> Email confirmation: email still doesn't show the proper discounted price after ordering. When I buy a 2000 credit item for 20% off, 1600 credits gets taken away from my acc but email confirmation shows 2000. Also the email CSS is still a little funky but I remember you mentioning someone else was looking into that already.

Correct sale price gets passed into the email template now.

> `GET /order`: I'm getting an empty response even after ordering items. As both an admin and a user, I get no items at all even after I order items and receive confirmation.

Fixed.

> `POST /order`: This route doesn't work when ordering more than 1 item now. I get a BadRequest [400]: There are duplicate or invalid items in this order when sending an order with more than 1 item.

Fixed.

> `PATCH /order`: I'm not sure this route properly fulfills the item. Boolean value fulfilled still remains "f" for OrderItems table when looking at database after sending those uuids into the request. This is after ordering the items I mentioned above as separate orders.
> `PATCH /order`: Doesn't reject my error when I had one item "f" and one item "t" and sent both to fulfill. I didn't get an error, but nothing changed in the database regardless.

Works as intended. Each UUID has a couple fields to go with it: `fulfilled: boolean` and `notes: string`, so a distributor can add a note to an OrderItem without fulfilling it.

src/email/templates/orderConfirmation.ejs
```diff
@@ -0,0 +1,132 @@
...
+<% for(var i=0; i < order.items.length; i++) { %>
+                    <p><%= order.items[i].itemName %></p>
+                    <p><small><%= order.items[i].salePrice %> x <%= order.items[i].quantityRequested %> = <%= order.items[i].total %></small></p>
+<% } %>
+    <h1><%= order.items[i].totalCost %></h1>
+                                        <p style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-bottom: 15px;">To pick up your items, find a merch distribution event on the membership portal and stop by.</p>
+                                        <p style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-bottom: 15px;">Regards, <br />ACM UCSD</p>
+                                    </td>
+                                </tr>
+                            </table>
```
shravanhariharan2, line 105: Getting an error caused by these lines when ordering any item. The order still goes through but I don't get any confirmation email.
```
InternalServerError: ejs:100
    98|                     <p><small><%= order.items[i].salePrice %> x <%= order.items[i].quantityRequested %> = <%= order.items[i].total %></small></p>
    99| <% } %>
 >> 100|     <h1><%= order.items[i].totalCost %></h1>
    101|                                         <p style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-bottom: 15px;">To pick up your items, find a merch distribution event on the membership portal and stop by.</p>
    102|                                         <p style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-bottom: 15px;">Regards, <br />ACM UCSD</p>
    103|                                     </td>

Cannot read property 'totalCost' of undefined
```

Input request:
```
{
	"order": [
		{
			"item": "6b65bec1-4bff-4334-bf64-792b83ed426a",
			"quantity": 1
		}
	]
}
```
`order` response when console.logged:
```
{
  uuid: '1ea1fbee-7aa4-4257-b869-d3a7173e51ce',
  orderedAt: 2020-05-20T18:56:15.360Z,
  totalCost: 1275,
  items: [
    {
      itemName: 'Hack School Sticker Pack (4)',
      picture: null,
      description: 'Make space on your laptop cover for these 2 in x 2 in stickers. Pack of 4.',
      quantityRequested: 1,
      salePrice: 1275,
      total: 1275
    }
  ]
```

sumeet-bansal: Fixed typo.

sumeet-bansal:
Oops, must've deleted some HTML by accident. Added it back.

PR #79, opened 2020-04-08
fixed users router

### LINKED ISSUE #78: fix broken users routes
Some users routes (e.g. `/activity`) are currently broken because of the `/:uuid?` route being placed first. Originated from #73.

sumeet-bansal:
Fixes #78.

### DIFF
--- a/app/api/v1/user/index.js
+++ b/app/api/v1/user/index.js
@@ -5,83 +5,6 @@ const { User, Activity, db } = require('../../../db');
 
 const router = express.Router();
 
-router.route('/:uuid?')
-
-  /**
-   * Gets user profile for current user.
-   */
-  .get((req, res, next) => {
-    if (!req.params.uuid || req.params.uuid === req.user.uuid) {
-      res.json({ error: null, user: req.user.getUserProfile() });
-      return;
-    }
-    User.findByUUID(req.params.uuid).then((user) => {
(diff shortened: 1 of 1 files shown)


app/api/v1/user/index.js
```diff
@@ -163,4 +86,81 @@ router.route('/bonus')
...
+    if (validateStringField(req.body.user.firstName) && req.body.user.firstName !== req.user.firstName) {
+      updatedUser.firstName = req.body.user.firstName;
+    }
+    if (validateStringField(req.body.user.lastName) && req.body.user.lastName !== req.user.lastName) {
+      updatedUser.lastName = req.body.user.lastName;
+    }
+    if (validateStringField(req.body.user.major) && req.body.user.major !== req.user.major) {
+      updatedUser.major = req.body.user.major;
+    }
+    if (req.body.user.graduationYear && !Number.isNaN(parseInt(req.body.user.graduationYear, 10))) {
```
leestanley, line 126: Should we change the password requirements with some [regex checking](https://stackoverflow.com/questions/19605150/regex-for-password-must-contain-at-least-eight-characters-at-least-one-number-a)?

sumeet-bansal: Thanks for the link; definitely something we can look into in a future PR. I don't particularly care about enforcing password requirements like "must have a special character" but if the rest of the team thinks it's a good idea, I'm all for it.

Issue #67, opened 2020-03-20
fix all the database index types


sumeet-bansal:
All the database indexes are currently `BTREE` because the model definitions used the deprecated [`method`](https://sequelize.readthedocs.io/en/v3/docs/models-definition/#indexes) instead of [`using`](https://sequelize.org/master/manual/indexes.html); some indexes should be `HASH`.

This problem shouldn't have a significant effect since indexes are for fine-tuning database reads at scale, and we're not operating at that level (yet!), but still worth fixing and doing right.

sumeet-bansal:
Closed by #82.

PR #82, opened 2020-04-08
added initial database migrations

### LINKED ISSUE #50: create initial database migrations
Should migrate a blank database to the original schema (see PR #9) so developers can easily test migrations.

sumeet-bansal:
Closes #50. Adds migrations from the initial database PR (#9) and updates some indexes.

**If you're not me, don't merge this**: since we're adding these migrations but don't want to run them on the production database, we need to manually mark these migrations as run in the `SequelizeMeta` table (also `date` to `timestamp`).


sumeet-bansal:
Closed #67.

PR #89, opened 2020-05-03
Create user DELETE route

### LINKED ISSUE #70: Route for Admins to delete users
Let admin accounts delete specified users

StormFireFox1:
Closes #70.

There could be a bit of debate left as to what specifically a "user delete" entails. I assumed it would also mean deleting their related objects in the database, however we could go with a "DISABLED" state for "deleted" accounts (although that might require more checks in other routes).

### DIFF
--- a/src/api/v1/user/index.js
+++ b/src/api/v1/user/index.js
@@ -173,12 +173,21 @@ router.route('/:uuid?')
     }
   })
 
+  /**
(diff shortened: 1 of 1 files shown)


sumeet-bansal:
Closing this because user deletion requires further discussion (e.g. true deletion vs. anonymizing accounts to preserve historical data about event attendances).

PR #93, opened 2020-05-05
Add Grad Year and Major to Public Profile

steets250:
This fixes an issue where the frontend tries to display the user's graduation year and major but comes up blank since the fields are not returned by the backend.


steets250:
@sumeet-bansal thanks, will do. I don't currently have write access to the repo, so I can't merge this request.

PR #98, opened 2020-05-21
reconfigure s3 file uploads

shravanhariharan2:
This is just a sample to demonstrate how a more OOP file upload service would look like and to get feedback to see if we really want to change to this sort of service. Changes still need to be made but the core functionality works the exact same (I still need to do error handling/final touches).

Essentially, I removed `multer-s3` and reconfigured `multer` to pipe the file into the s3 storage service (`src/storage/index.js`) where the file is uploaded to s3. The only drawback is that the file _has_ to be written to local storage in order for _any other s3 upload library other than `multer-s3`_ to process its contents. `multer-s3` is the only s3 uploading library that does not require writing/deleting to disk since its already configured w/ `multer`. The file is promptly deleted from disk after being uploaded to s3, and rest of the functionality is the same.

I'd like to hear your thoughts on this trade-off of cleaner, more object-oriented code and potentially slower response time due to writing/deleting every file vs extensive middleware use w/ faster response times. I don't notice any substantial response time increase, but definitely something to consider if we plan on making this change.

### DIFF
--- a/src/api/v1/user/index.js
+++ b/src/api/v1/user/index.js
@@ -21,13 +21,12 @@ router.get('/activity', (req, res, next) => {
 router.post('/picture', (req, res, next) => {
   try {
     const fileUpload = Storage.getFileUpload('image', 256);
-    fileUpload(req, res, async (err) => {
+    fileUpload(req, res, (err) => {
       if (err) next(err);
       if (req.file) {
-        const { path, originalname } = req.file;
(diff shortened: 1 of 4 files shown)


shravanhariharan2:
@sumeet-bansal what do you think of these changes for uploads? This is essentially the full functionality and is adaptable w/ different sizes and diff file inputs like resumes, event covers, etc, and everything seems to work as intended

PR #100, opened 2020-05-30
Add event cover uploads

### LINKED ISSUE #90: add support for uploading event covers
Should be similar to the route for uploading profile pictures.

shravanhariharan2:
Closes #90 .

So I know we were trying to stray away from using middleware for managing files, but I really want to push this particular solution for uploads using `multer` as a middleware since it solves all the problems we have at almost no cost (old `multer-s3` uploads took ~250ms, this implementation takes ~300ms, writing to disk took ~600ms). So hear me out.

The biggest problem we have with this current s3 setup that I made is that the `req.body` would only populate AFTER the file has been processed by `multer`, and since `multer-s3` was connected to `multer`, it meant that the file had to be uploaded to s3 before populating `req.body`, meaning we can't error check anything until the file has already been uploaded. We need `req.body` populated so we can access the `uuid` for events and other potential upload keys through a `POST` request.

The solution is to decouple `multer`'s processing of files and bodies from uploading the file to s3. Since the we want to error check as the first step in our code (to keep consistent w/ all other routes), we can use `multer` as a middleware (passing in the appropriate file limits and file name) to immediately have access to `req.body` and `req.file` on the first line of the route. This makes the code look sooooo much cleaner than using `multer` as a function, and this implementation doesn't write the file to disk (it stores the file in memory as a buffer, and the GC frees it once the request is done).

In my opinion this is the most optimal implementation because the code is _way_ cleaner and we are able to error check requests w/o uploading first, and there is basically no difference in runtime from using `multer-s3` (50ms is 0.05s which is unnoticeable).


src/storage/index.js
```diff
@@ -13,22 +13,30 @@ const s3 = new aws.S3({
...
+  const coverUrl = await s3.upload(params).promise()
+    .then((data) => data.Location)
+    .catch((error) => {
+      log.warn(`Failed to upload profile picture to S3: ${error}`);
+    });
+
+  return coverUrl;
+};
+
+const getFile = (maxFileSize) => {
```
sumeet-bansal, line 32: 1. `getFile` is a misnomer since `multer` is middleware that processes the image blob and stores it on disk or in a memory buffer; rename to something more accurate like `bufferImageBlob` or something.
2. The `multer` object doesn't specify a [`storage`](https://www.npmjs.com/package/multer#multeropts) engine. In its current form, `getFileUpload` specifies a `multer-s3` object as its `storage` engine, so you'll want to use the [memory storage engine](https://www.npmjs.com/package/multer#memorystorage) here.

src/storage/index.js
```diff
@@ -13,22 +13,30 @@ const s3 = new aws.S3({
...
+    Bucket: config.s3.bucket,
+    Key: `portal/${folder}/${uuid}${path.extname(file.originalname)}`,
+  };
+  const coverUrl = await s3.upload(params).promise()
+    .then((data) => data.Location)
+    .catch((error) => {
+      log.warn(`Failed to upload profile picture to S3: ${error}`);
+    });
+
+  return coverUrl;
```
sumeet-bansal, line 29: return the promise and let the controller catch the error; otherwise, this is gonna catch the error, write to log, and the controller sends a successful response for a failed request

src/storage/media.js
```diff
@@ -0,0 +1,33 @@
...
+  let mediaTypeConfig;
+  switch (mediaType) {
+    case mediaTypes.EVENT_COVER:
+      mediaTypeConfig = {
+        type: mediaTypes.EVENT_COVER,
+        maxFileSize: config.file.MAX_EVENT_COVER_FILE_SIZE,
+        uploadPath: config.file.EVENT_COVER_UPLOAD_PATH,
+        fileTag: 'image',
+      };
+      break;
```
sumeet-bansal, line 18: Return instead of breaking so we can throw an error if the input doesn't get pattern matched.

src/storage/media.js
```diff
@@ -0,0 +1,33 @@
...
+    case mediaTypes.PROFILE_PICTURE:
+      mediaTypeConfig = {
+        type: mediaTypes.PROFILE_PICTURE,
+        maxFileSize: config.file.MAX_PROFILE_PICTURE_FILE_SIZE,
+        uploadPath: config.file.EVENT_COVER_UPLOAD_PATH,
+        fileTag: 'image',
+      };
+      break;
+    default:
+      mediaTypeConfig = {};
```
sumeet-bansal, line 28: Throw an error here.

This pattern matching should be exhaustive, and if it's not, that's something we want to point out immediately instead of potentially failing silently, especially because it'll be an error caused by adding a new media type and not properly configuring it. Unfortunately, I don't think TypeScript does exhaustiveness checks like other languages do.

src/api/v1/event/index.js
```diff
@@ -130,4 +132,25 @@ router.route('/')
     }).catch(next);
   });
 
+/**
+ * Uploads a profile picture for a given event. Requires admin access
+ */
+router.post('/picture', Storage.bufferImageBlob(Media.mediaTypes.EVENT_COVER), async (req, res, next) => {
```
sumeet-bansal, line 138: I think something like `POST /picture/:uuid` would be a bit cleaner. We can also get rid of the UUID check in the controller then.

src/api/v1/event/index.js
```diff
@@ -130,4 +132,25 @@ router.route('/')
...
+  if (!req.user.isAdmin()) return next(new error.Forbidden());
+  if (!req.body.uuid) return next(new error.BadRequest('UUID must be provided in body'));
+  try {
+    const cover = await Storage
+      .uploadToS3(Media.mediaTypes.EVENT_COVER, req.file, req.body.uuid)
+      .then((data) => data.Location)
+      .catch((er) => next(er));
```
sumeet-bansal, line 145: This Promise is already in a `try-catch` so no need to individually catch errors.

src/api/v1/event/index.js
```diff
@@ -130,4 +132,25 @@ router.route('/')
...
 
+/**
+ * Uploads a profile picture for a given event. Requires admin access
+ */
+router.post('/picture', Storage.bufferImageBlob(Media.mediaTypes.EVENT_COVER), async (req, res, next) => {
+  if (!req.user.isAdmin()) return next(new error.Forbidden());
+  if (!req.body.uuid) return next(new error.BadRequest('UUID must be provided in body'));
+  try {
+    const cover = await Storage
+      .uploadToS3(Media.mediaTypes.EVENT_COVER, req.file, req.body.uuid)
```
sumeet-bansal, line 143: Nit: the controller shouldn't need to know where files are being stored, so rename to something more generic (e.g. `upload`).

src/api/v1/event/index.js
```diff
@@ -130,4 +132,25 @@ router.route('/')
...
+      .then((data) => data.Location)
+      .catch((er) => next(er));
+    const updatedEvent = await Event
+      .findByUUID(req.body.uuid)
+      .then((event) => event.updateCover(cover))
+      .catch(next);
```
sumeet-bansal, line 149: This Promise is already in a `try-catch` so no need to individually catch errors.

src/api/v1/user/index.js
```diff
@@ -3,6 +3,7 @@ const express = require('express');
 const error = require('../../../error');
 const { User, Activity, db } = require('../../../db');
 const Storage = require('../../../storage');
+const Media = require('../../../storage/media');
```
sumeet-bansal, line 6: Nit: `Media` is part of `Storage`, so keep the imports structured as such, e.g. `Storage.mediaTypes.PROFILE_PICTURE`.

src/api/v1/user/index.js
```diff
@@ -18,19 +19,16 @@ router.get('/activity', (req, res, next) => {
 /**
  * Uploads a profile picture for the current user.
  */
-router.post('/picture', (req, res, next) => {
+router.post('/picture', Storage.bufferImageBlob(Media.mediaTypes.PROFILE_PICTURE), async (req, res, next) => {
```
sumeet-bansal, line 22: Same here with `POST /picture/:uuid`.

src/api/v1/user/index.js
```diff
@@ -18,19 +19,16 @@ router.get('/activity', (req, res, next) => {
...
-        req.user.updateProfilePicture(location);
-        res.json({ error: null, location });
-      }
-    });
-  } catch (err) {
-    next(err);
+    const profilePicture = await Storage
+      .uploadToS3(Media.mediaTypes.PROFILE_PICTURE, req.file, req.user.uuid)
+      .then((data) => data.Location)
+      .catch((err) => next(err));
```
sumeet-bansal, line 27: Same here with not catching individual Promises in a `try-catch` block.

src/storage/index.js
```diff
@@ -13,22 +13,26 @@ const s3 = new aws.S3({
...
-  const fileUpload = multer({
-    storage: multerS3({
-      s3,
-      bucket: config.s3.bucket,
-      acl: 'public-read',
-      key: (req, file, next) => {
-        next(null, `portal/profiles/${req.user.uuid}${path.extname(file.originalname)}`);
-      },
-    }),
+const uploadToS3 = async (mediaType, file, uuid) => {
```
sumeet-bansal, line 16: Nit: rename to a generic `upload` since the caller doesn't care (and shouldn't know) where files get stored; it's abstracted away in this service.

src/api/v1/event/index.js
```diff
@@ -130,4 +132,25 @@ router.route('/')
...
+/**
+ * Uploads a profile picture for a given event. Requires admin access
+ */
+router.post('/picture', Storage.bufferImageBlob(Media.mediaTypes.EVENT_COVER), async (req, res, next) => {
+  if (!req.user.isAdmin()) return next(new error.Forbidden());
+  if (!req.body.uuid) return next(new error.BadRequest('UUID must be provided in body'));
+  try {
+    const cover = await Storage
+      .uploadToS3(Media.mediaTypes.EVENT_COVER, req.file, req.body.uuid)
+      .then((data) => data.Location)
```
sumeet-bansal, line 144: Move this to inside the `upload` function since the controller expects the function to upload the image and return the location, not some random object from the AWS SDK that has the location as a property. The storage service should handle anything related to S3, including pulling the location from the AWS SDK object.

src/storage/index.js
```diff
@@ -13,22 +13,26 @@ const s3 = new aws.S3({
...
+  const mediaTypeConfig = Media.getMediaTypeConfig(mediaType);
+  const params = {
+    ACL: 'public-read',
+    Body: file.buffer,
+    Bucket: config.s3.bucket,
+    Key: `portal/${mediaTypeConfig.uploadPath}/${uuid}${path.extname(file.originalname)}`,
```
sumeet-bansal, line 22: Nit: include `portal/` as part of the configured upload path. Maybe set like a `BASE_UPLOAD_PATH` or something since it's a single bucket for all ACM stuff and then in the config file, prepend the base path to the other upload paths.

src/storage/index.js
```diff
@@ -13,22 +13,26 @@ const s3 = new aws.S3({
...
+  };
+  return s3.upload(params).promise();
```
sumeet-bansal, line 24:
```javascript
return s3.upload(params).promise().then((data) => data.Location);
```

src/storage/index.js
```diff
@@ -13,22 +13,26 @@ const s3 = new aws.S3({
...
+  const mediaTypeConfig = Media.getMediaTypeConfig(mediaType);
+  return multer({
+    storage: multer.memoryStorage(),
     limits: {
-      fileSize: maxFileSize * BYTES_PER_KILOBYTE,
+      fileSize: mediaTypeConfig.maxFileSize * BYTES_PER_KILOBYTE,
     },
-  }).single(fileTag);
-  return fileUpload;
+  }).single(mediaTypeConfig.fileTag);
```
sumeet-bansal, line 34: Nit: `fileTag` is actually something the controller handles, since that's part of the request and it's getting passed into this service. Maybe include it as a parameter along with `mediaType`? So we still hide the `multer.single` call but have the controller specify from where in the request the file's coming from.

src/api/v1/event/index.js
shravanhariharan2: I'm not entirely sure how authentication works, but for some reason when I change this event route to `POST /picture/:uuid`, `req.user` becomes undefined (preventing me from checking admin status) unless I specify the `authenticated` middleware in the route parameters. With `POST /picture/`, `req.user` gets populated just fine.

The weirder thing is with the user `POST /picture/:uuid` route, `req.user` populates just fine (probably since it's authenticated by default).

Do you know why authentication is different for routes with/without params? Or why the event route in particular is failing? I can't seem to find an explanation online so I think it's likely to do with how the other event routes are setup, not too sure

src/api/v1/user/index.js
shravanhariharan2: I don't understand what you mean by this, are you saying I should have the media types/functions not in a separate file, but under the `storage/index.js`? Because I was thinking abstracting that away from the `storage/index.js` into a `storage/media.js` file makes that index file a lot less cluttered and structured

src/api/v1/event/index.js
sumeet-bansal: Yeah, so because some `/event` routes are public (namely, `/past`, `/future`, and maybe the individual `GET /:uuid` route), `/event` is declared as part of the public API [here](https://github.com/acmucsd/membership-portal/blob/master/src/api/v1/index.js). The `authenticated` middleware must then be run on individual `/event` routes. If you take a look at the other routes, you'll see that they're all preceded by a block of code that runs the `authenticated` middleware (which populates `req.user`) and adds an admin check. To add authentication to an individual route, you can just add `authenticated` to the route, e.g.
```javascript
router.post('/route', authenticated, async (req, res, next) => { ... })
```

src/api/v1/user/index.js
sumeet-bansal: That's fine, just re-export the structs so everything storage-related (especially the enum) can be accessed through the `Storage` import.

shravanhariharan2:
Thanks for giving such detailed reviews, definitely learned a ton here about designing cleaner infrastructure and code.

We can't merge this into master yet though, right? Since `POST /picture/` has been changed to `POST /picture/:uuid`, the front-end needs to make that route change in their API call so that profile picture uploads don't break.

sumeet-bansal:
> We can't merge this into master yet though, right? Since `POST /picture/` has been changed to `POST /picture/:uuid`, the front-end needs to make that route change in their API call so that profile picture uploads don't break.

The perils of split BE/FE repos. Can you make the corresponding change on the FE, get that approved, and then merge both together?

steets250:
We're going to have to time the force-push to master on the FE because merging a PR only affects development. If you let me know when the FE PR is ready, and I can ensure that the changes propagate to the live version of the site.

PR #103, opened 2020-06-04
Add main banner option for file upload

### LINKED ISSUE #99: Add Banner Upload Route for Admins
Given a picture payload, the backend should upload and replace the current banner image. This functionality should only be available to admin users.

shravanhariharan2:
Closes #99 

Designed the implementation where client can specify the type of banner to upload through `bannerType` field in body. This way the API supports different kinds of banner uploads for potentially the main site, communities, etc.


src/api/v1/index.js
```diff
@@ -6,6 +6,7 @@ const router = express.Router();
 // private API, requires authentication
 router.use('/attendance', authenticated, require('./attendance').router);
 router.use('/leaderboard', authenticated, require('./leaderboard').router);
+router.use('/static', authenticated, require('./static').router);
```
sumeet-bansal, line 9: Make this `/admin` since we'll be moving some of the `/user` routes into the admin controller (e.g. point bonuses and milestones).

PR #120, opened 2020-10-03
sliding leaderboards with right bounded queries cached in-memory

### LINKED ISSUE #122: cache right bounded leaderboard queries in-memory
Right bounded leaderboard query results won't change so they can be cached in-memory, maybe for a day or so. TypeORM doesn't have an option to cache in-memory but there's a `QueryResultsCache` interface we can extend and provide our own cache implementation, which could be as simple as a map.

sumeet-bansal:
The leaderboard route supports `from` and `to` query parameters, where each is a UNIX timestamp (in seconds). Left (`from`) bounds are rounded to the start of the day and right (`to`) bounds to the end of the day. Closes #122; caches right bounded queries in-memory.


sumeet-bansal:
Still testing; don't merge.

index.ts
```diff
@@ -23,6 +24,11 @@ createConnection({
   database: Config.database.name,
   entities,
   logging: Config.isDevelopment,
+  cache: {
+    provider(_connection) {
+      return new InMemoryDatabaseCache();
+    },
+  },
```
shravanhariharan2, line 31: should this be `InMemoryDatabaseCache.getInstance()`? doesn't seem like `getInstance()` and the singleton instance are being used anywhere else

shravanhariharan2 (approved):
I didn't test from/to functionality locally since it's pretty infeasible so I guess we'll just have that tested in testing/prod

index.ts
sumeet-bansal: TypeORM probably uses those somewhere since `InMemoryDatabaseCache` extends the TypeORM cache interface.

PR #129, opened 2020-10-06
refactored user services and added a transaction manager

sumeet-bansal:
+ ensure all emails are lowercase
+ updated sliding leaderboard queries to use index
+ cleaned up some of the old user services code
+ inlined some variables
+ removed `id` field from responses
+ set stricter isolation level for some transactions


repositories/ActivityRepository.ts
```diff
@@ -7,7 +7,7 @@ import { BaseRepository } from './BaseRepository';
 
 @EntityRepository(ActivityModel)
 export class ActivityRepository extends BaseRepository<ActivityModel> {
-  private static publicActivityTypes = [
+  private static publicActivities = [
     ActivityType.ACCOUNT_CREATE,
     ActivityType.ATTEND_EVENT,
     ActivityType.ATTEND_EVENT_AS_STAFF,
```
shravanhariharan2, line 13: nit: we can have this be a `Set` so that `isPublicActivityType(type)` is O(1). although the performance boost is most likely miniscule

shravanhariharan2 (approved):

PR #131, opened 2020-10-07
added API usage metrics

sumeet-bansal:
Records metrics for API usage in Datadog.


sumeet-bansal:
Tested with the `api_usage` metric so might need to use a different one in production.

shravanhariharan2 (approved):

PR #133, opened 2020-10-10
Add Event + User Feedback

### LINKED ISSUE #114: Portal Redesign: Feedback
Implement the functionality for users to provide event feedback for any particular event. This feedback can include a subset of 8 feedback options, or a custom feedback if the user wants.

Implement the functionality for users to provide general ACM feedback by filling out a form and for admins to view and confirm receipt of these forms.

API Routes:
- `GET /event/:uuid` - if admin, return event feedback for a particular event along with rest of data, bucketed into feedback name and count
- `POST /event/:uuid/feedback/` - submit event feedback to a particular event (limited to 3)
- `GET /user/feedback` - get all general feedback submitted by the current user - if admin, get all general feedback by all users
- `POST /user/feedback` - submit general feedback

shravanhariharan2:
Closes #114


sumeet-bansal:
Didn't look at the actual code too closely but have some questions about what event feedback is and how it's modeled.

sumeet-bansal (changes requested):
Didn't look at the actual code too closely but have some questions about what event feedback is and how it's modeled.

models/EventFeedbackModel.ts
```diff
@@ -0,0 +1,35 @@
...
+  @JoinColumn({ name: 'event' })
+  @Index('event_feedback_event_index')
+  event: EventModel;
+
+  @ManyToOne((type) => UserModel, (user) => user.eventFeedback, { nullable: false })
+  @JoinColumn({ name: 'user' })
+  @Index('event_feedback_user_index')
+  user: UserModel;
+
+  @Column()
```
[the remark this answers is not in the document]
shravanhariharan2: it definitely doesn't need to be limited to 255 chars but I think there should be some character limit (frontend can decide that) since event feedback is generally meant to be short (at least that's what it looks like from the designs)

repositories/EventRepository.ts
```diff
@@ -29,11 +31,11 @@ export class EventRepository extends BaseRepository<EventModel> {
   }
 
   public async findByUuid(uuid: Uuid): Promise<EventModel> {
-    return this.repository.findOne({ uuid });
+    return this.getBaseFindEventQuery().where({ uuid }).getOne();
   }
 
   public async findByAttendanceCode(attendanceCode: string): Promise<EventModel> {
-    return this.repository.findOne({ attendanceCode });
+    return this.getBaseFindEventQuery().where({ attendanceCode }).getOne();
```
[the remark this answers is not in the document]
shravanhariharan2: nope not needed

types/ApiRequests.ts
```diff
@@ -116,6 +125,10 @@ export interface AttendEventRequest {
   asStaff?: boolean;
 }
 
+export interface AddEventFeedbackRequest {
+  feedback: string[];
```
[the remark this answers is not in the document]
shravanhariharan2: image

judging from this it seems like its short comments / emojis as a reaction to the event. not sure what the custom feedback will allow for so I'll ask Design about that

repositories/EventRepository.ts
```diff
@@ -57,4 +59,18 @@ export class EventRepository extends BaseRepository<EventModel> {
...
+    return this.repository.createQueryBuilder('event')
+      .leftJoinAndSelect('event.feedback', 'feedback')
+      .leftJoinAndSelect('feedback.user', 'feedback.user');
+  }
+}
+
+@EntityRepository(EventFeedbackModel)
+export class EventFeedbackRepository extends BaseRepository<EventFeedbackModel> {
+  public async addEventFeedback(feedback: string[], event: EventModel, user: UserModel): Promise<EventFeedbackModel[]> {
+    const eventFeedback = feedback.map((comment) => EventFeedbackModel.create({ user, event, comment }));
```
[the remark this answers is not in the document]
shravanhariharan2: I think this way it's easier for frontend to visualize / bucket specific feedback by count to see what's the most common feedback and other stats like that. Plus feedback isn't just emojis, they can be custom (and I imagine they can have commas too) so that's why I stuck to splitting them into several records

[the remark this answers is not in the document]
sumeet-bansal: One tool we have at our disposal is serialization: we can't keep arrays in Postgres but what we can do is serialize the array into a string for storage, so the model has an array of strings that gets serialized into a single string, stored as `text`, and then deserialized as arrays for the app to use. You can see the item option metadata as an example of this, where the model property is a JSON object but it's stored as a string. A set of emoji strings don't really need to be their own models, so that's a trick that's available to us.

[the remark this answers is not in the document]
sumeet-bansal: The default `Array.prototype.toString` is pretty lousy for this since it only delimits using commas, but we can use `JSON.stringify` and `JSON.parse` on arrays to easily keep the strings intact, even if they have commas, quotation marks, or anything else, since they all get escaped.

types/ApiRequests.ts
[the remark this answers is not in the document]
shravanhariharan2: re: custom feedback- design says that it can be a few sentences for custom feedback. Also note that design wants us to reward users points (probably 2 pts) for this event feedback. I think leaving feedback as a string array is still fine though.

services/EventService.ts
```diff
@@ -68,8 +73,18 @@ export default class EventService {
     return this.transactions.readWrite(async (txn) => {
       const eventRepository = Repositories.event(txn);
       const event = await eventRepository.findByUuid(uuid);
-      if (!event) throw new NotFoundError();
+      if (!event) throw new NotFoundError('Event not found');
       await eventRepository.deleteEvent(event);
     });
   }
+
+  public async addEventFeedback(uuid: Uuid, feedback: string[], user: UserModel): Promise<PublicEventFeedback[]> {
```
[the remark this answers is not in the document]
shravanhariharan2: asked design about this.
1) and 2) - theyre adding a submit button so only 1 time with no edits
3) 2 days. I'll add that

models/AttendanceModel.ts
```diff
@@ -36,7 +36,14 @@ export class AttendanceModel extends BaseEntity {
       'timestamp',
       'asStaff',
     ]);
-    const publicAttendance: PublicAttendance = { ...rawAttendance };
+
+    const publicAttendance: PublicAttendance = {
+      ...rawAttendance,
+      event: {
+        ...rawAttendance.event,
+        feedback: rawAttendance.event.feedback.map((feedback) => feedback.getPublicEventFeedback()),
```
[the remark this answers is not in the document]
shravanhariharan2: yes it does technically leak feedback but that wouldn't be displayed anywhere since this frontend is just pulling attendance and the event. 

I'm not sure what you mean by having feedback reference attendance. I don't really see a reason for event feedback to be attached to an attendance object (other than the fact that feedback happens after attendance)? Feel free to clarify

models/EventFeedbackModel.ts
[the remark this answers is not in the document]
shravanhariharan2: update on this - custom event feedback can be as long as needed so no limit is necessary

models/AttendanceModel.ts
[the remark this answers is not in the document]
sumeet-bansal: Feedback is a natural extension of attendance: you attend an event then maybe you provide some feedback, where feedback cannot exist without attendance. Right now, feedback is its own model with references to user/event, which implies any user can provide feedback on any event, but that's not true; any user that's attended an event can provide feedback on that specific event, so feedback should be linked to attendance not a user/event pair since that's the relation it's dependent on. Given that feedback's just a string (or an array of strings), not an object/model on its own, it could be a property of the attendance model instead of a model that references attendance to further simplify our schema.

Another strong hint I see that feedback should be linked to attendance is the added complexity in 
 `AttendanceModel::getPublicAttendance`, where getting an attendance requires an additional join from `EventModel` to `EventFeedbackModel` and another join from `EventFeedbackModel` to `UserModel` even though `UserModel`'s already part of `AttendanceModel`. And then for `EventModel`, we join to feedback by default and need to pass in `canSeeFeedback` everywhere. `findByAttendanceCode` doesn't even need feedback, and all those extra joins to our most common queries by far (find past/future events, which're prompted by every login) when the feedback objects are going to be filtered out by a `getPublic` call almost every time are going to have a noticeable impact on our performance since that's a lot of wasted database reading/computation. If feedback's tied to attendance, we don't need to do that, because anyone who can view attendance can view feedback and since it's a rarer more intentional call, we can always return feedback.

And, modeling feedback as part of attendance either makes easier or doesn't affect any feedback queries. Consider the access patterns involved with getting event feedback for a user, for an event, or for an attendance. In all those cases, it'd be much simpler to keep feedback as a property of attendance and use the existing `GET /attendance` routes. If a user wants to see all the feedback they've provided or an event host wants to see all the feedback an event got, I think it'd also add some good context to show feedback as a proportion of attendance.

types/ApiRequests.ts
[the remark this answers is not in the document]
sumeet-bansal: We also need to add feedback as an activity then. I'll try wrapping up the enums-to-strings PR soon so you don't need to add a migration for the new ActivityType, just need to clean up all the changes I made getting it to work.

models/AttendanceModel.ts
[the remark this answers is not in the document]
shravanhariharan2: Thanks for the clear explanation, all these points make sense. 

I think my mind jumped to having event feedback tied to events because it made logical sense to me for feedback to be tied to the thing that is receiving the feedback (like movie feedback to a movie), and I never really brought attendance into the picture (which is tied to a user and event anyway). I'll restructure accordingly

migrations/0021-add-userFeedback-table.ts
```diff
@@ -0,0 +1,64 @@
+import { MigrationInterface, QueryRunner, Table } from 'typeorm';
+
+export class AddUserFeedbackTable1602394486221 implements MigrationInterface {
```
sumeet-bansal, line 3: Kinda annoying but we gotta make sure the migration timestamps are in order; the enum-to-strings migration has timestamp 1602914093929.

models/UserModel.ts
```diff
@@ -72,6 +73,9 @@ export class UserModel extends BaseEntity {
   @OneToMany((type) => OrderModel, (order) => order.user, { cascade: true })
   orders: OrderModel[];
 
+  @OneToMany((type) => FeedbackModel, (feedback) => feedback.user, { cascade: true })
+  Feedback: FeedbackModel;
```
sumeet-bansal, line 77: nit: lowercase to match everything else

models/FeedbackModel.ts
```diff
@@ -0,0 +1,41 @@
+import { BaseEntity, Column, Entity, Generated, Index, JoinColumn, ManyToOne, PrimaryGeneratedColumn } from 'typeorm';
+import { PublicFeedback, Uuid } from '../types';
+import { UserModel } from './UserModel';
+
+@Entity('Feedback')
+export class FeedbackModel extends BaseEntity {
+  @Column({ select: false })
```
sumeet-bansal, line 7: remove `id` from model to match the others

models/FeedbackModel.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  @PrimaryGeneratedColumn('uuid')
+  uuid: Uuid;
+
+  @ManyToOne((type) => UserModel, (user) => user.Feedback, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'user' })
+  @Index('user_feedback_by_user_index')
+  user: UserModel;
+
+  @Column()
```
sumeet-bansal, line 19: Can you specify the type for this?

models/FeedbackModel.ts
```diff
@@ -0,0 +1,41 @@
...
+  @Column('text', { nullable: true })
+  description: string;
+
+  @Column('timestamptz', { default: 'CURRENT_TIMESTAMP(6)' })
+  timestamp: Date;
+
+  @Column({ default: false })
+  acknowledged: boolean;
+
+  public getPublicFeedback(): PublicFeedback {
```
sumeet-bansal, line 31: Is this necessary once we exclude `id` from the model?

migrations/0021-add-userFeedback-table.ts
```diff
@@ -0,0 +1,64 @@
...
+          type: 'uuid',
+        },
+        {
+          name: 'title',
+          type: 'varchar(255)',
+        },
+        {
+          name: 'description',
+          type: 'text',
+          isNullable: true,
```
sumeet-bansal, line 31: Should this be nullable?

migrations/0020-add-eventFeedback-table.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+export class AddEventFeedbackTable1602120071159 implements MigrationInterface {
```
sumeet-bansal, line 3: Nit: not a table, `0021-add-event-feedback.ts` (21, because 20's already merged and we want to maintain linearity).

config/index.ts
```diff
@@ -72,4 +72,9 @@ export const Config = {
   metrics: {
     apiKey: process.env.DATADOG_API_KEY,
   },
+
+  pointReward: {
```
sumeet-bansal, line 76: nit: `feedback[Reward/PointBonus/etc]`

services/AttendanceService.ts
```diff
@@ -53,4 +55,27 @@ export default class AttendanceService {
       return attendance.getPublicAttendance();
     });
   }
+
+  public async addEventFeedback(feedback: string[], uuid: Uuid, user: UserModel): Promise<PublicAttendance> {
```
sumeet-bansal, line 59: what does the `uuid` parameter represent?

services/AttendanceService.ts
```diff
@@ -53,4 +55,27 @@ export default class AttendanceService {
...
+    return this.transactions.readWrite(async (txn) => {
+      const attendanceRepository = Repositories.attendance(txn);
+      const event = await Repositories.event(txn).findByUuid(uuid);
+      if (!event) throw new NotFoundError('Event not found');
+      const hasAttended = await attendanceRepository.hasUserAttendedEvent(user, event);
+      if (!hasAttended) throw new UserError('You cannot provide feedback for an event you didn\'t attend');
```
sumeet-bansal, line 65: nit: "cannot" vs "didn't" (be consistent about abbreviation)

services/AttendanceService.ts
```diff
@@ -53,4 +55,27 @@ export default class AttendanceService {
...
+      const hasProvidedFeedback = await attendanceRepository.hasUserProvidedEventFeedback(user, event);
+      if (hasProvidedFeedback) throw new UserError('You cannot submit feedback for this event more than once');
```
sumeet-bansal, line 67: Don't need to make multiple queries for this. Grab the attendance object, check if it exists + check if `feedback` exists.

services/AttendanceService.ts
```diff
@@ -53,4 +55,27 @@ export default class AttendanceService {
...
+      const eventStart = moment(event.start).valueOf();
```
sumeet-bansal, line 68: space these blocks out (event 404, can't provide/submit feedback, commit feedback, etc.)

services/AttendanceService.ts
```diff
@@ -53,4 +55,27 @@ export default class AttendanceService {
...
+      const twoDaysPastEventEnd = moment(event.end).add(2, 'days').valueOf();
+      if (moment.now() < eventStart) throw new UserError('You cannot submit feedback until the event has started');
```
sumeet-bansal, line 70: Is this error possible? If it's before event start, can't attend, and if didn't attend, there's an error above for that.

services/AttendanceService.ts
```diff
@@ -53,4 +55,27 @@ export default class AttendanceService {
...
+      if (moment.now() > twoDaysPastEventEnd) {
+        throw new UserError('You cannot submit feedback past 2 days of the event ending');
```
sumeet-bansal, line 72: for this and the other error messages: rephrase as "you must submit within 2 days" or "you must attend before feedback"

services/AttendanceService.ts
```diff
@@ -53,4 +55,27 @@ export default class AttendanceService {
...
+      }
+      const attendance = await attendanceRepository.addEventFeedback(user, event, feedback);
```
sumeet-bansal, line 74: if we get the attendance object above, can upsert feedback here

services/FeedbackService.ts
```diff
@@ -0,0 +1,59 @@
...
+      const feedbackRepository = Repositories.feedback(txn);
+      const latestFeedback = await feedbackRepository.getLatestFeedback(user);
+      if (latestFeedback) {
+        const oneWeekAfterFeedback = moment(latestFeedback.timestamp).add(1, 'week');
+        if (oneWeekAfterFeedback.valueOf() > moment.now()) {
+          const daysLeft = oneWeekAfterFeedback.diff(moment(), 'days', false) + 1;
+          throw new UserError(`You must wait ${daysLeft} days to submit more feedback`);
+        }
+      }
+      const fb = await feedbackRepository.upsertFeedback(feedbackObject);
```
sumeet-bansal, line 41: nit: space this out (checks, upsert, etc.)

services/FeedbackService.ts
```diff
@@ -0,0 +1,59 @@
...
+      await Repositories.user(txn).addPoints(user, pointsEarned);
+      return fb;
+    });
+    return addedFeedback.getPublicFeedback();
+  }
+
+  public async updateFeedback(uuid: Uuid, changes: Partial<Feedback>) {
+    const updatedFeedback = await this.transactions.readWrite(async (txn) => {
+      const feedbackRepository = Repositories.feedback(txn);
+      const feedback = await feedbackRepository.getOneFeedback(uuid);
```
sumeet-bansal, line 53: nit: `findByUuid` to match other repositories

types/ApiRequests.ts
```diff
@@ -49,10 +54,22 @@ export interface UserPatches {
   passwordChange?: PasswordUpdate;
 }
 
+export interface FeedbackPatches extends Partial<Feedback> {
```
sumeet-bansal, line 57: maybe `AcknowledgeFeedback` would be a more apt name? no need to extend `Partial<Feedback>` either

types/ApiRequests.ts
```diff
@@ -49,10 +54,22 @@ export interface UserPatches {
...
+
 export interface PatchUserRequest {
   user: UserPatches;
 }
 
+export interface AddFeedbackRequest {
+  feedback: Feedback;
+}
+
+export interface PatchFeedbackRequest {
```
sumeet-bansal, line 69: If someone acknowledges feedback, unacknowledges it, then acknowledges it again, does the user get more points than they should? To avoid this situation, should we disallow un-acknowledging feedback?

types/ApiResponses.ts
```diff
@@ -74,6 +75,11 @@ export interface VerifyAuthTokenResponse extends ApiResponse {
 }
 
 // EVENT
+export interface PublicEventFeedback {
+  uuid: Uuid,
+  user: Uuid,
```
sumeet-bansal, line 80: `user: PublicUser` instead of a uuid?

types/ApiResponses.ts
```diff
@@ -123,6 +129,10 @@ export interface CreateEventResponse extends ApiResponse {
   event: PublicEvent;
 }
 
+export interface AddEventFeedbackResponse extends ApiResponse {
```
sumeet-bansal, line 132: Nit: be consistent about language. There's "submit feedback" "provide feedback" "add feedback" so far.

types/ApiResponses.ts
```diff
@@ -253,6 +263,15 @@ export interface PrivateProfile extends PublicProfile {
   credits: number,
 }
 
+export interface PublicFeedback {
+  uuid: Uuid,
+  user: Uuid,
```
sumeet-bansal, line 268: same here with `PublicUser` vs uuid

sumeet-bansal (changes requested):
Why was the enums-to-strings migration renamed from 0020 to 0022? That'd break the order of the migrations since 0020's already committed.

models/FeedbackModel.ts
shravanhariharan2: think we still need it for `UserModel` -> `PublicProfile` on `this.user`

migrations/0021-add-userFeedback-table.ts
shravanhariharan2: I figured a description doesn't _need_ to be required by the user if they're content with just the title. Should this be `NOT NULL` and just store empty string descriptions if none is submitted?

sumeet-bansal: I think we should encourage some detail, but I guess as usual, ask design. Do we also want to be able to delete feedback/mark as spam or anything like that?

types/ApiRequests.ts
shravanhariharan2: I extended `Partial<Feedback>` here to allow editing title/description as well as acknowledged, but yeah doesn't look like editing feedback after submission is on the figma

migrations/0021-add-userFeedback-table.ts
shravanhariharan2: I think so in case people try to write noisy feedback to potentially gain points. Maybe a `hidden` field on feedback to hide it on the admin side but not delete it entirely?

sumeet-bansal: Tossup on if we want to hide or delete. On one hand, it's just spam so it can be deleted, but on the other someone could hide feedback they don't want others to see (although that's bad faith and someone could just as easily edit the database). Implementation-wise, it'd be simpler to just delete instead of a `markedAsSpam` field, but it could go either way. Ask design I suppose?

Should we email all feedback somewhere, so someone's actually looking at it? Although they'd still have to mark it as acknowledged in the portal. Also is some feedback worth more than others?

shravanhariharan2: A person who submitted feedback would be able to view all their submitted feedback, so with deleting feedback from the admin side, doesn't that also delete it for the person's view too? Like they would all of a sudden not see one of their feedback they submitted without being notified of that. I'll ask design about that

Hmm I'm not sure about emailing feedback somewhere, because we could potentially have hundreds of feedback per week from members so emailing all those would be extra, plus if its in the portal, admin can read + acknowledge in one place.

re: some feedback more than others - design has it where everyone gets 5 right now (not even on acknowledgement, just on submission, which should probably be moved to acknowledgement), and I think that's fine, like if the feedback passes the threshold of acknowledgement anyway then it deserves 5 points.

sumeet-bansal (changes requested):

sumeet-bansal:
Also I've been really bad about tagging releases but this PR should be 2.4.0; can you update the package.json and lockfile?

migrations/0023-add-feedback-table.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export class SubmitFeedbackTable1603825929793 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.createTable(new Table({
+      name: 'Feedback',
+      columns: [
+        {
+          name: 'id',
+          type: 'integer',
+          isGenerated: true,
```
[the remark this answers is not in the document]
shravanhariharan2: `isGenerated` should be the auto-increment thing. The `uuid` field is the primary key so that's the one w/ generation strategy and primary key fields

migrations/0023-add-feedback-table.ts
```diff
@@ -0,0 +1,67 @@
...
+          name: 'description',
+          type: 'text',
+        },
+        {
+          name: 'timestamp',
+          type: 'timestamptz',
+          default: 'CURRENT_TIMESTAMP(6)',
+        },
+        {
+          name: 'acknowledged',
```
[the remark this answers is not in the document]
shravanhariharan2: I don't think we're doing a "mark as spam" thing, since Kendall mentioned it would be dealt on a case-by-case basis and since we're only rewarding points on acknowledgement rather than submission. Do you think we should have it still?

models/FeedbackModel.ts
```diff
@@ -0,0 +1,41 @@
...
+export class FeedbackModel extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  uuid: Uuid;
+
+  @ManyToOne((type) => UserModel, (user) => user.feedback, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'user' })
+  @Index('user_feedback_by_user_index')
+  user: UserModel;
+
+  @Column('varchar')
```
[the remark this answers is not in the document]
shravanhariharan2: should we specify column types in the rest of the `*Model.ts` files? models like EventModel don't have column types for strings

services/FeedbackService.ts
```diff
@@ -0,0 +1,66 @@
...
+    const feedbackObject = FeedbackModel.create({ ...feedback, user });
+    const addedFeedback = await this.transactions.readWrite(async (txn) => {
+      const feedbackRepository = Repositories.feedback(txn);
+
+      const latestFeedback = await feedbackRepository.getLatestFeedback(user);
+      if (latestFeedback) {
+        const oneWeekAfterFeedback = moment(latestFeedback.timestamp).add(1, 'week');
+        if (oneWeekAfterFeedback.valueOf() > moment.now()) {
+          const daysLeft = oneWeekAfterFeedback.diff(moment(), 'days', false) + 1;
+          throw new UserError(`You must wait ${daysLeft} days to submit more feedback`);
```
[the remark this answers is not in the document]
shravanhariharan2: I agree, but how should we go about granting points only once a week? Feedback submitted within a week of another should still be acknowledgeable but not be rewarded points, and we don't want to refresh the week duration on feedback that hasn't been rewarded. So we need a way of keeping track of the last feedback submitted that has been rewarded points so that we can reward/not reward accordingly. Maybe a `hasBeenRewarded` field on each feedback? So we can query the latest feedback that has been rewarded and check the week long duration from that?

services/FeedbackService.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+    return addedFeedback.getPublicFeedback();
+  }
+
+  public async acknowledgeFeedback(uuid: Uuid) {
+    const acknowledgedFeedback = await this.transactions.readWrite(async (txn) => {
+      const feedbackRepository = Repositories.feedback(txn);
+      const feedback = await feedbackRepository.findByUuid(uuid);
+      if (!feedback) throw new UserError('Feedback not found');
+      if (feedback.acknowledged) throw new UserError('This feedback has already been acknowledged');
```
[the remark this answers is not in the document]
shravanhariharan2: But if this route is reached with a feedback that has already been acknowledged, doesn't that indicate some kind of user error made by the API caller? It might be confusing to return a successful response when nothing actually happened and that request shouldn't have been made.

services/FeedbackService.ts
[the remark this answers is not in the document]
shravanhariharan2: yep :^)

I can remove the +1 and return hours too

models/FeedbackModel.ts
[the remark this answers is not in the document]
sumeet-bansal: I think right now it might be that `text` is specified and `varchar` is the default. Your call if you want to leave it as that or make all string types explicit, but might be good to.

services/FeedbackService.ts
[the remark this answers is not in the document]
sumeet-bansal: Boolean check for acknowledged feedback within a week of the feedback being acknowledged?

But if we're acknowledging it maybe we should just grant points every time. This is where "acknowledged" and "acknowledged and useful" distinctions might come in handy, but again, maybe just grant on all acknowledged feedback? Maybe an internal "acknowledged but not useful" that's publicly displayed as just "submitted"?

services/FeedbackService.ts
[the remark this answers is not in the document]
sumeet-bansal: If you think of this request as "Acknowledge this unacknowledged feedback" then yes it's a user error. If you think of it as "Make sure this feedback is marked as acknowledged" then it's fine to return early. I prefer the latter framing because if two people acknowledge it at the same time, who cares which request goes through? To both of them, they marked it as acknowledged and don't care what its status was before.

services/FeedbackService.ts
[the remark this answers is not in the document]
shravanhariharan2: I'm thinking of adding an enum `status(SUBMITTED, MARKED_AS_SPAM, ACKNOWLEDGED, REWARDED)`. When admin acknowledges a specific feedback, if it's been more than a week since the last REWARDED feedback, then it gets marked as REWARDED, otherwise its marked as ACKNOWLEDGED. I'll include MARKED_AS_SPAM too since we aren't restricting submission rates so spam by trolls is somewhat inevitable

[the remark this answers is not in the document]
sumeet-bansal: Honestly, I think we should just grant points on all acknowledgement. People submitting a lot of useful feedback should be rewarded not penalized. I think `status(SUBMITTED, MARKED_AS_SPAM/DISCARDED/IGNORED, ACKNOWLEDGED)` is fine, where any acknowledgement carries a point bonus and `IGNORED` is shown to the submitter as `SUBMITTED`.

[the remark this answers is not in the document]
shravanhariharan2: sounds good

api/controllers/FeedbackController.ts
```diff
@@ -0,0 +1,42 @@
...
+  async getFeedback(@AuthenticatedUser() user: UserModel): Promise<GetFeedbackResponse> {
+    const canSeeAllFeedback = PermissionsService.canAcknowledgeFeedback(user);
+    const feedback = await this.feedbackService.getFeedback(canSeeAllFeedback, user);
+    return { error: null, feedback };
+  }
+
+  @Post()
+  async submitFeedback(@Body() submitFeedbackRequest: SubmitFeedbackRequest,
+    @AuthenticatedUser() user: UserModel): Promise<SubmitFeedbackResponse> {
+    const feedback = await this.feedbackService.submitFeedback(user, submitFeedbackRequest.feedback);
```
sumeet-bansal, line 30: Do we want to restrict leaving feedback to users that're email-verified or something?

api/decorators/Validators.ts
```diff
@@ -93,3 +93,48 @@ class MatchingPasswordsValidator implements ValidatorConstraintInterface {
...
+}
+
+@ValidatorConstraint()
+class FeedbackStatusValidator implements ValidatorConstraintInterface {
+  validate(feedbackStatus: FeedbackStatus): boolean {
+    return Object.values(FeedbackStatus).includes(feedbackStatus) && feedbackStatus !== FeedbackStatus.SUBMITTED;
+  }
+
+  defaultMessage(): string {
+    return 'Invalid feedback status';
```
sumeet-bansal, line 134: Some of our error messages (like this) could be more user-friendly. For example, `Invalid feedback status` &rarr; `Feedback status must be one of ['IGNORED', 'ACKNOWLEDGED']`.

api/validators/FeedbackControllerRequests.ts
```diff
@@ -0,0 +1,43 @@
...
+  FeedbackStatus,
+} from '../../types';
+
+export class Feedback implements IFeedback {
+  @IsDefined()
+  @IsNotEmpty()
+  title: string;
+
+  @IsDefined()
+  @Length(100)
```
sumeet-bansal, line 19: nit: `@MinLength(100)`

models/FeedbackModel.ts
```diff
@@ -0,0 +1,41 @@
...
+import { UserModel } from './UserModel';
+
+@Entity('Feedback')
+export class FeedbackModel extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  uuid: Uuid;
+
+  @ManyToOne((type) => UserModel, (user) => user.feedback, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'user' })
+  @Index('user_feedback_by_user_index')
```
sumeet-bansal, line 12: match `feedback_by_user_index` in the migration

services/FeedbackService.ts
```diff
@@ -0,0 +1,55 @@
...
+      const feedback = await feedbackRepository.findByUuid(uuid);
+      if (!feedback) throw new UserError('Feedback not found');
+      if (feedback.status !== FeedbackStatus.SUBMITTED) {
+        throw new UserError('This feedback has already been responded to');
+      }
+
+      const { user } = feedback;
+      if (status === FeedbackStatus.ACKNOWLEDGED) {
+        const pointsEarned = Config.pointReward.FEEDBACK_POINT_REWARD;
+        await Repositories.activity(txn).logActivity(user, ActivityType.SUBMIT_FEEDBACK, pointsEarned);
```
sumeet-bansal, line 47: Activity here isn't `SUBMIT_FEEDBACK` it's more like `FEEDBACK_ACKNOWLEDGED`, right?

services/PermissionsService.ts
```diff
@@ -12,6 +12,10 @@ export default class PermissionsService {
     return user.isAdmin();
   }
 
+  public static canAcknowledgeFeedback(user: UserModel): boolean {
```
sumeet-bansal, line 15: `canRespondToFeedback` maybe?

types/ApiResponses.ts
```diff
@@ -73,8 +74,6 @@ export interface VerifyAuthTokenResponse extends ApiResponse {
   authenticated: boolean;
 }
 
-// EVENT
```
sumeet-bansal, line 76: Can you add this back?

types/Enums.ts
```diff
@@ -30,4 +30,23 @@ export enum ActivityType {
   BONUS_POINTS = 'BONUS_POINTS',
   MILESTONE = 'MILESTONE',
   ORDER_MERCHANDISE = 'ORDER_MERCHANDISE',
+  SUBMIT_EVENT_FEEDBACK = 'SUBMIT_EVENT_FEEDBACK',
+  SUBMIT_FEEDBACK = 'SUBMIT_FEEDBACK',
```
sumeet-bansal, line 34: Should this be a public activity?

shravanhariharan2: I don't think so, other users don't really need to be aware whether or not a particular user submits feedback (similar to ordering merch)

sumeet-bansal: Not for other people but for the user that submitted the feedback to see.

sumeet-bansal: Maybe "public" goes from being a boolean to `enum(PUBLIC, PRIVATE, USER_ONLY)`?

shravanhariharan2: ok sounds good, yeah things like login/auth (basically whatever is `public = false` right now) should be `HIDDEN` and merch/feedback should be `PRIVATE`

shravanhariharan2:
@sumeet-bansal ready for review

repositories/FeedbackRepository.ts
```diff
@@ -0,0 +1,32 @@
...
+    return this.getBaseFindQuery().where({ uuid }).getOne();
+  }
+
+  public async getAllFeedbackForUser(user: UserModel): Promise<FeedbackModel[]> {
+    return this.getBaseFindQuery().where({ user }).getMany();
+  }
+
+  public async upsertFeedback(feedback: FeedbackModel,
+    changes?: Partial<FeedbackModel>): Promise<FeedbackModel> {
+    if (changes) FeedbackModel.merge(feedback, changes);
```
sumeet-bansal, line 23: nit: explicitly assign this
```typescript
if (changes) feedback = FeedbackModel.merge(feedback, changes);
```

services/FeedbackService.ts
```diff
@@ -0,0 +1,56 @@
...
+      return Repositories.feedback(txn).upsertFeedback(FeedbackModel.create({ ...feedback, user }));
+    });
+    return addedFeedback.getPublicFeedback();
+  }
+
+  public async updateFeedbackStatus(uuid: Uuid, status: FeedbackStatus) {
+    const acknowledgedFeedback = await this.transactions.readWrite(async (txn) => {
+      const feedbackRepository = Repositories.feedback(txn);
+      const feedback = await feedbackRepository.findByUuid(uuid);
+      if (!feedback) throw new UserError('Feedback not found');
```
sumeet-bansal, line 40: `NotFoundError`?

sumeet-bansal (approved):

shravanhariharan2:
this is a real application submissions PR type beat

PR #136, opened 2020-10-18
Remove all banner code

### LINKED ISSUE #119: Remove Banner Features
The frontend no longer uses the banner, so all code related to uploading the banner can be removed.

shravanhariharan2:
Closes #119 

Also removed all static banner files on S3


sumeet-bansal:
I think we should leave the banner code for now in case someone decides to use it later and revisit this down the road, like the end of the academic year. What do you think?

shravanhariharan2:
sure I guess it doesn't hurt to wait till then.

sumeet-bansal:
It's also an API break, so bad practice.

/remind me to follow up on this on July 1

PR #139, opened 2020-11-03
added test to verify migration timestamps are strictly increasing

sumeet-bansal:
Someone is 100% going to accidentally commit mis-ordered timestamps in the future. This hopefully prevents that. It'll still be merged, but it won't be deployed, although we can prevent bad timestamps from being merged by requiring all PRs are up-to-date with `master`.


shravanhariharan2 (approved):
good call. we should make all PR's (which should be just feedback) be up to date to master so that no bad timestamps get merged

sumeet-bansal:
No, "all PRs" like ever. If two people branch off after migration 30 and each create their own 31, then their PRs will merge fine but break when combined, so to merge a PR it must be up-to-date with `master` and contain any migrations added since they branched.

PR #143, opened 2020-11-21
Remove "public" activity reference in leaderboard query

shravanhariharan2:
hotfix oops


sumeet-bansal:
This index also needs to be updated: https://github.com/acmucsd/membership-portal/blob/4d5bd846b7488c9605528d5f43072b2d10ad9878/models/ActivityModel.ts#L14

sumeet-bansal:
The query that's fixed in this PR might also be incorrect now since an activity could be PRIVATE but still earn points.

shravanhariharan2:
I'll change it, but I think technically it still works as intended since all activities before now are either `PUBLIC` or `HIDDEN`, (and the earliest timestamp would always be before now). Unless we plan on only storing activities for the past x years/quarters since the table will become infinitely large otherwise? What's the plan for that?

sumeet-bansal:
> I'll change it, but I think technically it still works as intended since all activities before now are either `PUBLIC` or `HIDDEN`, (and the earliest timestamp would always be before now).

I guess that's true for earliest timestamp since that's based on what's in our database rn, but it's still incorrect and only works because of our current app state.

> Unless we plan on only storing activities for the past x years/quarters since the table will become infinitely large otherwise? What's the plan for that?

Not sure what this means, but the table will still be very finite. The portal after its first year had 20-30k rows. Our current plan has an upper bound of 10M rows, so even if the portal database adds 100k rows every year, it'd still take decades to reach our plan limit, so I don't think we need to worry about the database growing too big. I think it'd be p cool for people to look at historical ACM data from years ago, so I personally don't think we should delete data after some number of years.

PR #144, opened 2020-11-29
Add Test Suite for Feedback 

shravanhariharan2:
Setup the infrastructure for feedback testing and for future test suites


tests/data/FactoryUtils.ts
```diff
@@ -8,8 +8,9 @@ export default class FactoryUtils {
     return values[i];
   }
 
-  public static getRandomNumber(range: number, minimum = 0, scale = 1): number {
-    return (Math.floor(Math.random() * range) % scale) + minimum;
```
shravanhariharan2, line 12: this function wasn't working for ranges with `scale = 1` (since any integer mod 1 is 0, so it would always return `minimum`), so I refactored it a bit

shravanhariharan2 (commented):
@sumeet-bansal any thoughts on the whole setup? The idea is for a test suite for a particular API, the respective controller can be accessed as a singleton with services/repositories injected already. Will implement the rest of actual tests after making any setup changes needed.

tests/TestUtils.ts
```diff
@@ -0,0 +1,23 @@
+export class TestUtils {
+  public static comparator(sortKey: string): (object1: object, object2: object) => number {
```
sumeet-bansal, line 2: https://flaviocopes.com/how-to-sort-array-of-objects-by-property-javascript/

tests/TestUtils.ts
```diff
@@ -0,0 +1,23 @@
...
+    return (object1: object, object2: object) => {
+      if (object1[sortKey] < object2[sortKey]) return -1;
+      if (object1[sortKey] > object2[sortKey]) return 1;
+      return 0;
+    };
+  }
+
+  public static toMatchArrayContents(received: any[], expected: any[], sortKey = 'uuid') {
```
sumeet-bansal, line 10: https://jestjs.io/docs/en/expect#expectarraycontainingarray

tests/data/FactoryUtils.ts
sumeet-bansal: nit: rename `range` to `value` or something

tests/data/FeedbackFactory.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Feedback, FeedbackType } from '../../types';
+import FactoryUtils from './FactoryUtils';
+
+export class FeedbackFactory {
+  public static create(n: number) {
+    return FactoryUtils.create(n, FeedbackFactory.fake);
+  }
+
+  public static with(...substitutes: Partial<Feedback>[]): Feedback[] {
+    return substitutes.map((sub) => ({ ...FeedbackFactory.fake(), sub }));
```
sumeet-bansal, line 11: Should this should be `...sub`?

tests/data/FeedbackFactory.ts
```diff
@@ -0,0 +1,26 @@
...
+    return {
+      title: faker.random.hexaDecimal(10),
+      description: faker.lorem.sentences(2),
+      type: FeedbackFactory.randomFeedbackType(),
+    };
+  }
+
+  private static randomFeedbackType(): FeedbackType {
+    const randomIndex = FactoryUtils.getRandomNumber(0, Object.keys(FeedbackType).length - 1);
+    return Object.values(FeedbackType)[randomIndex];
```
sumeet-bansal, line 24:
```suggestion
    return FactoryUtils.pickRandomValue(Object.values(FeedbackType));
```

tests/data/FactoryUtils.ts
```diff
@@ -8,8 +8,9 @@ export default class FactoryUtils {
...
+  public static getRandomNumber(min: number, max: number, scale = 1): number {
+    const range = Math.floor(Math.random() * (max - min + 1)) + min;
+    return Math.floor(range / scale) * scale;
```
sumeet-bansal, line 13:
```suggestion
    return (range % scale) * scale;
```

tests/controllers/index.ts
```diff
@@ -0,0 +1,15 @@
+import Container from 'typedi';
+import FeedbackService from '../../services/FeedbackService';
+import { FeedbackController } from '../../api/controllers/FeedbackController';
+
+export class ControllerFactory {
+  private static feedbackController: FeedbackController = null;
+
+  public static feedback(): FeedbackController {
+    if (!ControllerFactory.feedbackController) {
+      const feedbackService = Container.get(FeedbackService);
```
sumeet-bansal, line 10: Can we avoid interacting with the container entirely and manually construct the feedback service?

tests/TestUtils.ts
shravanhariharan2: I think this is slightly different - we need a way to have `expectArrayContaining` but with objects that partially match (`expectArrayContaining` expects exact matches).  In the case with feedback, we're comparing two arrays of `Feedback[]` and `PublicFeedback[]`, and in other tests, we'd compare the `*Model[]` array to `Public*[]` which also only partially match. There doesn't seem to be a way to do this through jest's current API

sumeet-bansal: Why not map the Model arrays to the PublicModel arrays?

tests/data/FactoryUtils.ts
shravanhariharan2: I'm not sure these are doing the same thing (with the new min, max, scale thing). Let's say we want a value between 30 and 120 with a scale of 30. `range` will give us a value between 30 and 120 that has scale of 1 (e.g. 68), and current line 13 will floor this value to the nearest `scale` (so 68 -> 60, 56 -> 30, 90 -> 90).

New line 13 would do: `(range % scale) * scale` = (68 % 30) * 30 = 8 * 30 = 240 which isn't in the range

tests/TestUtils.ts
shravanhariharan2: oh yeah we can do that for the other tests, but the feedback factory creates `Feedback` objects rather than `FeedbackModel` ones because `FeedbackModel` relies on a `UserModel` to submit it. 

I personally think it better mimics the actual flow of feedback to have `Feedback` objects created in the factory and `PortalState::submitFeedback(feedback: Feedback, user: UserModel)` actually attaching a user to it, but I suppose having  `FeedbackFactory::create(n: number, user: UserModel): FeedbackModel[]` instead doesn't hurt too

sumeet-bansal: Current FeedbackFactory's fine; I meant for the assertions, we can map Model[] to PublicModel[] instead of creating our own matcher. I think the tests should use the controller to submit feedback instead though.

shravanhariharan2: I was thinking about having a different test for each controller alone so that only one thing's tested at a time, so all data that would need to be inserted for a test is done with `PortalState` and just 1 controller is tested rather than 2+.

sumeet-bansal: > I was thinking about having a different test for each controller alone so that only one thing's tested at a time

Do you mean one test per route/controller method? Most of the POST/GET routes are pretty basic CRUD operations so I think those are fine to test together, e.g. submit a feedback and then get it. Looking at `FeedbackService` the only code that's executed in that flow is pretty simple stuff like creating and inserted via TypeORM and then a basic `SELECT` query.

shravanhariharan2:
@sumeet-bansal A while back we talked about how we should test validation, and you said that we can do manual validation with class-validator for that. Doesn't that cause a lot of duplicate validation checks between the test suite and middleware though? This along with the need to implement our own DI with this current setup makes me wonder if actually spinning up the server and sending requests is better than just using the controller methods. Thoughts?

Also, since we were planning on exposing API wrapper functions as a package anyway, I think this would a good time to set that up for both testing and for the package. This would also keep the responsibility of DI / validation in the server's hands and not ours.

sumeet-bansal:
> Doesn't that cause a lot of duplicate validation checks between the test suite and middleware though?

From what I remember, routing-controllers attaches its own middleware for validation, so I don't think any controllers we initialize validate inputs. If you want to test something, you should only test that one thing, which's why I said if we're testing validation we can use the class-validator `validate` call to only test the validators.

In general, I think it's better to avoid DI magic since it can be really confusing to understand how something that uses it works. Spinning up a server and sending requests works if we're testing the server as a black box, but if we want to make sure specific things are working as intended (e.g. validation, the logic in some service method, database queries) then we need to test those pieces in isolation.

Using the server as a black box also means the entire test server needs to be in a testable state since all the code runs instead of just the bits you're testing, so for example if you want to make sure a member can't access some object like feedback or a merch order, you need to create the db connection and server, get
 auth tokens,

sumeet-bansal:
> Also, since we were planning on exposing API wrapper functions as a package anyway, I think this would a good time to set that up for both testing and for the package.

I'm waiting for this PR at least to go through since a bunch of files are gonna get moved around, but yeah, I'll get on that soon. I started a little on it and it doesn't seem too difficult.

shravanhariharan2:
@sumeet-bansal in this setup, I try to keep the tests as small as possible by hiding away details like connection and portal state in separate factories, but there's some inconsistency in abstraction (such as calling `conn.manager.findOne()` inside tests) and some code duplication that I'm not sure how to change. In general, I'm still not really sure where to draw the line between clean abstraction and overcomplicating the architecture, so let me know what you think of the design and its sustainability.

Also feel free to suggest any additional test cases

shravanhariharan2:
@sumeet-bansal can you re-review this when you get the chance?

types/HttpEnums.ts
```diff
@@ -0,0 +1,9 @@
+export enum HttpStatusCode {
```
sumeet-bansal, line 1: Is this used anywhere? I don't see any uses besides re-exporting in `types/Enums`.

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    const errors = await validate(plainToClass(Feedback, feedback));
+
+    expect(errors).toBeDefined();
+    expect(errors).toHaveLength(1);
+    expect(errors[0].property).toEqual('description');
+    expect(errors[0].constraints.minLength).toBeDefined();
+  });
+
+  test('persists submission data and activity in database on successful submission', async () => {
```
sumeet-bansal, line 50: I think it'd be a little cleaner here if we used the GET route instead of querying the database, plus we get to make sure that route's working too.

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    await FeedbackControllerMock.mock([user]);
+    const submittedFeedback = await FeedbackControllerMock.submitFeedback(feedback, user);
+
+    const conn = await DatabaseConnection.get();
+    const [persistedFeedback] = await conn.manager.find(FeedbackModel, { relations: ['user'] });
+    const activity = await conn.manager.find(ActivityModel, { relations: ['user'] });
+    const feedbackSubmissionActivity = activity[1];
+
+    expect(submittedFeedback).toStrictEqual(persistedFeedback.getPublicFeedback());
```
sumeet-bansal, line 62: Same thing here with matching the entire object once instead of separate assertions for each property.

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    expect(feedbackSubmissionActivity.type).toEqual(ActivityType.SUBMIT_FEEDBACK);
+    expect(feedbackSubmissionActivity.scope).toEqual(ActivityScope.PRIVATE);
+    expect(feedbackSubmissionActivity.user).toStrictEqual(user);
+  });
+
+  test('is visible from all users if user is admin, otherwise from only current user', async () => {
```
sumeet-bansal, line 68: nit: this was a little confusing to read, maybe something like "admins can feedback from any/all user, members can only see their own"

sumeet-bansal: might also be worth breaking up into separate tests: "admins can see all users' feedback" and "members can only see their feedback"

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    expect(submittedFeedback1).toHaveLength(1);
+    expect(submittedFeedback2).toHaveLength(1);
+    expect(submittedFeedback1[0]).toMatchObject(feedback1);
+    expect(submittedFeedback2[0]).toMatchObject(feedback2);
+    expect(allSubmittedFeedback).toHaveLength(2);
+    expect(allSubmittedFeedback).toEqual(expect.arrayContaining([...submittedFeedback1, ...submittedFeedback2]));
+  });
+
+  test('can be acknowledged and rewarded points by admin', async () => {
```
sumeet-bansal, line 89: nit: "admin can acknowledge and reward points for feedback"

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    const conn = await DatabaseConnection.get();
+    const persistedUser = await conn.manager.findOne(UserModel, { where: { uuid: user.uuid } });
+    const feedbackPointReward = Config.pointReward.FEEDBACK_POINT_REWARD;
+
+    expect(acknowledgedFeedback.status).toEqual(FeedbackStatus.ACKNOWLEDGED);
+    expect(persistedUser.points).toEqual(user.points + feedbackPointReward);
+  });
+
+  test('can be ignored and not rewarded anything by admin', async () => {
```
sumeet-bansal, line 106: nit: "admin can ignore and not reward points for feedback"

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    expect(persistedUser.points).toEqual(user.points);
+  });
+
+  test('cannot be ackowledged or ignored after already being acknowledged or ignored', async () => {
+    const [user] = UserFactory.create(1);
+    const [admin] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+    const [feedback1, feedback2] = FeedbackFactory.create(2);
+
+    await FeedbackControllerMock.mock([user, admin]);
+    const submittedFeedback1 = await FeedbackControllerMock.submitFeedback(feedback1, user);
```
sumeet-bansal, line 128: nit: `feedbackToAcknowledge`, `feedbackToIgnore`

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    const submittedFeedback2 = await FeedbackControllerMock.submitFeedback(feedback2, user);
+
+    const uuid1 = submittedFeedback1.uuid;
+    const uuid2 = submittedFeedback2.uuid;
+
+    await FeedbackControllerMock.acknowledgeFeedback(uuid1, admin);
+    await FeedbackControllerMock.ignoreFeedback(uuid2, admin);
+
+    const acknowledgeAcknowledgedFeedback = () => FeedbackControllerMock.acknowledgeFeedback(uuid1, admin);
```
sumeet-bansal, line 137: Why create single-use functions here and then call them right after? vs inlining them.

tests/index.ts
```diff
@@ -0,0 +1,7 @@
+import 'reflect-metadata';
+import { useContainer as routingUseContainer } from 'routing-controllers';
+import { useContainer as ormUseContainer } from 'typeorm';
+import Container from 'typedi';
+
+routingUseContainer(Container);
+ormUseContainer(Container);
```
sumeet-bansal, line 7: Is this all necessary if we're manually instantiating everything through constructors?

tests/feedback.test.ts
shravanhariharan2: Mainly for conveying via. function name what the function is doing, but with the variable renaming change above, these functions aren't needed anymore.

tests/index.ts
shravanhariharan2: Don't think it's necessary anymore

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+  await DatabaseConnection.clear();
+});
+
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('feedback submission', () => {
+  test('returns proper feedback response on successful submission', async () => {
```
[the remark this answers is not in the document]
shravanhariharan2: Wouldn't that be an extra unnecessary query though? Also, this test is explicitly for checking the `submitFeedback` response object's validity, and if we want to check the response of `getFeedback`, then I think that should be in a different test

tests/controllers/FeedbackControllerMock.ts
```diff
@@ -0,0 +1,41 @@
+import { Connection } from 'typeorm';
+import { UserModel } from '../../models/UserModel';
+import { DatabaseConnection, PortalState } from '../data';
+import { PublicFeedback, Feedback, Uuid, FeedbackStatus } from '../../types';
+import { ControllerFactory } from './ControllerFactory';
+
+export class FeedbackControllerMock {
```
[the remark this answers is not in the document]
shravanhariharan2: I mainly built it to abstract away the database connection `conn` (and things like `PortalState`) from the tests to make them simpler and easier to read

tests/feedback.test.ts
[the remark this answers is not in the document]
sumeet-bansal: I just don't get the point of the test. Like, the response object is something we return and we have a type for it so why do we need to check its validity. If we want to make sure the submission went through, then we could call GET or query the database, but I don't see the point in checking the response object since it's basically just whatever we put in the request plus some constants.

tests/feedback.test.ts
shravanhariharan2: Wouldn't that make this test case dependent on whether or not a different controller is working? If that controller wasn't working, then this test case would fail, which might cause some confusion / extra looking since the actual feedback functionality would be working

tests/feedback.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    const [user] = UserFactory.create(1);
+    const [feedback] = FeedbackFactory.create(1);
+
+    await FeedbackControllerMock.mock([user]);
```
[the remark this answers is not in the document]
shravanhariharan2: See comment above about purpose of FeedbackControllerMock

tests/feedback.test.ts
sumeet-bansal: The GET's a pretty simple route, just a permissions check and db query.

tests/controllers/FeedbackControllerMock.ts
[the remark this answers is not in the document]
sumeet-bansal: The `beforeAll/beforeEach` already take care of the db connection, so it's
```typescript
await FeedbackControllerMock.mock(users);
```
vs.
```typescript
await new PortalState().createUsers(users).write();
```
At that point you basically have a whole wrapper for that change alone, and you'd have to do something similar for all the other controllers. Makes more sense to just keep the `PortalState` stuff in the test itself since that's what it was made for. Also as a nit, nothing's being "mocked"; we're creating real users in the database and letting the controller execute normally instead of mocking behavior for whatever components are in the controller.

shravanhariharan2:
@sumeet-bansal here's what it looks like without the FeedbackControllerMock class. IMO the tests themselves are a little less clean since now there's a tiny bit more setup involved within the connection and the controller parameters are inconsistently nested (i.e. some params such as feedback are wrapped while others such as user are not). Plus, the tests have different layers of abstraction now with manually writing portal state and calling top-level controllers, which may make it slightly more confusing to read.

I personally think it's worth wrapping each controller in an extra class that manages connections and cleans up requests/responses, but maybe there's a better way to design it, or maybe I'm too critical on having small tests and in reality it doesn't really matter. Let me know what you think

tests/controllers/ControllerFactory.ts
```diff
@@ -0,0 +1,15 @@
+import { Connection } from 'typeorm';
+import FeedbackService from '../../services/FeedbackService';
+import { FeedbackController } from '../../api/controllers/FeedbackController';
+
+export class ControllerFactory {
+  private static feedbackController: FeedbackController = null;
```
sumeet-bansal, line 6: `DatabaseConnection` has a class-wide connection because it makes managing connections simpler like with closing them out, but it's fine for `ControllerFactory` to instantiate the controller  every time.

tests/feedback.test.ts
```diff
@@ -0,0 +1,195 @@
...
+describe('feedback submission', () => {
+  test('properly persists on successful submission', async () => {
+    const conn = await DatabaseConnection.get();
+    const [user] = UserFactory.create(1);
+    const [feedback] = FeedbackFactory.create(1);
+
+    const feedbackController = ControllerFactory.feedback(conn);
+    await new PortalState().createUsers([user]).write(conn);
+
+    await ControllerFactory.feedback(conn).submitFeedback({ feedback }, user);
```
sumeet-bansal, line 35:
```suggestion
    await feedbackController.submitFeedback({ feedback }, user);
```

tests/feedback.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const errors = await validate(plainToClass(Feedback, feedback));
+
+    expect(errors).toBeDefined();
+    expect(errors).toHaveLength(1);
+    expect(errors[0].property).toEqual('description');
+    expect(errors[0].constraints.minLength).toBeDefined();
+  });
+
+  test('persists submission data and activity in database on successful submission', async () => {
```
sumeet-bansal, line 58: Doesn't the sample test already verify that things get persisted to the database?

tests/feedback.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const feedbackController = ControllerFactory.feedback(conn);
+    await new PortalState().createUsers([user1, user2, admin]).write(conn);
+
+    const submittedFeedback1Response = await feedbackController.submitFeedback({ feedback: feedback1 }, user1);
+    const submittedFeedback2Response = await feedbackController.submitFeedback({ feedback: feedback2 }, user2);
+    const allSubmittedFeedbackResponse = await feedbackController.getFeedback(admin);
+
+    expect(submittedFeedback1Response.feedback).toMatchObject(feedback1);
+    expect(submittedFeedback2Response.feedback).toMatchObject(feedback2);
```
sumeet-bansal, line 93: I still don't understand the need to verify the submission responses.

tests/feedback.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const submittedFeedbackResponse = await feedbackController.submitFeedback({ feedback }, user);
+    const { uuid } = submittedFeedbackResponse.feedback;
+    const acknowledgedFeedback = await feedbackController.updateFeedbackStatus(uuid, { status }, admin);
+
+    const persistedUser = await conn.manager.findOne(UserModel, { where: { uuid: user.uuid } });
+    const feedbackPointReward = Config.pointReward.FEEDBACK_POINT_REWARD;
+
+    expect(acknowledgedFeedback.feedback.status).toEqual(FeedbackStatus.ACKNOWLEDGED);
+    expect(persistedUser.points).toEqual(user.points + feedbackPointReward);
```
sumeet-bansal, line 134: Can we do this via the user controller? I think it's fine to use the simple GETs for different controllers.

tests/feedback.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    await new PortalState().createUsers([user, admin]).write(conn);
+
+    const submittedFeedbackResponse = await feedbackController.submitFeedback({ feedback }, user);
+    const { uuid } = submittedFeedbackResponse.feedback;
+    const ignoredFeedbackResponse = await feedbackController.updateFeedbackStatus(uuid, { status }, admin);
+
+    const persistedUser = await conn.manager.findOne(UserModel, { where: { uuid: user.uuid } });
+
+    expect(ignoredFeedbackResponse.feedback.status).toEqual(FeedbackStatus.IGNORED);
+    expect(persistedUser.points).toEqual(user.points);
```
sumeet-bansal, line 154: Same here about simple GETs vs direct queries.

tests/feedback.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const feedbackController = ControllerFactory.feedback(conn);
+    await new PortalState().createUsers([user, admin]).write(conn);
+    const feedbackToAcknowledgeResponse = await feedbackController.submitFeedback({ feedback: feedback1 }, user);
+    const feedbackToIgnoreResponse = await feedbackController.submitFeedback({ feedback: feedback2 }, user);
+
+    const acknowledgedFeedbackUuid = feedbackToAcknowledgeResponse.feedback.uuid;
+    const ignoredFeedbackUuid = feedbackToIgnoreResponse.feedback.uuid;
+    const acknowledgedStatus = { status: FeedbackStatus.ACKNOWLEDGED };
+    const ignoredStatus = { status: FeedbackStatus.IGNORED };
```
sumeet-bansal, line 171: nit: these can be inlined

types/HttpEnums.ts
sumeet-bansal: ^

sumeet-bansal (commented):
The thing about wrapped responses makes sense but IMO it's pretty minor/not worth the tradeoff of creating wrappers for every controller, and I think a test for a controller should deal directly with the controller's full response.

For the connection thing, `ControllerFactory` and `PortalState::write` could call `DatabaseConnection` themselves instead of requiring a connection as a parameter, but PortalStates were made to be used in tests so I don't think it's a problem that they're being used in tests. They were made to be flexible and to accommodate whatever scenario a test might need to setup, and I'm sure other tests are going to use them in weird and niche ways, so abstracting them away behind rigid functions for all possible test cases seem unsustainable: merch tests would require their own function for writing users and store things, sliding leaderboard tests would require their own function for writing users and events and attendances, etc. It'd be simpler and more readable to directly define the PortalState in the test.

tests/feedback.test.ts
shravanhariharan2: Yeah you're right, this assertion is pretty redundant here

tests/feedback.test.ts
shravanhariharan2: Yup - I think what we want to assert instead is that the scope/type are correct for feedback submission

shravanhariharan2: Also, slightly unrelated, do we want `PublicActivity` to return the `scope` as well? Since `PublicActivity` can now be either private or public depending on if it's fetched for the current user or a different one. Also we should probably rename `PublicActivity` since it's technically incorrect but still conventional since everything returned to the client is considered "public"

sumeet-bansal: I don't think `PublicActivity` needs to return `scope` since permissions are internal, but I don't have a strong opinion about it. The type name's probably fine to leave as is.

tests/controllers/ControllerFactory.ts
shravanhariharan2: I feel like it's alright to leave the controllers as singletons, just so that service/controller objects don't have to be re-created again every time a controller is invoked, even though that overhead is fairly small.

api/controllers/FeedbackController.ts
```diff
@@ -14,9 +13,12 @@ import {
 @UseBefore(UserAuthentication)
 @JsonController('/feedback')
 export class FeedbackController {
-  @Inject()
   private feedbackService: FeedbackService;
 
+  constructor(feedbackService: FeedbackService) {
```
sumeet-bansal, line 18: We don't need to inject the service in the constructor? I see in the typedi and routing-controllers docs that we don't need the `@Inject` if the class has the `@Service` annotation, but I don't see anything like that for controller annotations.

tests/feedback.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    const errors = await validate(plainToClass(Feedback, feedback));
+
+    expect(errors).toBeDefined();
+    expect(errors).toHaveLength(1);
+    expect(errors[0].property).toEqual('description');
+    expect(errors[0].constraints.minLength).toBeDefined();
+  });
+
+  test('has has proper activity scope and type', async () => {
```
sumeet-bansal, line 55:
```suggestion
  test('has proper activity scope and type', async () => {
```

tests/feedback.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    await new PortalState().createUsers([user1, user2, admin]).write(conn);
+
+    const feedbackController = ControllerFactory.feedback(conn);
+    const submittedFeedback1Response = await feedbackController.submitFeedback({ feedback: feedback1 }, user1);
+    const submittedFeedback2Response = await feedbackController.submitFeedback({ feedback: feedback2 }, user2);
+    const allSubmittedFeedbackResponse = await feedbackController.getFeedback(admin);
+
+    expect(allSubmittedFeedbackResponse.feedback).toHaveLength(2);
+    expect(allSubmittedFeedbackResponse.feedback).toEqual(
```
sumeet-bansal, line 84: Can we do a `toStrictEqual([...])` here (and in the other array assertions) or do we need to use the `arrayContaining`?

tests/controllers/ControllerFactory.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class ControllerFactory {
+  private static userController: UserController = null;
+
+  private static feedbackController: FeedbackController = null;
+
+  public static user(conn: Connection): UserController {
+    if (!ControllerFactory.userController) {
+      const userService = new UserAccountService(conn.manager);
+      const storageService = new StorageService();
```
sumeet-bansal, line 16: As an aside, we should prob pass in the S3 config to the `StorageService` constructor instead of importing.

sumeet-bansal (approved):

sumeet-bansal:
Approved but @michl1001 should review it too before merging.

michl1001:
I looked through the changes, _haven't tested it on my machine yet_ but so far I can't see anything wrong, in the comments for `feedback.test.ts`, Sumeet made a good point on checking arrays and idk if it was ever addressed. Orher than that, everything looks in order.

api/controllers/FeedbackController.ts
shravanhariharan2: Looks like we don't need them - in fact, including them throws an error when sending the controller a request:
`ServiceNotFoundError: Cannot determine a class of the requesting service \"undefined\"`

tests/feedback.test.ts
shravanhariharan2: Nope don't think so, since the database tables have indices on uuid, they are pulled in sorted order by uuid which is unpredictable (since feedback factory doesn't produce models)

tests/controllers/ControllerFactory.ts
shravanhariharan2: I personally think it's fine as is, just so that different services/controllers don't need to know about lower-level aws configs when instantiating the storage service, and so that aws doesn't need to get imported everywhere

PR #145, opened 2020-12-21
Remove References to Activity.public for Milestones + Bonuses

shravanhariharan2:
Removed some more references to public column in ActivityRepository for milestones + bonuses

### DIFF
--- a/repositories/ActivityRepository.ts
+++ b/repositories/ActivityRepository.ts
@@ -1,4 +1,4 @@
-import { EntityRepository } from 'typeorm';
+import { EntityRepository, Raw } from 'typeorm';
 import * as moment from 'moment';
 import { ActivityScope, ActivityType, Uuid } from '../types';
 import { UserModel } from '../models/UserModel';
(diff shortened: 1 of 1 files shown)


shravanhariharan2:
I'm going to force-merge this since it's a hotfix that is currently blocking census rewards for membership. I'll create a separate PR addressing all the other changes mentioned in #143 for the public -> scope activity change.

sumeet-bansal:
Are there any mislabeled Activity rows that need to be fixed in the database?

shravanhariharan2:
@sumeet-bansal If you mean for entries in the scope column, then no there shouldn't be. This query initially would error out so nothing would get written, and the other activity queries properly scope them as `PUBLIC` or `HIDDEN`.

sumeet-bansal:
Not incorrect scope, incorrect type, so bonus points mislabeled as milestones.

shravanhariharan2:
Oh no that was an issue within this PR from the first commit. There was no labeling at all before this PR, it was just wrongly using `public` instead of `scope`

PR #147, opened 2020-12-23
Add public user activity route and scope index 

shravanhariharan2:
Resolves issues in #143

### DIFF
--- a/api/controllers/UserController.ts
+++ b/api/controllers/UserController.ts
@@ -1,5 +1,5 @@
 import {
-  JsonController, Param, Get, Post, Patch, NotFoundError, UseBefore, UploadedFile, Body,
+  JsonController, Param, Get, Post, Patch, UseBefore, UploadedFile, Body,
 } from 'routing-controllers';
 import { Inject } from 'typedi';
 import { UserModel } from '../../models/UserModel';
@@ -12,6 +12,7 @@ import {
   Uuid,
   File,
   GetUserActivityStreamResponse,
+  GetCurrentUserActivityStreamResponse,
   UpdateProfilePictureResponse,
   GetUserResponse,
   GetCurrentUserResponse,
@@ -28,10 +29,22 @@ export class UserController {
   @Inject()
   private storageService: StorageService;
 
+  @Get('/:uuid/activity/')
+  async getUserActivityStream(@Param('uuid') uuid: Uuid,
+    @AuthenticatedUser() currentUser: UserModel): Promise<GetUserActivityStreamResponse> {
+    if (uuid === currentUser.uuid) {
+      return this.getCurrentUserActivityStream(currentUser);
+    }
+    const user = await this.userAccountService.findByUuid(uuid);
+    const activityStream = await this.userAccountService.getUserActivityStream(user.uuid);
+    return { error: null, activity: activityStream };
+  }
+
   @Get('/activity')
-  async getUserActivityStream(@AuthenticatedUser() user: UserModel): Promise<GetUserActivityStreamResponse> {
-    const stream = await this.userAccountService.getUserActivityStream(user.uuid);
-    return { error: null, activity: stream };
+  async getCurrentUserActivityStream(@AuthenticatedUser() user: UserModel):
+  Promise<GetCurrentUserActivityStreamResponse> {
+    const activityStream = await this.userAccountService.getCurrentUserActivityStream(user.uuid);
+    return { error: null, activity: activityStream };
   }
 
   @Post('/picture')
@@ -49,9 +62,6 @@ export class UserController {
       return this.getCurrentUser(currentUser);
     }
     const user = await this.userAccountService.findByUuid(uuid);
-    if (!user) {
-      throw new NotFoundError('User was not found');
-    }
     return { error: null, user: user.getPublicProfile() };
   }
 
--- a/api/controllers/UserController.ts
+++ b/api/controllers/UserController.ts
@@ -12,7 +12,6 @@ import {
   Uuid,
   File,
   GetUserActivityStreamResponse,
-  GetCurrentUserActivityStreamResponse,
   UpdateProfilePictureResponse,
   GetUserResponse,
   GetCurrentUserResponse,
@@ -41,7 +40,7 @@ export class UserController {
 
   @Get('/activity')
   async getCurrentUserActivityStream(@AuthenticatedUser() user: UserModel):
-  Promise<GetCurrentUserActivityStreamResponse> {
+  Promise<GetUserActivityStreamResponse> {
     const activityStream = await this.userAccountService.getCurrentUserActivityStream(user.uuid);
     return { error: null, activity: activityStream };
   }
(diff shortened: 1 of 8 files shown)


models/ActivityModel.ts
```diff
@@ -11,7 +11,7 @@ export class ActivityModel extends BaseEntity {
 
   @ManyToOne((type) => UserModel, (user) => user.activities, { nullable: false })
   @JoinColumn({ name: 'user' })
-  @Index('public_activities_by_user_index', { where: 'public IS true' })
+  @Index('public_activities_by_user_index', { where: 'scope = \'PUBLIC\'' })
   user: UserModel;
```
shravanhariharan2, line 15: Should this partial index be `PUBLIC or PRIVATE` or just `PUBLIC`? A user (after redesign) can only view other users' public activities, but can view their own public and private activities, so do we need to optimize for the latter?

sumeet-bansal: Maybe index on both `user` and `scope`?

migrations/0025-add-public-activity-scope-index.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationInterface, QueryRunner, TableIndex } from 'typeorm';
+
+export class AddActivityScopeIndex1608574428247 implements MigrationInterface {
```
sumeet-bansal, line 3: This is updating an existing index not creating a new one, so does this still work or do we need to drop the original index before creating the new one?

migrations/0025-add-public-activity-scope-index.ts
```diff
@@ -0,0 +1,16 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.createIndex('Activities',
+      new TableIndex({
+        name: 'public_activities_by_user_index',
+        columnNames: ['scope'],
```
sumeet-bansal, line 8: The current index is on the `user` column (groups all the activities by user).

migrations/0025-add-public-activity-scope-index.ts
```diff
@@ -0,0 +1,16 @@
...
+        where: 'scope = \'PUBLIC\'',
+      }));
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.dropIndex('Activities', 'public_activities_by_user_index');
```
sumeet-bansal, line 14: This should restore the old index.

models/ActivityModel.ts
sumeet-bansal: WHERE can be for all non-hidden activities.

repositories/ActivityRepository.ts
```diff
@@ -69,7 +69,7 @@ export class ActivityRepository extends BaseRepository<ActivityModel> {
   public async getEarliestTimestamp(): Promise<number> {
     const earliestPointsRecord = await this.repository.createQueryBuilder()
       .select('MIN("timestamp")', 'timestamp')
-      .where('scope = \'PUBLIC\' AND "pointsEarned" > 0')
+      .where('(scope = \'PUBLIC\' OR scope = \'PRIVATE\') AND "pointsEarned" > 0')
```
sumeet-bansal, line 72: WHERE can just be `"pointsEarned" > 0` here.

sumeet-bansal (changes requested):
@michl1001 can you also review this?

models/ActivityModel.ts
sumeet-bansal: Something like `visible_activities_by_user_index` would probably be a better index name if the WHERE's for non-hidden activities.

migrations/0025-add-public-activity-scope-index.ts
shravanhariharan2: When the `public` column was dropped in the previous migration, so was the original index

migrations/0025-add-public-activity-scope-index.ts
shravanhariharan2: This migration didn't actually drop the old index (the previous one did), so if we restore it here then if we rollback just this migration and not the last, there would be an error since the `public` column referenced by that old index doesn't exist

michl1001 (commented):
As far as I can tell, things seem right.

Just making sure, this is to make the "public" column be a string in stead of a boolean (presumedly to make more subcategories) - right?

sumeet-bansal:
@michl1001 yeah instead of `public: boolean` it's `scope: [PUBLIC;PRIVATE;HIDDEN]`.

@shravanhariharan2 Something that just occurred to me: activities can be public, but the activity stream endpoint only works for viewing your own activities, so there's no difference between public/private right now. Can you update the endpoint to be able to get another user's public activities? Then, the index should work for `scope=PUBLIC` and `scope=[PUBLIC;PRIVATE]`, so maybe it should be on `['user', 'scope'] WHERE scope = 'PUBLIC' or scope = 'PRIVATE'` (or `WHERE scope != 'HIDDEN'` if we might add more visible scopes in the future).

shravanhariharan2:
@sumeet-bansal ready for review

models/ActivityModel.ts
```diff
@@ -5,13 +5,13 @@ import { UserModel } from './UserModel';
 
 @Entity('Activities')
 @Index('sliding_leaderboard_index', ['timestamp', 'pointsEarned'], { where: '"pointsEarned" > 0' })
+@Index('visible_activities_by_user_index', ['user', 'scope'], { where: 'scope = \'PUBLIC\' OR scope = \'PRIVATE\'' })
```
sumeet-bansal, line 8: I think this index works for both `public` and `public+private` queries but I'm not completely sure. Can you verify the index is being used for both queries? You can look up how to do that; IIRC it involves looking at the query planning using `EXPLAIN`.

api/controllers/UserController.ts
```diff
@@ -28,10 +29,22 @@ export class UserController {
   @Inject()
   private storageService: StorageService;
 
+  @Get('/:uuid/activity/')
```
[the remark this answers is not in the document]
shravanhariharan2: Nope, doesn't seem to break any existing routes. Maybe routing-controllers handles pattern matching differently than vanilla express? Since `EventController` also has sub-routes that don't break existing ones

models/ActivityModel.ts
shravanhariharan2: yep, `public`, `private`, and `public+private` all are indexed. 

It also looks like when selecting any column, not just user or scope, the index gets used. Is there any reason we need to specify `['user', 'scope']` for partial column indexes then?

shravanhariharan2:
will bump version depending on which open PR gets merged first

models/ActivityModel.ts
sumeet-bansal: See the part about "order_nr" in the docs about partial indexes ([here](https://www.postgresql.org/docs/8.0/indexes-partial.html)). Without those columns, it'd still use the index but it'd scan the entire index looking for all rows matching a user instead of navigating to the part of the tree that has all rows for a user. Same for scope, if you wanted to get a user's public activities, you'd go to that part of the tree instead of scanning the entire index.

sumeet-bansal (approved):
`GetUserActivityStreamResponse` and `GetCurrentUserActivityStreamResponse` are identical types, so I think it'd be better to just use the former for both.

PR #148, opened 2021-01-10
Throw UserError when an event update reuses an attendance code

sumeet-bansal:
There was a QueryFailedError because an event attempted to reuse an existing attendance code. The event creation endpoint has an attendance code check, but the event update endpoint doesn't, so this PR adds that check to the event update endpoint.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@acmucsd/membership-portal",
-  "version": "2.5.0",
+  "version": "2.5.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@acmucsd/membership-portal",
-  "version": "2.5.0",
+  "version": "2.5.1",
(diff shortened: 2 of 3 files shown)


shravanhariharan2 (changes requested):
https://github.com/acmucsd/membership-portal-ui/blob/master/src/containers/admin/EditEventForm.tsx#L83

When editing events, UI passes in all event fields to our API, and not just changes, meaning `changes.attendanceCode` is always present. So if the attendance code doesn't change on edit, then this would throw the user error because the attendance code already exists for an event (namely, this current one).

IMO UI should change their request payload to only send the fields changed, since that's how our API works functionally, but it isn't really clear anywhere in our specification. But since this change would throw an error on every event update that doesn't change its attendance code, maybe we should have `isUnusedAttendanceCode` count the events that aren't the current one?

sumeet-bansal:
I'll take out all the unchanged fields in `changes` then.

sumeet-bansal:
@shravanhariharan2 Updated the condition to make sure the attendance code changed from the current one before checking that the new code's unique.

shravanhariharan2 (approved):

PR #149, opened 2021-01-13
Default to reverse order for past events endpoint

sumeet-bansal:
By request but also it just makes sense.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "typedi": "^0.8.0",
     "typeorm": "^0.2.25",
     "typeorm-typedi-extensions": "^0.2.3",
-    "typescript": "^3.9.5",
+    "typescript": "^4.0.0",
     "underscore": "^1.10.2",
     "uuid": "^8.3.1",
     "winston": "^3.3.3",
--- a/repositories/EventRepository.ts
+++ b/repositories/EventRepository.ts
@@ -49,7 +49,7 @@ export class EventRepository extends BaseRepository<EventModel> {
     let query = this.repository.createQueryBuilder()
       .skip(options.offset)
       .take(options.limit)
-      .orderBy('start', 'ASC');
+      .orderBy('start', options.reverse ? 'DESC' : 'ASC');
     if (options.committee) {
       query = query
         .where('committee = :committee')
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "typedi": "^0.8.0",
     "typeorm": "^0.2.25",
     "typeorm-typedi-extensions": "^0.2.3",
-    "typescript": "^3.9.5",
+    "typescript": "^4.0.0",
```
sumeet-bansal, line 64: This is to pick up the `??=` operator (see [here](https://github.com/microsoft/TypeScript/issues/37255)).

shravanhariharan2 (approved):

package.json
trulyronak: this is god tier ty for sharing

repositories/EventRepository.ts
```diff
@@ -49,7 +49,7 @@ export class EventRepository extends BaseRepository<EventModel> {
     let query = this.repository.createQueryBuilder()
       .skip(options.offset)
       .take(options.limit)
-      .orderBy('start', 'ASC');
+      .orderBy('start', options.reverse ? 'DESC' : 'ASC');
```
trulyronak, line 52: while this makes more sense, honestly lets just default to `ASC` to avoid a synchronized frontend push

trulyronak (approved):
lgtm, but i dont think we should make a breaking change / rely on frontend to change in sync (but idrc)

repositories/EventRepository.ts
trulyronak: instead for [402](https://github.com/acmucsd/membership-portal-ui/pull/402) the code change would simply be to pass in the `reverse` option in the request (so we don't waste time doing the reverse, but we avoid making a breaking change)

package.json
sumeet-bansal: @trulyronak check out the TypeScript roadmap ([here](https://github.com/Microsoft/TypeScript/wiki/Roadmap)) for all sorts of nifty things.

repositories/EventRepository.ts
sumeet-bansal: The ternary does default to `ASC` if `options.reverse` isn't set. And yeah, we should just pass it in as part of the request but I think it makes more sense for the past events endpoint to default to `DESC` anyways.

PR #153, opened 2021-01-16
Add get all emails admin route

### LINKED ISSUE #151: Add admin route for returning all portal user emails
Create a route for returning all portal user emails. Useful to frontend for creating auto-complete functionality for searching for users.

shravanhariharan2:
Closes #151

### DIFF
--- a/api/controllers/AdminController.ts
+++ b/api/controllers/AdminController.ts
@@ -1,4 +1,4 @@
-import { JsonController, Post, UploadedFile, UseBefore, ForbiddenError, Body } from 'routing-controllers';
+import { JsonController, Post, UploadedFile, UseBefore, ForbiddenError, Body, Get } from 'routing-controllers';
 import { Inject } from 'typedi';
 import { UserAuthentication } from '../middleware/UserAuthentication';
 import { CreateBonusRequest, CreateMilestoneRequest } from '../validators/AdminControllerRequests';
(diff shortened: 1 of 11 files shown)


steets250 (approved):
Tested on Postman and works correctly.

sumeet-bansal (commented):
LGTM but I think this endpoint might better fit in an `AdminController`. What do you think? Also can you bump the version?

shravanhariharan2:
@sumeet-bansal my thinking was that the route name `/user/email` made more sense intuitively than something like `/admin/email`, and the functionality like viewing members emails (with potential suborg filters after user roles are implemented) could be useful for future roles other than just admins.

Also, do we want to return all members emails, or only return those emails where `user.state !== BLOCKED`? And similar question for milestones, since right now it can grant all users points, even potentially blocked ones.

sumeet-bansal:
I think of the user controller as anything that's got to do with specific accounts, like getting/updating a profile, so getting all emails doesn't really fit into that since it's more of an admin action.

sumeet-bansal:
As for returning blocked emails, I think this would be used for accounts in good standing but I also don't know enough to speculate on how the route might be used. Maybe let FE filter to whatever it wants? Or for now, default to excluding blocked users.

shravanhariharan2:
@steets250 any preferences for returning all emails vs only emails "in good standing" (i.e. where `UserStatus !== BLOCKED` ?

Also, do we want to implement a way for blocking members in the portal anytime soon? Blocked status currently exists in the BE but there's no actual functionality for blocking users.

steets250:
Returning every user regardless of status should be fine. The goal was to make it easier to award bonus points by allowing an admin to search from a drop down rather than typing them in and having email matching issues. I don't think a member who has a blocked status is likely to be rewarded bonus points, but that decision should be left up to Casey or the membership team.

shravanhariharan2:
@sumeet-bansal @michl1001 ready for review (can't re-request while awaiting previous request)

sumeet-bansal (approved):

PR #154, opened 2021-01-19
Add retroactive attendance admin route

### LINKED ISSUE #152: Add admin/attendance route for retroactively submitting attendance for a user
Create an admin route for attendance that allows an admin to check-in user(s) (by their email(s)) to an event either during or after that event.

shravanhariharan2:
Closes #152

### DIFF
--- a/services/AttendanceService.ts
+++ b/services/AttendanceService.ts
@@ -3,7 +3,7 @@ import { InjectManager } from 'typeorm-typedi-extensions';
 import { NotFoundError } from 'routing-controllers';
 import { EntityManager } from 'typeorm';
 import * as moment from 'moment';
-import { ActivityType, PublicAttendance, Uuid } from '../types';
+import { ActivityType, PublicAttendance, PublicEvent, Uuid } from '../types';
 import { Config } from '../config';
 import { UserModel } from '../models/UserModel';
 import { UserError } from '../utils/Errors';
@@ -31,7 +31,7 @@ export default class AttendanceService {
     return attendances.map((attendance) => attendance.getPublicAttendance());
   }
 
-  public async attendEvent(user: UserModel, attendanceCode: string, asStaff = false): Promise<PublicAttendance> {
+  public async attendEvent(user: UserModel, attendanceCode: string, asStaff = false): Promise<PublicEvent> {
     return this.transactions.readWrite(async (txn) => {
       const event = await Repositories.event(txn).findByAttendanceCode(attendanceCode);
       if (!event) throw new NotFoundError('Oh no! That code didn\'t work.');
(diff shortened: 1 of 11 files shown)


services/AttendanceService.ts
```diff
@@ -51,8 +51,36 @@ export default class AttendanceService {
...
+    });
+  }
+
+  public async attendEventRetroactively(userUuid: Uuid, eventUuid: Uuid): Promise<PublicEvent> {
+    return this.transactions.readWrite(async (txn) => {
+      const attendanceRepository = Repositories.attendance(txn);
+
+      const event = await Repositories.event(txn).findByUuid(eventUuid);
+      if (!event) throw new UserError('This event doesn\'t exist');
+      if (!event.hasStarted()) throw new UserError('This event hasn\'t started yet');
```
sumeet-bansal, line 65: Why this constraint? Could an event host potentially want to check someone in ahead of time?

services/AttendanceService.ts
```diff
@@ -51,8 +51,36 @@ export default class AttendanceService {
...
+
+      const user = await Repositories.user(txn).findByUuid(userUuid);
+      if (!user) throw new NotFoundError('This user was not found');
+
+      const hasAlreadyAttended = await attendanceRepository.hasUserAttendedEvent(user, event);
+      if (hasAlreadyAttended) throw new UserError('This user has already attended this event');
+
+      const pointsEarned = event.pointValue;
+      const attendedAsStaff = false;
```
sumeet-bansal, line 74: Why this constraint? Could an event host potentially want to check someone in as staff, e.g. they helped with pre-work/setup but couldn't attend the actual event?

shravanhariharan2: IMO this route shouldn't be used for things like pre-checkin / checking in as a staff if they didn't actually attend the event. This could bloat things like membership points or attendance head count if there's a bunch of people who weren't actually present but still counted for attendance. It's more meant for people who attended late, couldn't check, and who actually need the attendance as a requirement for something (e.g. Bit-Byte requires >= 2 attendances to bit-byte events). Either way, membership / ACM board should probably be the ones enforcing these constraints and not the backend so I'll remove them.

services/AttendanceService.ts
```diff
@@ -37,21 +39,43 @@ export default class AttendanceService {
...
+      attendedAsStaff ? ActivityType.ATTEND_EVENT_AS_STAFF : ActivityType.ATTEND_EVENT,
+      pointsEarned,
+    );
+    await Repositories.user(txn).addPoints(user, pointsEarned);
+
+    const attendance = await Repositories.attendance(txn).attendEvent(user, event, attendedAsStaff);
+    return attendance;
+  }
+
+  public async submitAttendanceForUser(userUuid: Uuid, eventUuid: Uuid, asStaff = false): Promise<PublicAttendance> {
```
sumeet-bansal, line 66: Do you think it'd be worthwhile to track somewhere that this attendance was made by another user? e.g. in the Activity comment or something

services/AttendanceService.ts
```diff
@@ -37,21 +39,43 @@ export default class AttendanceService {
...
+
+    await Repositories.activity(txn).logActivity(
+      user,
+      attendedAsStaff ? ActivityType.ATTEND_EVENT_AS_STAFF : ActivityType.ATTEND_EVENT,
+      pointsEarned,
+    );
+    await Repositories.user(txn).addPoints(user, pointsEarned);
+
+    const attendance = await Repositories.attendance(txn).attendEvent(user, event, attendedAsStaff);
+    return attendance;
```
sumeet-bansal, line 63:
```suggestion
    return Repositories.attendance(txn).attendEvent(user, event, attendedAsStaff);
```

sumeet-bansal (approved):

services/AttendanceService.ts
shravanhariharan2: Yeah good idea, I'll add a short activity comment for those attendances

PR #163, opened 2021-02-23
Create more robust database seeds

### LINKED ISSUE #132: Add more robust database seeds
Add database seeds for activities, orders, and attendances to allow for a better and more realistic testing experience locally

sumeet-bansal:
Closes #132.

> Our goal is to define some small system from which most actions are testable,
e.g. checking into events (past/future, already-attended, ongoing), ordering
merch (out of stock, insufficient credits, lifetime limit, successful), getting
a decent-sized sliding leaderboard.
>
> Users: an admin, a couple committees (Hack, AI) and their accompanying
staff, and a handful of members.
>
> Events: some staffed committee workshops, unstaffed general/committe socials,
and staffed industry events, in the past, present, and future.
>
> There're also some "edge cases" here and there inspired by real data, e.g.
admins granting bonus points to members that forgot/were unable to check-in.

TODO
+ add bonuses
+ add retroactive check-in


sumeet-bansal:
@shravanhariharan2 @michl1001 @steets250 thoughts so far? Basically defined a bunch of users,  past/ongoing/future events, merch, and attendances for past/ongoing events, so you can test most actions from the current state. Anything I should add seeds for? Or anything seem useless?

shravanhariharan2:
I really like the setup, especially the fact that all points for all users are earned organically rather than having some people start off with a base amount of points. I think that'll make testing activities, sliding leaderboard, and other point-related things a lot easier. A few other comments:
* I think there should be a few ongoing events created that have no check-ins, so that if someone logs into any of the seeded accounts, they have a few events they can check into rather than having to pick a certain account that hasn't checked into a specific event
* Something I noticed in `PortalState::attendEvents()` - the `asStaff` should only apply if `user.isStaff() && event.requiresStaff`, but right now it applies regardless, so people who aren't staff or people who attend events that don't need staff would still get logged as attending as staff in the seeds
* With merch order seeds, you can do a similar thing where all users place orders for a few items but there are some items that no user has ordered
* Retroactive attendance - is that something we need to seed? Looks like `PortalState::attendEvents` already supports retroactive attendance (perhaps unintentionally since it doesn't check if the event has passed or not), so I think we should change that method to check in only current events and create a retroactive attendance method that follows the behavior of `AttendanceService::submitAttendanceForUser`

sumeet-bansal:
1. Added an ongoing event with no check-ins.
2. This is intentional.
3. There're no seeded merch orders because merch orders depend on the events attended for a user, so it's easy to introduce impossible states with seeded merch orders, e.g. we accidentally add a merch order for a user with insufficient credits.
4. I don't think we need that here, but we can do it in a follow-up if someone asks.

tests/Seeds.ts
```diff
@@ -0,0 +1,489 @@
...
+    firstName: 'Andrea',
+    lastName: 'Sudharta',
+    graduationYear: getGraduationYear(2),
+  });
+  const MEMBER_SENIOR = UserFactory.fake({
+    email: 's3bansal@ucsd.edu',
+    accessType: UserAccessType.STANDARD,
+    firstName: 'Sumeet',
+    graduationYear: getGraduationYear(0),
+  });
```
shravanhariharan2, line 84: super nit: graduation years are going from 4 to 3 to 2 to 0, I think they should either be (3,2,1,0) or (4,3,2,1) for consistency (though I'm pretty sure it doesn't matter which)

shravanhariharan2 (approved):
LGTM, I don't think `npm run db:seed:rollback` is used in any cron job / schedule anywhere but I'll double check on that. It'd only be on the testing API so you can merge either way

tests/Seeds.ts
sumeet-bansal: It's ordered (freshman, sophomore, junior, senior), so `getGraduationYear` takes years left and returns the current year minus years left.

PR #165, opened 2021-02-26
Accept multiple users for retroactive attendance

### LINKED ISSUE #164: Accept multiple emails for retroactive attendance route
The retroactive attendance route currently submits attendance for one user at a time, but the route should be able to accept multiple users as input so that clients don't need to hit the route multiple times.

shravanhariharan2:
Closes #164

### DIFF
--- a/repositories/ActivityRepository.ts
+++ b/repositories/ActivityRepository.ts
@@ -30,11 +30,28 @@ export class ActivityRepository extends BaseRepository<ActivityModel> {
     const activity = {
       user,
       type,
-      description,
       pointsEarned,
+      description,
       scope: ActivityRepository.activityScopes[type],
     };
     return this.repository.save(ActivityModel.create(activity));
+  }
+
+  public async batchLogActivity(
+    users: UserModel[], types: ActivityType[], pointsEarnedBatch?: number[], descriptionBatch?: string[],
+  ): Promise<ActivityModel[]> {
+    const activities: ActivityModel[] = [];
+    for (let i = 0; i < users.length; i += 1) {
+      const activity = {
+        user: users[i],
+        type: types[i],
+        pointsEarned: pointsEarnedBatch[i],
+        description: descriptionBatch[i],
+        scope: ActivityRepository.activityScopes[types[i]],
+      };
+      activities.push(ActivityModel.create(activity));
+    }
+    return this.repository.save(activities);
   }
 
   public async logMilestone(description: string, pointsEarned: number): Promise<void> {
--- a/repositories/ActivityRepository.ts
+++ b/repositories/ActivityRepository.ts
@@ -4,26 +4,10 @@ import { ActivityScope, ActivityType, Uuid } from '../types';
 import { UserModel } from '../models/UserModel';
 import { ActivityModel } from '../models/ActivityModel';
 import { BaseRepository } from './BaseRepository';
+import { Activity, ActivityTypeToScope } from '../types/internal';
 
 @EntityRepository(ActivityModel)
 export class ActivityRepository extends BaseRepository<ActivityModel> {
-  private static activityScopes = {
-    [ActivityType.ACCOUNT_CREATE]: ActivityScope.PUBLIC,
-    [ActivityType.ATTEND_EVENT]: ActivityScope.PUBLIC,
-    [ActivityType.ATTEND_EVENT_AS_STAFF]: ActivityScope.PUBLIC,
-    [ActivityType.BONUS_POINTS]: ActivityScope.PUBLIC,
-    [ActivityType.MILESTONE]: ActivityScope.PUBLIC,
-    [ActivityType.FEEDBACK_ACKNOWLEDGED]: ActivityScope.PRIVATE,
-    [ActivityType.ORDER_MERCHANDISE]: ActivityScope.PRIVATE,
-    [ActivityType.SUBMIT_EVENT_FEEDBACK]: ActivityScope.PRIVATE,
-    [ActivityType.SUBMIT_FEEDBACK]: ActivityScope.PRIVATE,
-    [ActivityType.ACCOUNT_ACTIVATE]: ActivityScope.HIDDEN,
-    [ActivityType.ACCOUNT_LOGIN]: ActivityScope.HIDDEN,
-    [ActivityType.ACCOUNT_RESET_PASS]: ActivityScope.HIDDEN,
-    [ActivityType.ACCOUNT_RESET_PASS_REQUEST]: ActivityScope.HIDDEN,
-    [ActivityType.ACCOUNT_UPDATE_INFO]: ActivityScope.HIDDEN,
-  };
-
   public async logActivity(
     user: UserModel, type: ActivityType, pointsEarned?: number, description?: string,
   ): Promise<ActivityModel> {
@@ -32,30 +16,18 @@ export class ActivityRepository extends BaseRepository<ActivityModel> {
       type,
       pointsEarned,
       description,
-      scope: ActivityRepository.activityScopes[type],
+      scope: ActivityTypeToScope[type],
     };
     return this.repository.save(ActivityModel.create(activity));
   }
 
-  public async batchLogActivity(
-    users: UserModel[], types: ActivityType[], pointsEarnedBatch?: number[], descriptionBatch?: string[],
-  ): Promise<ActivityModel[]> {
-    const activities: ActivityModel[] = [];
-    for (let i = 0; i < users.length; i += 1) {
-      const activity = {
-        user: users[i],
-        type: types[i],
-        pointsEarned: pointsEarnedBatch[i],
-        description: descriptionBatch[i],
-        scope: ActivityRepository.activityScopes[types[i]],
-      };
-      activities.push(ActivityModel.create(activity));
-    }
-    return this.repository.save(activities);
+  public async logActivityBatch(activities: Activity[]): Promise<ActivityModel[]> {
+    const activityModels = activities.map((activity) => ActivityModel.create(activity));
+    return this.repository.save(activityModels);
   }
 
   public async logMilestone(description: string, pointsEarned: number): Promise<void> {
-    const scope = ActivityRepository.activityScopes[ActivityType.MILESTONE];
+    const scope = ActivityTypeToScope[ActivityType.MILESTONE];
     return this.repository.query(
       'INSERT INTO "Activities" ("user", "type", "description", "pointsEarned", "scope") '
       + `SELECT uuid, '${ActivityType.MILESTONE}', '${description}', '${pointsEarned}', '${scope}' `
@@ -64,7 +36,7 @@ export class ActivityRepository extends BaseRepository<ActivityModel> {
   }
 
   public async logBonus(users: UserModel[], description: string, pointsEarned: number): Promise<void> {
-    const scope = ActivityRepository.activityScopes[ActivityType.BONUS_POINTS];
+    const scope = ActivityTypeToScope[ActivityType.BONUS_POINTS];
     const uuids = users.map((user) => `'${user.uuid}'`);
     return this.repository.query(
       'INSERT INTO "Activities" ("user", "type", "description", "pointsEarned", "scope") '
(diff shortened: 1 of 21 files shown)


services/AttendanceService.ts
```diff
@@ -63,22 +63,30 @@ export default class AttendanceService {
...
-    admin: UserModel): Promise<PublicAttendance> {
+  public async submitAttendanceForUsers(emails: string[], eventUuid: Uuid, asStaff = false,
+    admin: UserModel): Promise<PublicAttendance[]> {
     return this.transactions.readWrite(async (txn) => {
       const event = await Repositories.event(txn).findByUuid(eventUuid);
       if (!event) throw new NotFoundError('This event doesn\'t exist');
 
-      const user = await Repositories.user(txn).findByUuid(userUuid);
-      if (!user) throw new NotFoundError('This user was not found');
+      const users = await Repositories.user(txn).findByEmails(emails);
```
sumeet-bansal, line 72: `Map<String, Maybe<UserModel>>` might be a more useful data type for `UserRepository::findByEmails` to return.

```typescript
const usersByEmails: Map<String, Maybe<UserModel>> = await Repositories
  .user(txn)
  .findByEmails(emails);
const emailsNotFound: string[] = usersByEmails.entries()
  .filter([_email, u] => u.isNone())
  .map([email, _u] => email);
if (emailsNotFound.length > 0) {
  throw new BadRequestError('Couldn\'t find accounts matching one or more emails', {
    emails: emailsNotFound,
  });
}
```

sumeet-bansal: I can't actually find an official `Maybe<T>` type though.

sumeet-bansal: There's an `Option<T>` in the [`fp-ts`](https://www.npmjs.com/package/fp-ts) package that we can use.

sumeet-bansal: I guess email is part of `UserModel` though, so we could just do
```typescript
const users: UserModel[] = await Repositories.user(txn).findByEmails(emails);
const foundEmails: string[] = users.map(u => u.email);
const emailsNotFound = emails.filter(e => !foundEmails.contains(e));
```
but I like the idea of returning a type that's keyed by the search params.

sumeet-bansal: Lmk what you think.

services/AttendanceService.ts
```diff
@@ -63,22 +63,30 @@ export default class AttendanceService {
     return Repositories.attendance(txn).attendEvent(user, event, attendedAsStaff);
   }
 
-  public async submitAttendanceForUser(userUuid: Uuid, eventUuid: Uuid, asStaff = false,
-    admin: UserModel): Promise<PublicAttendance> {
+  public async submitAttendanceForUsers(emails: string[], eventUuid: Uuid, asStaff = false,
+    admin: UserModel): Promise<PublicAttendance[]> {
```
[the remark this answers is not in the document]
shravanhariharan2: I agree, there could potentially be other roles that can submit attendances

services/AttendanceService.ts
shravanhariharan2: I also like the idea of returning the keyed param type. I think another way we can do it within SQL is running a query like:

```sql
SELECT emails.id, "Users".* FROM "Users" RIGHT JOIN (VALUES ('email1@ucsd.edu'), ('email2@ucsd.edu'), ('email3@ucsd.edu')) AS emails(id) ON "Users".email = emails.id;
```
which will return all the emails present in the `emails` array mapped to a user if it exists, otherwise null. Thoughts on this?

services/AttendanceService.ts
```diff
@@ -63,22 +63,30 @@ export default class AttendanceService {
...
+        throw new BadRequestError('Couldn\'t find accounts matching one or more emails');
+      }
 
-      const activityDescription = `Attendance submitted by user ${admin.uuid}`;
-      const attendance = await this.writeEventAttendance(user, event, asStaff, txn, activityDescription);
-      return attendance.getPublicAttendance();
+      return Promise.all(users.map(async (user) => {
+        const attendanceRepository = Repositories.attendance(txn);
+        const hasAlreadyAttendedEvent = await attendanceRepository.hasUserAttendedEvent(user, event);
+        if (hasAlreadyAttendedEvent) {
```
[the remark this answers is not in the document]
shravanhariharan2: Looking back I think it's probably better to just skip instead of throwing, since the service should just flag everyone as attended, and if they already attended, then no work needs to be done for that user

services/AttendanceService.ts
sumeet-bansal: IMO we should keep the query simple and do the mapping in the code. The goal of using an `Option<T>` type is to avoid nulls and have a type specifically for "this might or might not exist" since nulls can be finicky to work with and cause weird errors, so we'd want to return `Option.none()` instead of `null`.

tests/admin.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('retroactive attendance submission', () => {
+  test('logs activity, attendance, and points for users who have not attended', async () => {
+    const conn = await DatabaseConnection.get();
+    const [user1, user2] = UserFactory.create(2);
+    const emails = [user1, user2].map((user) => user.email);
+    const [onBehalfOfUser] = UserFactory.with({ accessType: UserAccessType.ADMIN });
```
sumeet-bansal, line 23: I know I suggested it but `onBehalfOf` is a bit misleading since the caller's doing it on behalf of other users. Apparently the legal term would be "per procuration" but that's too obscure for our purposes. Maybe let's just stick with `user` and rename `user1` and `user2` to `onBehalfOf1` and `onBehalfOf2`.

tests/admin.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    expect(activityResponse1.activity).toHaveLength(2);
+    expect(activityResponse1.activity[1].pointsEarned).toEqual(event.pointValue);
+    expect(activityResponse1.activity[1].type).toEqual(ActivityType.ATTEND_EVENT);
+    expect(activityResponse1.activity[1].scope).toEqual(ActivityScope.PUBLIC);
+
+    expect(activityResponse2.activity).toHaveLength(2);
+    expect(activityResponse2.activity[1].pointsEarned).toEqual(event.pointValue);
+    expect(activityResponse2.activity[1].type).toEqual(ActivityType.ATTEND_EVENT);
+    expect(activityResponse2.activity[1].scope).toEqual(ActivityScope.PUBLIC);
```
sumeet-bansal, line 62: This can go in a `for` loop.

sumeet-bansal (changes requested):

tests/admin.test.ts
```diff
@@ -0,0 +1,91 @@
...
+});
+
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('retroactive attendance submission', () => {
+  test('logs activity, attendance, and points for users who have not attended', async () => {
+    const conn = await DatabaseConnection.get();
```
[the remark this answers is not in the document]
shravanhariharan2: Think you mentioned it earlier in the feedback testing PR but it slipped through, my bad. 

With this change the one thing that's fishy to me is having to await the factory fetch (e.g. `const adminController = await ControllerFactory.admin()`) for every time a controller is used, which doesn't make as much sense when reading it and which means you can't do something like `const response = await ControllerFactory.admin().submitAttendanceForUsers(...);`, since it needs 2 awaits (unless we use `.then` and `.catch` which breaks our codebase's conventions)

tests/admin.test.ts
shravanhariharan2: I think the noun form of [proxy](https://www.lexico.com/definition/proxy) ("a person authorized to act on behalf of another") is a possible word, though it might be confused with the CS meaning of it. I'd rather rename just the admin instead of all the users if we can

repositories/ActivityRepository.ts
```diff
@@ -30,13 +30,30 @@ export class ActivityRepository extends BaseRepository<ActivityModel> {
...
+      description,
       scope: ActivityRepository.activityScopes[type],
     };
     return this.repository.save(ActivityModel.create(activity));
   }
 
+  public async batchLogActivity(
+    users: UserModel[], types: ActivityType[], pointsEarnedBatch?: number[], descriptionBatch?: string[],
+  ): Promise<ActivityModel[]> {
+    const activities: ActivityModel[] = [];
```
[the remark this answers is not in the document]
shravanhariharan2: My thinking was to have `batchLogActivity` match `logActivity` with arrays, but yeah it ends up being less clean. I added an internal `Activity` and `Attendance` type to pass in, should we also change the arguments of `logActivity` and `attendEvent` to take in that type rather than multiple args?

services/AttendanceService.ts
```diff
@@ -48,38 +48,72 @@ export default class AttendanceService {
...
-      if (!user) throw new NotFoundError('This user was not found');
+      const users = await Repositories.user(txn).findByEmails(emails);
+      const emailsFound = users.map((user) => user.email);
+      const emailsNotFound = emails.filter((email) => !emailsFound.includes(email));
 
-      const attendanceRepository = Repositories.attendance(txn);
-      const hasAlreadyAttended = await attendanceRepository.hasUserAttendedEvent(user, event);
-      if (hasAlreadyAttended) throw new UserError('This user has already attended this event');
+      if (emailsNotFound.length > 0) {
+        throw new BadRequestError(`Couldn't find accounts matching these emails: ${emailsNotFound}`);
```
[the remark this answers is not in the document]
shravanhariharan2: Yeah makes sense, but we'd need to define specific error types / constructors for that in a different PR like you mentioned since we can't pass in a body to `HttpError` (it just accepts a `message`)

tests/admin.test.ts
```diff
@@ -0,0 +1,83 @@
...
+});
+
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('retroactive attendance submission', () => {
+  test('logs activity, attendance, and points for users who have not attended', async () => {
+    const users = UserFactory.create(3);
```
sumeet-bansal, line 20: Maybe make one of these users a staff member to test staff check-in as well.

tests/admin.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      .createEvents([event])
+      .write();
+
+    const adminController = await ControllerFactory.admin();
+    const userController = await ControllerFactory.user();
+    const attendanceController = await ControllerFactory.attendance();
+
+    await adminController.submitAttendanceForUsers({ users: emails, event: event.uuid }, proxyUser);
+
+    await Promise.all(users.map(async (user) => {
```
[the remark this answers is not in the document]
shravanhariharan2: Just tried a few ways of iterating over and it seems like `Promise.all` is the only one that works currently:
* `for/of` throws a linter error to not use `for/of` (no-restricted-syntax)
* just `for` throws linter errors for no-await-in-loop
* `users.forEach(async (user) => {...});` fails the test cases because for some reason a `NotFoundError` is being thrown saying that a user is not found. Currently have no clue why this is happening - I tried reading the users table from `conn.manager.find` and all the data is there.
* `Promise.all` works just fine, with the one downside that the order is non-deterministic

shravanhariharan2:
Do we want to keep the single-object functions' parameters consistent with the batch function parameters? E.g. 
```ts
public async logActivity(user: UserModel, type: ActivityType, pointsEarned?: number, description?: string) {}
// to
public async logActivity(activity: Activity) {}
// so that it matches
public async logActivityBatch(activities: Activity[]) {}
```
or is it fine to have different param structures for single vs. batch

tests/admin.test.ts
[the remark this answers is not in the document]
sumeet-bansal: We could disable `no-await-in-loop` for tests in the linter config (set a pattern match for file names `*.test.ts` or something similar). I personally find the `Promise.all` pretty weird and prefer looping, but it's your call.

sumeet-bansal:
Since we have the `Activity` object, might as well use it and clean up the method signature.

tests/admin.test.ts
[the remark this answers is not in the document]
shravanhariharan2: Yeah agreed `Promise.all` is somewhat sketch in this situation since it's just a hacky way of looping through elements to pass the linter

shravanhariharan2:
@sumeet-bansal since there's also the `Attendance` object now (and potentially other internal types we can create) I'll do a larger parameter refactor in a different PR since this one is getting quite large already. Everything else should be ready for review

sumeet-bansal (approved):

PR #171, opened 2021-04-01
Implement event cover resizing

### LINKED ISSUE #108: Setup Bucket w/Thumbnailing for Event Cover Photos
All images are current stored through Imgur - need to port over to s3. Need to also implement a way for [thumbnailing to reduce load size on ](https://docs.aws.amazon.com/lambda/latest/dg/with-s3-example.html)page render

Front end needs to implement lazy loading

shravanhariharan2:
Closes #108 

Implemented a function to resize event cover images to 960 x 540 (as discussed earlier). Turns out it's much easier to implement it locally rather than having a Lambda function that interacts directly with S3, without much performance loss too (takes <1s to resize + upload + send response to a event cover request)


shravanhariharan2:
What should we do about having maximum file sizes now that we have automatic resizing (at least for event covers)? Should we keep that restraint? Relax it a bit? Get rid of it? @steets250 @sumeet-bansal thoughts on what we should do?

sumeet-bansal:
How does the events team produce event covers? Do they have fixed dimensions for every cover? Do covers have some expected file size?

I'm a little hesitant to automatically resize images in a black box; another option might be to push those decisions to the user, who has more context about the image, e.g. offer a checkbox to resize images to some specified dimensions. And at that point it might be easier to do the resizing locally in the UI and keep the backend limits.

sumeet-bansal:
Also, quick nit: this would be a feature not a bug fix so the version bump should be to 2.9.0 not 2.8.3. I can update the auto-comment bot with some guidance if you think that'd be helpful.

sumeet-bansal:
Thinking about this some more, I think the backend should continue to reject images based on file size, and instead of the backend resizing images by itself the UI should help the user edit the image as needed to get under whatever limits the backend enforces. That makes more sense to me in terms of UI vs backend responsibilities.

steets250:
I can let events know that they should scale the covers down before posting. They're always the same size (1920 x 1080), so it should be easy for them to just produce another smaller size for the portal.

shravanhariharan2:
> How does the events team produce event covers? Do they have fixed dimensions for every cover? Do covers have some expected file size?

As of now (and Steven can correct me if I'm wrong here), all event covers are designed on Figma with the same exact width/height. Then, once the image is ready, the event admin fills in the event form on the portal, attaches the image, and sends the request to us

The reason we initially were thinking of resizing in the backend is because there is a set width/height that we wanted all covers to be sized to so that the amount of data needing to be loaded is minimized. The resizing functionality isn't necessarily so that the events team can get the image size below a threshold that we defined, it's more that there is a set size that all covers should conform to and the resizing functionality automates that. 

In terms of having that functionality in the UI vs backend, my thoughts are that the events team should at least know that their covers have to be a set size (perhaps through some UI message, or like Steven mentioned, letting them know that it should be that size), but since it is a set size, the backend can automate that resizing before uploading it.

sumeet-bansal:
Ok I see, I guess resizing's fine if we know all event covers are always the same ratio.

I still think the UI should handle resizing: since it wants to define some fixed height/width for images, it should also do the resizing to whatever dimensions it wants, plus that whole process can be transparent to the user, e.g. it resizes and then maybe gets the sign-off from the user before uploading. Like, someone shouldn't upload a cover, download it, and wonder why it's not the same image.

But otherwise, the PR LGTM and if y'all still think resizing in the backend's the way to go, I'll approve.

sumeet-bansal:
Whoa what's up with the lockfile diff? Looks like the entire thing was rewritten after upgrading to `lockfileVersion: 2`.

shravanhariharan2:
Hmm yeah that's a large diff. Apparently lockfile version 2 is created when an npm version >= 7 is used (which I must've recently updated on my end), and it has a [different / better structure than lockfile v1](https://nitayneeman.com/posts/catching-up-with-package-lockfile-changes-in-npm-v7/) which is why the diff's huge. We should probably stick with this newer lockfile version since npm version 5/6 is considered legacy now.

shravanhariharan2:
@sumeet-bansal we've decided to stick with backend for resizing (see discussion in #portal-backend), so this PR is ready for review. We also changed the way resizing fits (also see discussion) such that the image is guaranteed to a) maintain aspect ratio and b) be >= 960 x 540 pixels.

shravanhariharan2:
Closing due to resolution of discussion on Discord that the client should handle implementing functionality specific to itself (as @sumeet-bansal mentioned above)

PR #172, opened 2021-04-06
Refactor repository parameters to single types 

### LINKED ISSUE #167: Refactor repository parameters
Now that we have some internal types made for batch routes, we should explore refactoring many of the repository methods with high parameter counts to use internal types to reduce the number of params passed

shravanhariharan2:
Closes #167

### DIFF
--- a/repositories/ActivityRepository.ts
+++ b/repositories/ActivityRepository.ts
@@ -8,21 +8,16 @@ import { Activity, ActivityTypeToScope } from '../types/internal';
 
 @EntityRepository(ActivityModel)
 export class ActivityRepository extends BaseRepository<ActivityModel> {
-  public async logActivity(
-    user: UserModel, type: ActivityType, pointsEarned?: number, description?: string,
-  ): Promise<ActivityModel> {
-    const activity = {
-      user,
-      type,
-      pointsEarned,
-      description,
-      scope: ActivityTypeToScope[type],
-    };
+  public async logActivity(activity: Activity): Promise<ActivityModel> {
+    activity.scope = ActivityTypeToScope[activity.type];
     return this.repository.save(ActivityModel.create(activity));
   }
 
   public async logActivityBatch(activities: Activity[]): Promise<ActivityModel[]> {
-    const activityModels = activities.map((activity) => ActivityModel.create(activity));
+    const activityModels = activities.map((activity) => {
+      activity.scope = ActivityTypeToScope[activity.type];
+      return ActivityModel.create(activity);
+    });
     return this.repository.save(activityModels);
   }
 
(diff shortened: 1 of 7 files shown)


shravanhariharan2:
@sumeet-bansal @michl1001 I just started with the activity repository change that we discussed [here](https://github.com/acmucsd/membership-portal/pull/165#issuecomment-794565431) but I'm noticing that there's a bit of duplication that emerges, since the creation of the activity object is pushed up to the service level so a bunch of services need to create the object themselves now.

Not sure if we want to continue refactoring params or not since this would likely be the case for the rest after refactoring the other repository methods so I'd like to hear both your thoughts on it

sumeet-bansal:
Can you give an example?

shravanhariharan2:
I think it's more the fact that every 1 line of `logActivity()` that used to exist is now replaced with 3-4 lines since the object now is passed in for every call to `logActivity()` (for example,  [here](https://github.com/acmucsd/membership-portal/blob/884899249fe4de4fb4f4a420848bb58330747e9c/services/AttendanceService.ts#L57), [here](https://github.com/acmucsd/membership-portal/blob/884899249fe4de4fb4f4a420848bb58330747e9c/services/AttendanceService.ts#L151), and [here](https://github.com/acmucsd/membership-portal/blob/884899249fe4de4fb4f4a420848bb58330747e9c/services/MerchStoreService.ts#L272) ), so it feels a bit longer/more verbose. But I do like that now the activity fields are explicitly there (which make it easier to see what's being passed in) instead of having to guess what field is being passed based on its order in the parameter list / its type.

sumeet-bansal:
Oh that's fine. Like you said, it's easier to see what's being passed in.

sumeet-bansal (approved):
LGTM. Nits:
+ Might as well inline the activity types instead of doing `type: activityType`.
+ If `pointsEarned: 0` can we just not include that in the object? We do that elsewhere and it should just default to 0.

shravanhariharan2:
@sumeet-bansal are we doing patch bumps for refactors like this? Or should they be a minor bump like features? I recall you mentioning earlier that we can add more detail in the auto comment bot about this and I don't think I responded there, but yeah that would be great.

shravanhariharan2:
re: inlining activity types- for more complex types that are dependent on things like `attendedAsStaff`, I think it's better to not inline since that's what we're doing with `pointsEarned` for those cases too. I'll change the ones that can only hold 1 value to be inline

sumeet-bansal:
> @sumeet-bansal are we doing patch bumps for refactors like this? Or should they be a minor bump like features? I recall you mentioning earlier that we can add more detail in the auto comment bot about this and I don't think I responded there, but yeah that would be great.

Don't need to bump versions for refactors. I'll update the auto-comment bot.

sumeet-bansal:
> re: inlining activity types- for more complex types that are dependent on things like `attendedAsStaff`, I think it's better to not inline since that's what we're doing with `pointsEarned` for those cases too. I'll change the ones that can only hold 1 value to be inline

`pointsEarned` shouldn't be inlined since it needs to be reused for creating attendance and adding points, but activity type is never reused and can be inlined.

Issue #140, opened 2020-11-09
validate route parameters


sumeet-bansal:
Inspired by
```
Nov 08 20:42:45 acmucsd-membership-portal-api app/web.1 warn: 2020-11-09T04:42:45.558Z [request cc194ab68495de09305ea3e883adf55b]: QueryFailedError [500]: invalid input syntax for type uuid: "test" 
Nov 08 20:42:45 acmucsd-membership-portal-api app/web.1 Mon, Nov 9, 2020 4:42 AM PT [IP 172.251.122.132] [Trace cc194ab68495de09305ea3e883adf55b] GET /api/v2/user/test 500 34.405ms
```

shravanhariharan2:
Resolved via. https://github.com/acmucsd/membership-portal/pull/173

PR #173, opened 2021-04-07
Validate Route Parameters

michl1001:
- Addresses #140.

- Param (uuid, email, accessCode) validation should be working - haven't gotten a chance to test it yet)

- Still writing the route param classes


sumeet-bansal:
Looks like you have some merge conflicts and a lot of the PR diffs aren't from this PR. Try merging `master` and pushing; that should clean up this PR a bit.

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "eslint ./ --ext .ts",
     "lint:fix": "eslint ./ --ext .ts --fix",
     "db:migrate:create": "ts-node ./node_modules/.bin/typeorm migration:create -n",
-    "db:migrate": "ts-node ./node_modules/.bin/typeorm migration:run",
```
sumeet-bansal, line 18: What's this change for?

types/ApiParams.ts
```diff
@@ -0,0 +1,17 @@
+import { IsEmail, IsUUID, IsHexadecimal, Length } from 'class-validator';
+import { Uuid } from '.';
+
+export class ValidEmail{
```
sumeet-bansal, line 4: Since these classes are for validating the API requests, they belong in `/api/validators/GenericRequests.ts`, not in the `types` folder. `types` is for defining API interfaces, like specifications that an API implementation should conform to, but these validators are part of the API implementation not its specification. Can explain further it that doesn't make sense.

api/requestParams/AuthRequestParams.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    @IsEmpty()
+    authenticatedUser?:any;
+    
+    @IsNotEmpty()
+    @IsDefined()
+    requestTrace?:string;
+}
+
+export class resendEmailVerificationRequestParam implements request{
```
sumeet-bansal, line 53: Nit: classes are uppercase, and these aren't validating request params they're validating the entire request.

```suggestion
export class ResendEmailVerificationRequest implements Request {
```

sumeet-bansal: These classes should also go in `/api/validators`.

sumeet-bansal: Also tbh I haven't thought much about validating entire requests but I imagine we'd want something like
```typescript
export class ResendEmailVerificationRequest implements IResendEmailVerificationRequest {}
```
Probably worth discussing further in Discord before we make more changes, since validating entire requests is gonna add significantly more types.

sumeet-bansal: For now, can we move these changes in `/api/requestParams` to another branch to keep this PR's scope manageable? This PR can just be the `ValidUuid/Email/etc` validators.

package.json
shravanhariharan2: Michael was running into issues on Windows 10 (and for some reason, WSL too) while running `npm run db:migrate` and `npm run db:seed` (the same error message appearing [in this issue](https://github.com/typeorm/typeorm/issues/2316)), and this change seemed to fix that while still working on my end.

I told him to try it out on WSL2 and see if that works, but yeah it might be one of those many issues with Windows 10,  not too sure.

sumeet-bansal: oof should've guessed Windows things

michl1001:
Alright! Made the changes in Sumeet's comments so the scope is much more reasonable. Also should passing the lint test now.

The orginal code is in the branch `RouteRequestClass`.

api/controllers/AuthController.ts
```diff
@@ -18,21 +19,19 @@ import { authActionMetadata } from '../../utils/AuthActionMetadata';
...
 @JsonController('/auth')
 export class AuthController {
+  @Inject()
   private userAccountService: UserAccountService;
 
+  @Inject()
   private userAuthService: UserAuthService;
 
+  @Inject()
   private emailService: EmailService;
```
shravanhariharan2, line 34: Why's the `@Inject()` decorator added back? Are these merged changes from earlier or are they required for the controller now? 

I think we removed those (and added the constructor in below lines) back in an earlier PR for getting testing setup, so this change would break most of the test suites

api/controllers/AuthController.ts
```diff
@@ -52,32 +51,32 @@ export class AuthController {
   }
 
   @Get('/emailVerification/:email')
-  async resendEmailVerification(@Param('email') email: string): Promise<ResendEmailVerificationResponse> {
-    const user = await this.userAuthService.setAccessCode(email.toLowerCase());
+  async resendEmailVerification(@Params() vEmail: ValidEmail): Promise<ResendEmailVerificationResponse> {
+    const user = await this.userAuthService.setAccessCode(vEmail.email.toLowerCase());
     await this.emailService.sendEmailVerification(user.email, user.firstName, user.accessCode);
     return { error: null };
```
shravanhariharan2, line 57: nit: the parameter name `vEmail` (and the other renamed parameters) seems somewhat cryptic/less readable and can probably be renamed to follow the same pattern as request bodies (i.e. having the same name as its type, so something like `emailParams`).

shravanhariharan2: Actually, @sumeet-bansal is there any particular reason we have the body as the same name as its type? Since both are the same, we could just have it be `body` (or just `params` in this case) since the type keeps it clear what its holding. That way the function definition/calls are a bit less verbose

api/validators/GenericRequests.ts
```diff
@@ -8,3 +8,18 @@ export class Pagination implements IPagination {
...
+export class ValidUuid {
+  @IsUUID()
+  uuid:Uuid;
+}
+
+export class ValidAccessCode {
+  @IsHexadecimal()
+  @Length(32, 32)
+  accessCode:string;
+}
```
shravanhariharan2, line 25: nit: with our validation checks now I think it's implied that all uuids/emails/accessCodes that are passed in are "valid", so we don't need to mention that within the class name. Something like `EmailParam` or `UuidParam` might make it more clear the intention of the classes

package.json
```diff
@@ -15,7 +15,7 @@
...
+    "db:migrate": "ts-node ./node_modules/typeorm/cli.js migration:run",
     "db:rollback": "ts-node ./node_modules/.bin/typeorm migration:revert",
     "db:seed:create": "ts-node ./node_modules/.bin/typeorm migration:create -c seed -n",
     "db:seed": "ts-node ./node_modules/.bin/typeorm migration:run -c seed",
```
shravanhariharan2, line 21: can you also change these to use `./node_modules/typeorm/cli.js` so that they would work Windows 10 machines? The new path seems to work on Windows + WSL (and thus Linux), and most likely on Mac.

shravanhariharan2 (changes requested):
Mostly nitpicks but also one change that does break our test suite/script. Also, make sure everything works functionally too once you have your environment fully setup (and let me know if you're still running into any issues on the WSL 2 setup)

api/controllers/AuthController.ts
sumeet-bansal: Since we switched from `@Param('uuid')` to `@Params()`, that variable no longer represents just the UUID but all the request params, so `@Params() params` would be a more fitting name.

But then the issue becomes, `@Params() params` is still only the validated UUID/email/etc since that's what the class is defined as, and we'd have to create another class like `@Params() params: GetAttendancesForEventQueryParams`.

So if we want to use `@Params()` everywhere, that becomes a lot of work, since we'd have to do that for every controller method. Probably easiest to just stick with `@Param('email') email: Email` for the time being.

sumeet-bansal: Yeah, we could rename the variables `body`. I personally prefer using the type to name the variable because `resendEmailVerificationRequest` seems like a better name to me, but `body` works too, as long as we're consistent.

api/controllers/AuthController.ts
michl1001: Oooop, I must have accidently kept them when merging, my b

michl1001:
@shravanhariharan2 

So, te test that failed was because we internally call functions that pass uuid as a string since that is how it is stored for users

Should I change the user ORM or should I just cast the uuids where we call the function?

michl1001:
Wait, nevermind, the test suite just needs to change how uuid is passed in

api/validators/GenericRequests.ts
```diff
@@ -8,3 +8,18 @@ export class Pagination implements IPagination {
   @Min(0)
   limit?: number;
 }
+
+export class EmailParam {
+  @IsEmail()
+  email:string;
+}
```
shravanhariharan2, line 15: nit: leave a space between the key and value in an object for consistency and readability: `email: string;` vs. `email:string`

tests/admin.test.ts
```diff
@@ -36,15 +36,18 @@ describe('retroactive attendance submission', () => {
 
     for (let u = 0; u < users.length; u += 1) {
       const user = users[u];
-      const userResponse = await userController.getUser(user.uuid, proxyUser);
+      const userUuid = {
+        uuid: user.uuid,
+      };
+      const userResponse = await userController.getUser(userUuid, proxyUser);
```
shravanhariharan2, line 42: nit: you can inline a lot of these variable declarations to reduce LoC since they aren't used that often and aren't _that_ cryptic when inlined:
```js
const userResponse = await userController.getUser({ uuid: user.uuid }, proxyUser)
```

shravanhariharan2 (approved):
Functionality looks good, just a few comments for consistency.

shravanhariharan2:
Another note on consistency - for PR names, we usually use imperative-style messages (e.g. "Allow user bio to be empty on update"), so can you change the name to match that style?

api/validators/GenericRequests.ts
sumeet-bansal: I'm surprised this wasn't caught by the linter.

api/controllers/AttendanceController.ts
```diff
@@ -17,10 +18,10 @@ export class AttendanceController {
   }
 
   @Get('/:uuid')
-  async getAttendancesForEvent(@Param('uuid') uuid: Uuid,
+  async getAttendancesForEvent(@Params() params: UuidParam,
```
sumeet-bansal, line 21: @shravanhariharan2 thoughts on using `@Params()` vs `@Param('uuid')`? I think we should stick with the singular if possible because otherwise if we want to add multiple params, e.g. an endpoint that accepts a UUID and pagination, we'd need some `UuidAndPaginationParams` object and etc. for all combinations. But I don't see a way to validate `@Param('uuid')`. Maybe it's nbd?

tests/feedback.test.ts
```diff
@@ -140,11 +145,16 @@ describe('feedback submission', () => {
     const feedbackController = ControllerFactory.feedback(conn);
 
     const submittedFeedbackResponse = await feedbackController.submitFeedback({ feedback }, user);
-    const { uuid } = submittedFeedbackResponse.feedback;
+    const { uuid } = {
+      uuid: submittedFeedbackResponse.feedback,
+    };
```
sumeet-bansal, line 150: This should be inlined as well, but
```suggestion
    const uuid = submittedFeedbackResponse.feedback;
```

sumeet-bansal:
> Another note on consistency - for PR names, we usually use imperative-style messages (e.g. "Allow user bio to be empty on update"), so can you change the name to match that style?

"Validate route parameters" or something along those lines.

api/controllers/AttendanceController.ts
shravanhariharan2: I definitely prefer singular `Param('uuid')` over `Params()` since that kind of injection is the main benefit of routing-controllers, but yeah it doesn't look like `Param('uuid')` can be validated with class-validator. Not sure if there's any other route we can take like having a global middleware that checks if a uuid is passed in and validates it that way? Though that seems a bit more convoluted. 

If nothing else looks promising then I think it's nbd either, better to have a proper form of API validation at the expense of potentially more interfaces than no validation at all.

michl1001: Additionaly, if we do make an encapsulation of all parts of each request, these parameter class will be used like this lmore down the line

sumeet-bansal:
🚀 🚀 🚀

shravanhariharan2:
Congrats again on your first PR merge @michl1001 !!! The first of many 😎

PR #178, opened 2021-04-12
Change auth type from bearer token to cookies

### LINKED ISSUE #176: Return JWT Token via cookie
For user authorization, if the backend returns the JWT token via cookie, rather than via the response body, it simplifies token storage on frontend. In addition, if the cookie domain is set to `.acmucsd.com`, it would allow for auth across multiple sites, if at some point we have multiple web apps that require signing in.

shravanhariharan2:
Closes #176

### DIFF
--- a/api/controllers/AuthController.ts
+++ b/api/controllers/AuthController.ts
@@ -1,4 +1,5 @@
-import { JsonController, Param, Body, Get, Post, UseBefore } from 'routing-controllers';
+import { JsonController, Param, Body, Get, Post, UseBefore, Res } from 'routing-controllers';
+import { Response } from 'express';
 import {
   RegistrationResponse,
   LoginResponse,
@@ -44,9 +45,14 @@ export class AuthController {
   }
 
   @Post('/login')
-  async login(@Body() loginRequest: LoginRequest, @RequestTrace() trace: string): Promise<LoginResponse> {
+  async login(
+    @Body() loginRequest: LoginRequest,
+      @RequestTrace() trace: string,
+      @Res() res: Response,
+  ): Promise<LoginResponse> {
     const user = await this.userAuthService.checkCredentials(loginRequest.email.toLowerCase(), loginRequest.password);
     const token = UserAuthService.generateAuthToken(user);
+    res.cookie('token', token);
     log.info('user authentication (login)', authActionMetadata(trace, user));
     return { error: null, token };
   }
--- a/api/middleware/UserAuthentication.ts
+++ b/api/middleware/UserAuthentication.ts
@@ -10,9 +10,10 @@ export class UserAuthentication implements ExpressMiddlewareInterface {
   private userAuthService: UserAuthService;
 
   async use(request: express.Request, response: express.Response, next?: express.NextFunction) {
-    const authHeader = request.get('Authorization');
-    if (!authHeader) throw new ForbiddenError('Missing auth token');
-    request.user = await this.userAuthService.checkAuthToken(authHeader);
+    const { token } = request.cookies;
+    console.log('authToken', token);
+    if (!token) throw new ForbiddenError('Missing auth token');
+    request.user = await this.userAuthService.checkAuthToken(token);
     log.info('user authentication (middleware)', authActionMetadata(request.trace, request.user));
     return next();
   }
--- a/index.ts
+++ b/index.ts
@@ -1,6 +1,9 @@
 import 'reflect-metadata'; // this shim is required
 
-import { createExpressServer, useContainer as routingUseContainer } from 'routing-controllers';
+import * as express from 'express';
+import * as cookieParser from 'cookie-parser';
+
+import { useExpressServer, useContainer as routingUseContainer } from 'routing-controllers';
 
 import { createConnection, useContainer as ormUseContainer } from 'typeorm';
 import { Container } from 'typedi';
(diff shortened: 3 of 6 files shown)


sumeet-bansal:
If we want to support cookies, I think we should do it alongside tokens in the `Authorization` header. If we exclusively use cookies, that breaks auth for side projects that call the portal (which have a `herokuapp.com` domain not `acmucsd.com`) and means the portal doesn't work when cookies are blocked (e.g. in incognito).

Tbh I also (maybe irrationally) find it kinda weird for an API to be setting cookies; is setting/using the cookie something that could be handled in the frontend instead? Also curious @steets250, what's the issue with using tokens in the frontend?

I've heard using cookies is tricky with mobile apps, and if we're thinking about auth for multiple web apps that require signing in, should we also be keeping mobile apps in mind? Don't recall if it was Dimensions but I remember there was some interest in mobile app development at some point.

steets250:
I agree that maintaining support for localStorage / some other local token option is a good idea just for compatibility with projects. Cookies do work in incognito, the same way that localStorage does, where everything is cleared when the window/session is closed. (To the best of my knowledge), cookies (especially auth-related ones), are meant to be created and sent by the backend. When cookies are marked as `HttpOnly`, it prevents the frontend from knowing that they exist and makes the site more secure from potential frontend vulnerabilities (although the exact types I don't know). In terms of mobile apps, by keeping the direct token option open, that should provide an alternate path to mobile apps in the case that cookies can't be implemented.

For implementing cookies, something like...

```js
const generateToken = async (res: any, username: string, remember: boolean): Promise<any> => {
    const expiration = testing ? 86400000 : 604800000;
    const token = jwt.sign({ username }, process.env.JWT_SECRET, {
        expiresIn: testing ? '1d' : '7d',
    });

    const properties = {
        expires: new Date(Date.now() + expiration),
        secure: testing ? false : true,
        httpOnly: true,
        domain: testing ? 'localhost' : 'acmucsd.com',
    };

    if (!remember) {
        delete properties.expires;
    }

    return res.cookie('token', token, properties);
};
```

...generates a secure cookie that expires after a set time. And then for logging out...

```js
const clearToken = async (res): Promise<any> => {
    const properties = {
        expires: new Date(0),
        secure: testing ? false : true,
        httpOnly: true,
        domain: testing ? 'localhost' : 'acmucsd.com',
    };

    return res.cookie('token', '', properties);
};
```

... causes the cookie to be deleted by the browser.

If you feel like the benefits of using cookies are outweighed by the code/time needed to implement them, we can stick with just storing in localstorage, but implementing these would help with cleaning the frontend up a bit and additionally help with having multiple `.acmucsd.com` sites.

shravanhariharan2:
Is there anything wrong about having both cookies and token as authentication like Sumeet mentioned? I can't currently think of any besides just additional complexity/checks in the backend. The biggest issue I see with pure cookie-based auth is the backwards-incompatibility with all of our current clients.

Seems like a hybrid approach would work just fine, where our API checks if the `Authorization` header was provided, and if not, if a cookie with the auth token is provided. Both would have the same expiration/config, but this way the client can decide how it wants to authenticate with us by including the `Authorization` header or not. That way we're still backwards-compatible with the current clients/side-projects while also allowing the multiple `*.acmucsd.edu` clients authenticate with the same token by using cookies.

steets250:
@shravanhariharan2 sounds like a solid plan to me.

shravanhariharan2:
> Tbh I also (maybe irrationally) find it kinda weird for an API to be setting cookies since tokens are more universal; is setting/using the cookie something that could be handled in the frontend instead?

I also found it weird at first but yeah it seems like the convention is to manage cookies within the backend so that any client that interacts with it does not have to worry about auth functionality

sumeet-bansal:
I meant I find it weird for the API to allow multiple ways to authorize when tokens can be used by any client. The frontend auth handling seems to only be setting/getting the token with local storage in the auth actions, so I don't see how it simplifies the frontend that much, but I don't have a strong opinion on this. Using the same cookie across the `acmucsd.com` domain makes sense.

shravanhariharan2:
Been reading up and thinking about which option (token vs hybrid vs cookie) would be the best but it looks like all of them have downsides. I looked a bit at [this article](https://dzone.com/articles/cookies-vs-tokens-the-definitive-guide) comparing the two, but to recap:
* "Only token" downsides: cannot use same login across multiple subdomains of `*.acmucsd.com`, _slightly_ more work needed in frontend (but not that much)
* "Token or cookie" downsides: backend is more complicated / less conventional than it needs to be since 2 forms of auth are used, which adds a bit of complexity and redundancy since tokens alone can do the job
* "Only cookie" downsides: mobile app auth setup is harder, slightly more security vulnerability (mentioned in the above article) but not sure how big of an issue this is considering that we're all college students and that our data isn't really sensitive. Then again our API has been abused before (the whole email deal back in person) so maybe that's something we still want to look out for?

From what most online sources are saying, token storage via. localhost seems to be the most common and I personally think that's the way to go it's the simplest and least restrictive (i.e. every client can use it). Making users login again on a different `.acmucsd.com` domain IMO doesn't seem like that big of an issue since it's a different website anyway, but let me know your guys' thoughts.

sumeet-bansal:
Agreed, I'm still fully on the "token only" train. For cookies, I see the appeal of not having to log in across multiple subdomains, but that doesn't really seem like a big deal to me. Maybe we can revisit this when there're actually multiple ACM sites that users are logging in and out of?

shravanhariharan2:
Sounds good - we discussed this a bit yesterday in our backend meeting and pretty much reached the same consensus. If more flaws with our current auth setup show up after more clients are introduced, then we can look back into some of these alternatives.

PR #180, opened 2021-04-18
Allow user bio to be empty on update

shravanhariharan2:
Previously users wouldn't be able to clear their bios after updating them since our validation didn't accept empty strings, so this change allows empty bios to be sent.


shravanhariharan2:
Looks like `v2.8.3` was already published after it was created in the event cover resizing PR, how should we go about reverting published versions that don't make it to master?

shravanhariharan2:
Never mind, I can just remove the `v2.8.3` tag and `npm version patch` will work as intended

sumeet-bansal (approved):

PR #181, opened 2021-04-24
Write events test suite

### LINKED ISSUE #175: Write event controller test suite
Assertions we should test:
* Fetching events
    * GET / returns all lifetime events
    * GET /past returns only the events that have already ended
    * GET /future returns only the events that have not started 
* Writing events
    * POST / creates an event if the attendance code provided is unused
    * POST / throws an error if a duplicate attendance code is provided
    * PUT / properly updates changed fields
* Event covers
    * POST /picture/:uuid updates the event's `cover` field
    * POST /picture/:uuid updates the S3 bucket on AWS with a properly named file
* Event feedback
    * Can be persisted and rewarded points when submitted for an event already attended
    * Is rejected on submission to an event not attended
    * Is rejected if sent multiple times to the same event
    * Is rejected if sent after 2 days of the event ending

shravanhariharan2:
Closes #175


tests/data/EventFactory.ts
```diff
@@ -24,21 +24,45 @@ export class EventFactory {
 
   public static fake(): EventModel {
     const [start, end] = EventFactory.randomTime();
+    return EventFactory.fakeWithTime(start, end);
+  }
+
+  private static fakeWithTime(start: Date, end: Date): EventModel {
```
sumeet-bansal, line 30: You can use the `with` method instead of `fakeWithTime`, e.g.
```typescript
private static fakeWithTime(start: Date, end: Date): EventModel {
  return with({ start, end });
}
```

tests/data/EventFactory.ts
```diff
@@ -24,21 +24,45 @@ export class EventFactory {
...
+    const event = EventFactory.fakeWithoutTime();
+    event.start = start;
+    event.end = end;
+    return event;
+  }
+
+  public static fakeWithoutTime(): EventModel {
```
sumeet-bansal, line 37: What's the different between this and `fake`? Or does this method exist because of `fakeWithTime`?

tests/data/EventFactory.ts
```diff
@@ -48,7 +72,41 @@ export class EventFactory {
     const duration = FactoryUtils.getRandomNumber(30, 150, 30);
     const start = moment().subtract(days, 'days').hour(hour);
     const end = moment(start.valueOf()).add(duration, 'minutes');
-    return [new Date(start.valueOf()), new Date(end.valueOf())];
+    return [start.toDate(), end.toDate()];
+  }
+
+  private static randomPastTime(daysAgo: number): [Date, Date] {
```
sumeet-bansal, line 78: Can we reuse code between these `randomTime` methods? Maybe we have optional params where we can set some `days`, `hour`, and `duration` fields?

tests/data/FeedbackFactory.ts
```diff
@@ -19,6 +19,10 @@ export class FeedbackFactory {
     };
   }
 
+  public static createEventFeedback(n: number): string[] {
+    return new Array(n).fill(faker.random.word());
```
sumeet-bansal, line 23: Should this be a random word or some value from a predefined array of values? IIRC feedback's supposed to be some predefined emojis.

tests/event.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    const [admin] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+
+    await new PortalState()
+      .createEvents([event1])
+      .createUsers([admin])
+      .write();
+
+    const createEventResponse = await ControllerFactory.event(conn).createEvent({ event: event2 }, admin);
+
+    expect(createEventResponse.error).toBeNull();
```
sumeet-bansal, line 73: Let's create both events through the controller here.

tests/event.test.ts
```diff
@@ -0,0 +1,225 @@
...
+  });
+
+  test('events cannot be created with duplicate attendance codes', async () => {
+    const conn = await DatabaseConnection.get();
+    const [event1] = EventFactory.with({ attendanceCode: 'code' });
+    const [event2] = EventFactory.with({ attendanceCode: 'code' });
+    const [admin] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+
+    await new PortalState()
```
sumeet-bansal, line 82: Same here; let's use the event controller.

tests/event.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    );
+  });
+
+  test('events can be created with unused attendance codes', async () => {
+    const conn = await DatabaseConnection.get();
+    const [event1] = EventFactory.with({ attendanceCode: 'code' });
+    const [event2] = EventFactory.with({ attendanceCode: 'different-code' });
+    const [admin] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+
+    await new PortalState()
```
sumeet-bansal, line 66: I think we should prefer controllers wherever we can instead of relying on `PortalState`. Like here, the test's simple enough that we could create via the controller.

sumeet-bansal (changes requested):
Looks good overall. I think we should mock `StorageService` instead of actually writing/reading to an S3 bucket (and definitely not to the same folder that prod uses), and we should start passing in our configs to constructors to make it easier to mock things.

michl1001 (approved):
Looks good overall though I'm not as intimately familiar with our test suite (whether there are better functions to use) and the higher-level design

services/EventService.ts
```diff
@@ -18,7 +18,7 @@ export default class EventService {
   public async create(event: Event) {
     const eventCreated = await this.transactions.readWrite(async (txn) => {
       const eventRepository = Repositories.event(txn);
-      const isUnusedAttendanceCode = eventRepository.isUnusedAttendanceCode(event.attendanceCode);
+      const isUnusedAttendanceCode = await eventRepository.isUnusedAttendanceCode(event.attendanceCode);
```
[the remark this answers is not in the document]
shravanhariharan2: That was caught with test - I saw that the duplicate attendance code error was thrown by Postgres' duplicate key constraint and not by the event service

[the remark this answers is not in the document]
sumeet-bansal: Nice, our first bug caught by testing!

tests/data/EventFactory.ts
shravanhariharan2: Totally forgot about `with()` - we actually don't even need `fakeWithTime` or `fakeWithoutTime` anymore if `with()` is used for `fakePastEvent(), fakeOngoingEvent(),` and `fakeFutureEvent()`.

tests/data/EventFactory.ts
shravanhariharan2: As mentioned above, this method is not needed anymore

shravanhariharan2: But the idea before was to have `fake, fakePastEvent, fakeOngoingEvent, fakeFutureEvent` to all call `fakeWithTime`, but specifying different times for each section. `fakeWithTime` would then call `fakeWithoutTime` to get a random object for all other fields

tests/data/FeedbackFactory.ts
shravanhariharan2: I'm not sure the predefined emojis have been confirmed yet since a lot of the design materials are moving around right now, and membership hasn't acknowledged them either since the portal redesign has been pushed back significantly (till after the store frontend is finished). I think it's fine to keep them at random words for now and then adjust later once the set of emojis /text is confirmed.

tests/event.test.ts
shravanhariharan2: Hmm, I personally think it's clearer to use `PortalState` for mocking all the data reads/writes that aren't explicitly tested, just so that it is clear what exactly is being tested just by looking at the controller calls that happen within the test. That way we are also reducing the number of dependent controllers that a given test is using so that if a test fails, it is clear that the controllers that are actually being tested are the ones failing (even though like you mentioned, the controller would be used for a simple case).

The one issue I'm seeing with `PortalState` is that it needs to keep being updated to match our API's functionality, but there aren't that many huge changes to the base functionality (like attending events, ordering merch) anyway to warrant a consistent updating of `PortalState`.

tests/event.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    const user = UserFactory.fake();
+
+    await new PortalState()
+      .createEvents([pastEvent, ongoingEvent, futureEvent])
+      .createUsers([user])
+      .write();
+
+    const eventController = ControllerFactory.event(conn);
+
+    const pastEventsResponse = await eventController.getPastEvents({}, user);
```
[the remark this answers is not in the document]
shravanhariharan2: How come? I personally find it clearer to have all the assertions at the bottom so that it's easier to see what exactly is being asserted instead having assertions being tied within controller calls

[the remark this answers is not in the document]
sumeet-bansal: In general, it makes more sense to me to immediately assert instead of collecting all the responses so tests can fail faster and more obviously, e.g. if `getFutureEvents` throws then we fail and don't need to unnecessarily call `getAllEvents`. Also it isn't the case in this test but one call might influence the return value of another, so we should assert ASAP and make sure everything's correct at all times.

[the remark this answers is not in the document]
sumeet-bansal: Neither reason's particularly applicable here but I think it's still worth sticking to.

tests/event.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    const [event] = EventFactory.create(1);
+    const [admin] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+    const cover = FileFactory.image(Config.file.MAX_EVENT_COVER_FILE_SIZE / 2);
+
+    await new PortalState()
+      .createUsers([admin])
+      .createEvents([event])
+      .write();
+
+    const eventCoverResponse = await ControllerFactory.event(conn).updateEventCover(cover, event.uuid, admin);
```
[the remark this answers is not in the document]
shravanhariharan2: I'm not too sure what we should pass in to the `StorageService` constructor, since I believe most of our config (such as bucket to be written to) is unchangeable. The `S3_BUCKET, S3_REGION`, and `S3_CREDENTIALS` are all constant through the entire app, and there's separate buckets for local (`acmucsd-local`, for what is used for locally running the test suite and in CircleCI jobs), testing (`acmucsd-membership-portal`, for testing.api.acmucsd.com), and prod (`acmucsd`, for what is used in production), so tests don't actually get written to our prod bucket.

tests/utils/StorageUtils.ts
```diff
@@ -0,0 +1,40 @@
+import * as aws from 'aws-sdk';
+import { Config } from '../../config';
+
+export class StorageUtils {
```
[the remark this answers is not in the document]
shravanhariharan2: Makes sense, I don't think we need to test that `aws-sdk` works. We can probably just get away with asserting that the URL returned has put it in the right bucket? Though that might also be testing `aws-sdk` functionality.

Not sure if you remember the one bug we had a few months ago where profile pic uploads were uploading to the wrong folder, but ideally we want to catch stuff like that by asserting the directory uploaded to. Let me know what you think we should assert

tests/data/FileFactory.ts
```diff
@@ -0,0 +1,24 @@
+import { File } from '../../types';
+
+const ONE_BYTE_FILLER = 'a';
+
+// Factory for mocking in-memory files for testing with multer
+export class FileFactory {
```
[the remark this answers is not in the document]
shravanhariharan2: Right, mocking should be good enough.

shravanhariharan2:
@sumeet-bansal how do you think we should go about mocking `StorageService` and passing in config into the constructor? To my understanding we can't just wrap that service like we do with controllers since the service would still upload to S3 and I don't think we wanted that, and I'm still not really sure what we'd pass in to the constructor of `StorageService` since all the config is the same except the bucket, which is changed within the environment variables.

sumeet-bansal:
What do you mean by wrapping controllers? Jest should have some sort of mocking functionality where we can do something like
```typescript
const storageService = mock(IStorageService.class);
when(storageService.upload(any(), any(), eq(MediaType.EVENT_COVER), eq("some-uuid")).thenReturn("fileLocation");
```
(That's the syntax for the Mockito library in Java.)

But the idea is, we create a "mock" and then we can define custom responses based on the parameters, and instead of executing normally, the mock returns our defined response.

sumeet-bansal:
This looks like what we need ([link](https://jestjs.io/docs/mock-functions#mocking-modules)).

sumeet-bansal:
Mock `StorageService` not the S3 library.

shravanhariharan2:
Wait, why mock `StorageService`? Isn't that something we'd want to actually test, but just not have anything uploaded? My thinking was that we don't want anything uploaded, so we just replace the `this.s3.upload` call with a different function that doesn't upload anything, but that still returns the URL that would be returned by `aws-sdk`.

shravanhariharan2:
I might just be generally confused of the concept/use case of mocking cause I think I've been the word throwing around here and there with the controllers without knowing if that's what we're actually doing with them.

sumeet-bansal:
No you're right, we could mock `aws-sdk`. `S3Mock` looks more difficult than it needs to be but maybe that's just because of jest. There isn't much of a difference between mocking `aws-sdk` and `StorageService` functionality-wise (we wouldn't verify the upload path but I don't think that's a big deal), so I figured it'd be easier to just mock `StorageService` instead.

shravanhariharan2:
> S3Mock looks more difficult than it needs to be but maybe that's just because of jest.

Yeah it's like that since that's how jest does mocking for modules. If we were to mock StorageService instead, then all we can assert is just that `event.cover` exists right? Instead of being able to check that event covers are uploaded to the event path instead of banner path, etc. (since that was also a bug that happened earlier).

sumeet-bansal:
1. Yikes, maybe we should look into a better library for mocking.
2. We could separately test the static method that takes an enum and returns the upload info.

services/EventService.ts
```diff
@@ -61,7 +61,7 @@ export default class EventService {
       const currentEvent = await eventRepository.findByUuid(uuid);
       if (!currentEvent) throw new NotFoundError('Event not found');
       if (changes.attendanceCode !== currentEvent.attendanceCode) {
-        const isUnusedAttendanceCode = eventRepository.isUnusedAttendanceCode(changes.attendanceCode);
+        const isUnusedAttendanceCode = await eventRepository.isUnusedAttendanceCode(changes.attendanceCode);
         if (!isUnusedAttendanceCode) throw new UserError('Attendance code has already been used');
```
shravanhariharan2, line 65: image

tests/controllers/index.ts
```diff
@@ -1 +1,148 @@
...
+import AttendanceService from '../../services/AttendanceService';
+import { AttendanceController } from '../../api/controllers/AttendanceController';
+import { AuthController } from '../../api/controllers/AuthController';
+import { EventController } from '../../api/controllers/EventController';
+import { LeaderboardController } from '../../api/controllers/LeaderboardController';
+import { MerchStoreController } from '../../api/controllers/MerchStoreController';
+import UserAuthService from '../../services/UserAuthService';
+import EmailService from '../../services/EmailService';
+import EventService from '../../services/EventService';
+import MerchStoreService from '../../services/MerchStoreService';
```
sumeet-bansal, line 17: Nit: can we clean these imports up a bit in a follow-up PR? Let's import by `api/controllers` and `services`.

tests/controllers/index.ts
```diff
@@ -1 +1,148 @@
...
+
+export class ControllerFactory {
```
sumeet-bansal, line 19: This class is getting messy. Let's remove the controller singletons and initialize a new controller each time the factory's called.

sumeet-bansal, line 19: Nit: rename file to the class name.

api/validators/FeedbackControllerRequests.ts
```diff
@@ -24,11 +23,6 @@ export class Feedback implements IFeedback {
   type: FeedbackType;
 }
 
-export class SubmitEventFeedbackRequest implements ISubmitEventFeedbackRequest {
```
sumeet-bansal, line 27: What's this change for?

tests/data/FileFactory.ts
sumeet-bansal, line 6: Do we still need this or can we mock the `File`?

tests/event.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    // TODO: implement once API wrappers exist (since multer validation can't be verified with function calls)
+  });
+});
+
+describe('event feedback', () => {
+  test('can be persisted and rewarded points when submitted for an event already attended', async () => {
+    const conn = await DatabaseConnection.get();
+    const event = EventFactory.fakeOngoingEvent();
+    const user = UserFactory.fake();
+    const feedback = FeedbackFactory.createEventFeedback(3);
```
sumeet-bansal, line 171: I think `createEventFeedback` should go in `EventFactory`.

tests/event.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      .createEvents([event])
+      .attendEvents([user], [event], false)
+      .write();
+
+    await ControllerFactory.event(conn).submitEventFeedback({ uuid: event.uuid }, { feedback }, user);
+
+    const attendanceResponse = await ControllerFactory.attendance(conn).getAttendancesForCurrentUser(user);
+
+    expect(attendanceResponse.attendances[0].feedback).toEqual(feedback);
+    expect(user.points).toEqual(event.pointValue + Config.pointReward.EVENT_FEEDBACK_POINT_REWARD);
```
sumeet-bansal, line 184: How is this passing? Shouldn't we need to refresh the `UserModel` before the points field is updated?

tests/event.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    ).rejects.toThrow('You must submit feedback within 2 days of the event ending');
+  });
+
+  test('is rejected if more than 3 feedback is provided', async () => {
+    const feedback = FeedbackFactory.createEventFeedback(4);
+    const errors = await validate(plainToClass(SubmitEventFeedbackRequest, { feedback }));
+
+    expect(errors).toBeDefined();
+    expect(errors).toHaveLength(1);
+    expect(errors[0].property).toEqual('feedback');
```
sumeet-bansal, line 246: Can we add an `expect` that checks that the reason validation failed is because >3?

tests/feedback.test.ts
```diff
@@ -118,8 +118,8 @@ describe('feedback submission', () => {
     const submittedFeedbackResponse = await feedbackController.submitFeedback({ feedback }, user);
 
     const status = FeedbackStatus.ACKNOWLEDGED;
-    const uuid = submittedFeedbackResponse.feedback;
-    const acknowledgedFeedback = await feedbackController.updateFeedbackStatus(uuid, { status }, admin);
+    const { uuid } = submittedFeedbackResponse.feedback;
+    const acknowledgedFeedback = await feedbackController.updateFeedbackStatus({ uuid }, { status }, admin);
```
sumeet-bansal, line 122: I don't see any changes to the feedback controller so why this change?

tests/mocks/index.ts
```diff
@@ -0,0 +1,10 @@
+import { anything, instance, mock, when } from 'ts-mockito';
+import StorageService from '../../services/StorageService';
+
+export default class Mocks {
+  public static storage(fileLocation: string) {
```
sumeet-bansal, line 5: Inline this and use `verify` to check that we called the mock correctly.

tests/event.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    await new PortalState()
+      .createUsers([admin])
+      .createEvents([event])
+      .write();
+
+    const eventCoverResponse = await ControllerFactory
+      .event(conn, null, Mocks.storage(fileLocation))
+      .updateEventCover(cover, { uuid: event.uuid }, admin);
+
+    expect(eventCoverResponse.event.cover).toEqual(fileLocation);
```
sumeet-bansal, line 158: I think we should get the event and check the cover from that instead of checking the `updateEventCoverResponse`.

sumeet-bansal (changes requested):

tests/controllers/index.ts
shravanhariharan2: Sure, just created an issue for it

tests/controllers/index.ts
shravanhariharan2: True. Is the reason we are willing to trade-off readability for memory because this class is used only in tests and isn't persisting any memory afterwards? (meaning we are okay using a small amount more of memory since it's all present only during the tests running)

api/validators/FeedbackControllerRequests.ts
shravanhariharan2: Moved SubmitEventFeedbackRequest from FeedbackControllerRequests to EventControllerRequests

tests/feedback.test.ts
shravanhariharan2: I believe it was a missing change from [this PR](https://github.com/acmucsd/membership-portal/pull/173) that wasn't causing the test to fail for some reason.

tests/event.test.ts
shravanhariharan2: https://github.com/acmucsd/membership-portal/blob/master/repositories/UserRepository.ts#L52

This line updates the points field directly on the `user` object, so no refresh is necessary.

tests/data/FileFactory.ts
shravanhariharan2: I think we'd still need this for verifying that a file that's too large gets rejected by the controller (the TODO test).

tests/mocks/index.ts
shravanhariharan2: I'm not too sure what you mean by this. What should we be inlining, and where should we be verifying the call?

shravanhariharan2:
Closing this for now since this will not be worked on for quite a bit, as merch store issues are being worked on for the October launch. I'll reopen this once things have settled with the store.

PR #184, opened 2021-05-11
Update and clean up dependencies

sumeet-bansal:
Super nit: I noticed this while reviewing an earlier PR, but we use different versions of Node: `package.json` specifies `^12.0.0` and the Dockerfile specifies `node:14.15.3` so I set the `package.json` to `^14.0.0`.

Node 14's currently LTS but 16 was just released a month ago and becomes active in October. Which do we want to use?

Also includes the Dependabot changes. Would've assumed our `npm ci` check included `npm audit` but I guess not?


package.json
```diff
@@ -33,7 +33,8 @@
     "url": "https://github.com/acmucsd/membership-portal/issues"
   },
   "engines": {
-    "node": "^12.0.0"
+    "node": "^14.0.0",
```
sumeet-bansal, line 36: To match the Dockerfile, but it's nbd.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14.15.3
+FROM node:latest
```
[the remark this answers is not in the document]
shravanhariharan2: I think we should keep it at `v14.15.3`, just so that we're using the LTS across all instances

shravanhariharan2:
> Node 14's currently LTS but 16 was just released a month ago and becomes active in October. Which do we want to use?

I'm thinking we can stick with LTS for a bit until Node 16 becomes LTS / more widely used

shravanhariharan2:
Also, something I noticed when starting the package restructure - if you delete `package-lock.json` and then run `npm install`, then `npm install` fails because of a dependency tree issue with `routing-controllers:0.9.0-alpha4` and `class-transformer:0.2.3`. Changing them both to `v0.9.0` and `v0.3.1` respectfully will succeed, but then the app fails to build because of the `underscore` version. I have no clue why it's not failing now, because if you remove `package-lock.json` and try again it will fail on the same versions as specified here. Not sure if this is something we wanted to look into?

sumeet-bansal:
I deleted the lockfile and `npm install` ran fine for me. Bumped routing-controllers and `npm run build` runs fine too.

sumeet-bansal:
I'll replace the `pick` uses.

Dockerfile
shravanhariharan2, line 1: Like mentioned earlier I think we should pin a version just for consistency

shravanhariharan2 (approved):
Just a follow up on Dockerfile node version consistency, good to merge after that's resolved

Dockerfile
sumeet-bansal: Updated to `node:lts`.

sumeet-bansal:
There's also some deprecated dependencies that we pull in. I'll look into those later today before merging.

sumeet-bansal:
I bumped jest and now the migrations are broken 🙃

PR #187, opened 2021-05-14
Add AdminController Test Suite

### LINKED ISSUE #174: Write admin controller test suite
Assertions we should test:
* /email:
    * Gets all the emails of stored users
* /bonus:
    * Updates points and activity to the users in the bonus request
    * Does not update points and activity to the users who aren't in the bonus request
*  /attendance (already implemented):
    * Logs activity, attendance, and points for users who have not attended
    * Does not log activity, attendance, and points for users who already attended
    * Logs proper activity and point rewards for staff attendance

StormFireFox1:
Closes #174.


sumeet-bansal:
Heads up, you need to rebase (merge conflict).

StormFireFox1:
Just did a squash on top of a rebase, since the merge conflict broke the lockfile pretty bad.

I'll keep this as a draft for another day or two, since I might be able to come up with more tests for the suite.

tests/admin.test.ts
```diff
@@ -117,3 +118,83 @@ describe('retroactive attendance submission', () => {
     expect(staffActivityResponse.activity[1].type).toEqual(ActivityType.ATTEND_EVENT_AS_STAFF);
   });
 });
+
+describe('email retrieval', () => {
+  test('gets all the emails of stored users', async () => {
+    const conn = await DatabaseConnection.get();
+    const users = UserFactory.create(5);
+    const emails = users.map((user) => user.email.toLowerCase());
+    const [proxyUser] = UserFactory.with({ accessType: UserAccessType.ADMIN });
```
sumeet-bansal, line 127: Nit: rename to `admin`.

tests/admin.test.ts
```diff
@@ -117,3 +118,83 @@ describe('retroactive attendance submission', () => {
...
+
+    await new PortalState()
+      .createUsers([...users, proxyUser])
+      .write();
+
+    const adminController = ControllerFactory.admin(conn);
+    const response = await adminController.getAllEmails(proxyUser);
+    expect([...emails, proxyUser.email].sort()).toEqual(response.emails.sort());
```
sumeet-bansal, line 135: Instead of sorting the arrays, there's an array equality check we can use. See the other tests for examples.

tests/admin.test.ts
```diff
@@ -117,3 +118,83 @@ describe('retroactive attendance submission', () => {
...
+  });
+
+  test('no other emails present asides from registered users', async () => {
```
sumeet-bansal, line 138: I don't understand the point of this test. If the email doesn't exist, it wouldn't be returned because it doesn't exist. I'd understand if the endpoint returned some filtered list of addresses but that's not the case.

tests/admin.test.ts
```diff
@@ -117,3 +118,83 @@ describe('retroactive attendance submission', () => {
...
+    expect(response.emails).not.toContain(extraneousUser.email.toLowerCase());
+  });
+});
+
+describe('bonus points submission', () => {
+  test('updates points and activity to the users in the bonus request', async () => {
+    const conn = await DatabaseConnection.get();
+    const users = UserFactory.create(5);
+    const emails = users.map((user) => user.email.toLowerCase());
+    const [proxyUser] = UserFactory.with({ accessType: UserAccessType.ADMIN });
```
sumeet-bansal, line 159: Nit: rename to `admin`.

tests/admin.test.ts
```diff
@@ -117,3 +118,83 @@ describe('retroactive attendance submission', () => {
...
+      .createUsers([...users, proxyUser])
+      .write();
+
+    const adminController = ControllerFactory.admin(conn);
+    const userController = ControllerFactory.user(conn);
+    const request: CreateBonusRequest = { bonus: { description: 'Test addition of bonus points',
+      users: emails,
+      points: 200 } };
+    const bonusResponse = await adminController.addBonus(request, proxyUser);
+    const userResponse = await userController.getUser({ uuid: users[0].uuid }, proxyUser);
```
sumeet-bansal, line 171:
```suggestion
    const bonus = {
      description: 'Test addition of bonus points',
      users: emails,
      points: 200,
    };
    const createBonusResponse = await ControllerFactory.admin(conn).addBonus({ bonus }, admin);
    const getUserResponse = await ControllerFactory.user(conn).getUser({ uuid: users[0].uuid }, admin);
```

tests/admin.test.ts
```diff
@@ -117,3 +118,83 @@ describe('retroactive attendance submission', () => {
...
+
+    expect(userResponse.user.points).toEqual(200);
```
sumeet-bansal, line 173: If we're granting bonuses to multiple accounts, let's verify them all.

tests/admin.test.ts
```diff
@@ -117,3 +118,83 @@ describe('retroactive attendance submission', () => {
...
+    expect(bonusResponse.emails.sort()).toEqual(emails.sort());
+  });
+
+  test("Does not update points and activity to the users who aren't in the bonus request", async () => {
```
sumeet-bansal, line 177: This can be combined with the above test: a couple users get the bonus, a couple don't.

tests/admin.test.ts
shravanhariharan2: Agreed, no filtering is happening or is planned, so I don't think this test is necessary

tests/admin.test.ts
shravanhariharan2, line 135: nit: you can inline `adminController` since it's only used once:
```ts
const getAllEmailsResponse = await ControllerFactory.admin(conn).getAllEmails(admin);
```

tests/admin.test.ts
shravanhariharan2, line 171: nit: with multiple assertions, we should assert right after getting a response (see Sumeet's explanation [here](https://github.com/acmucsd/membership-portal/pull/181#discussion_r626136760)), i.e.
```ts
const createBonusResponse = ...
expect(createBonusResponse.emails)...

const getUserResponse = ...
expect (getUserResponse.user)...
```

tests/admin.test.ts
StormFireFox1: Resolved in 11c362c.

tests/admin.test.ts
StormFireFox1: Resolved in fdbbd08.

tests/admin.test.ts
StormFireFox1: Removed in 3a27e7f.

tests/admin.test.ts
StormFireFox1: Resolved in 11c362c.

tests/admin.test.ts
StormFireFox1: Resolved in 5fc6a9a.

tests/admin.test.ts
StormFireFox1: Resolved in 5fe3084.

tests/admin.test.ts
StormFireFox1: Resolved in 1e37dd2.

tests/admin.test.ts
StormFireFox1: Resolved in 5fc6a9a. There are a few instances I inlined as well.

tests/admin.test.ts
StormFireFox1: Resolved in 5ad4381.

Some other tests are required rearragements in the same file; I arranged those as well.

tests/admin.test.ts
```diff
@@ -20,31 +20,31 @@ describe('retroactive attendance submission', () => {
     const conn = await DatabaseConnection.get();
     const users = UserFactory.create(3);
     const emails = users.map((user) => user.email);
-    const [proxyUser] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+    const [adminUser] = UserFactory.with({ accessType: UserAccessType.ADMIN });
     const [event] = EventFactory.create(1);
```
shravanhariharan2, line 24: nit: this should remain as `proxyUser`. The reason we have it named `proxyUser` here, but not elsewhere, is because this user is submitting attendance on behalf of the other users, and is hence a "proxy" (see definition 3 of [proxy definition](https://www.merriam-webster.com/dictionary/proxy)). We originally had it named as `onBehalfOfUser` but it didn't flow off the tongue as easily

tests/admin.test.ts
```diff
@@ -104,16 +105,67 @@ describe('retroactive attendance submission', () => {
...
   });
 });
+
+describe('email retrieval', () => {
+  test('gets all the emails of stored users', async () => {
+    const conn = await DatabaseConnection.get();
+    const users = UserFactory.create(5);
+    const emails = users.map((user) => user.email.toLowerCase());
+    const [adminUser] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+
```
shravanhariharan2, line 130: nit: rename `adminUser` to `admin` (for consistency with other tests + "user" is redundant since `admin` implies that the object is a user)

tests/admin.test.ts
```diff
@@ -104,16 +105,67 @@ describe('retroactive attendance submission', () => {
...
+    await new PortalState()
+      .createUsers([...users, adminUser])
+      .write();
+
+    const adminController = ControllerFactory.admin(conn);
+    const response = await adminController.getAllEmails(adminUser);
+    expect([...emails, adminUser.email]).toEqual(expect.arrayContaining(response.emails));
+  });
```
shravanhariharan2, line 138: nit: the assertions are flipped here between expected and actual. Every other test has:
```ts
expect(actual).toEqual(expected)
```  
but this one's:
```ts
expect(expected).toEqual(actual)
```

tests/admin.test.ts
```diff
@@ -104,16 +105,67 @@ describe('retroactive attendance submission', () => {
...
+});
+
+describe('bonus points submission', () => {
+  test('updates points and activity to the users in the bonus request', async () => {
+    const conn = await DatabaseConnection.get();
+    const users = UserFactory.create(5);
+    const emails = users.map((user) => user.email.toLowerCase());
+    const [adminUser] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+    const [extraneousUser] = UserFactory.create(1);
```
shravanhariharan2, line 147: nit: I believe you can do:
```ts
const [...users, extraneousUser] = UserFactory.create(5);
```
to declare both variables at once. I think this way of declaration makes the relationship between users and the extraneous user clearer than having them be declared separately

shravanhariharan2: also `adminUser` -> `admin`

tests/admin.test.ts
shravanhariharan2: also nit: inline `adminController`

shravanhariharan2 (commented):
Small nitpicks on nomenclature/style for consistency, otherwise tests themselves LGTM

tests/admin.test.ts
sumeet-bansal: I think `admin`'s fine since it's an admin account.

tests/admin.test.ts
sumeet-bansal:
```typescript
const [...usersGettingBonus, userNotGettingBonus] = UserFactory.create(5);
```
or something similar.

tests/admin.test.ts
StormFireFox1: I think `admin` would make the most sense, since the bonus route can only be used by an `admin` user anyway, so it also suggests proper flow for the endpoint as well.

Resolved in `df1a901`.

tests/admin.test.ts
StormFireFox1: Resolved in `df1a901`.

tests/admin.test.ts
StormFireFox1: Resolved in `aca6574`.

tests/admin.test.ts
StormFireFox1: Resolved in `2afef57`.

shravanhariharan2 (approved):

tests/admin.test.ts
```diff
@@ -71,15 +71,16 @@ describe('retroactive attendance submission', () => {
...
     );
 
-    const userResponse = await userController.getUser({ uuid: user.uuid }, proxyUser);
-    const attendanceResponse = await attendanceController.getAttendancesForCurrentUser(user);
-    const activityResponse = await userController.getCurrentUserActivityStream(user);
-
+    const userResponse = await userController.getUser({ uuid: user.uuid }, admin);
     expect(userResponse.user.points).toEqual(user.points);
+
+    const attendanceResponse = await attendanceController.getAttendancesForCurrentUser(user);
```
sumeet-bansal, line 80: Can we verify this attendance is what we expect?

tests/admin.test.ts
```diff
@@ -71,15 +71,16 @@ describe('retroactive attendance submission', () => {
...
     expect(attendanceResponse.attendances).toHaveLength(1);
+
+    const activityResponse = await userController.getCurrentUserActivityStream(user);
```
sumeet-bansal, line 83: Same here, let's verify the activity objects.

tests/admin.test.ts
```diff
@@ -104,16 +105,67 @@ describe('retroactive attendance submission', () => {
...
+      .createUsers([...users, userNotGettingBonus, admin])
+      .write();
+
+    const bonus = {
+      description: 'Test addition of bonus points',
+      users: emails,
+      points: 200,
+    };
+
+    const userController = ControllerFactory.user(conn);
```
sumeet-bansal, line 157: Nit: let's declare this as late as possible, i.e. right before the loop.

tests/admin.test.ts
```diff
@@ -104,16 +105,67 @@ describe('retroactive attendance submission', () => {
...
+
+    const createBonusResponse = await ControllerFactory.admin(conn).addBonus({ bonus }, admin);
+    expect(createBonusResponse.emails).toEqual(expect.arrayContaining(emails));
+
+    for (let u = 0; u < users.length; u += 1) {
+      const user = users[u];
+      const getUserResponse = await userController.getUser({ uuid: user.uuid }, admin);
+      expect(getUserResponse.user.points).toEqual(200);
```
sumeet-bansal, line 165: Let's also check that the users' activities are correct.

sumeet-bansal (commented):
LGTM, just a couple assertions to add.

tests/admin.test.ts
StormFireFox1: Resolved in 2c591a5.

tests/admin.test.ts
StormFireFox1: Resolved in 2c591a5.

tests/admin.test.ts
StormFireFox1: Resolved in 6688e90.

tests/admin.test.ts
StormFireFox1: Resolved in 6688e90.

tests/admin.test.ts
```diff
@@ -71,28 +71,33 @@ describe('retroactive attendance submission', () => {
...
-    const attendanceResponse = await attendanceController.getAttendancesForCurrentUser(user);
-    const activityResponse = await userController.getCurrentUserActivityStream(user);
-
+    const userResponse = await userController.getUser({ uuid: user.uuid }, admin);
     expect(userResponse.user.points).toEqual(user.points);
+
+    const attendanceResponse = await attendanceController.getAttendancesForCurrentUser(user);
     expect(attendanceResponse.attendances).toHaveLength(1);
+    expect(attendanceResponse.attendances[0].event.uuid).toEqual(event.uuid);
+    expect(attendanceResponse.attendances[0].asStaff).toEqual(false);
```
sumeet-bansal, line 83: What's this assertion for? This doesn't seem useful and can be removed.

sumeet-bansal (approved):

tests/admin.test.ts
StormFireFox1: As per the previous review, this verifies the attendance response matches what we'd expect to receive when attending a specific event.

sumeet-bansal: Ah right. LGTM but one of the tests is failing.

StormFireFox1:
Due to a bug in the factory code causing events to randomly have start times before account creation times, it's possible for non-deterministic test suite failures because of account creation occurring after event attendance.

This bug is (supposedly) solved by PR #181 including the `fakeOngoingEvent` method. This PR is blocked until that one is merged.

shravanhariharan2:
image

Here's an example - the event attendance timestamp occurs before the user creation since presumably, the event is randomly created in the past and has its timestamp set to a random time between start and end date. See https://github.com/acmucsd/membership-portal/blob/master/tests/data/PortalState.ts#L88

sumeet-bansal:
@StormFireFox1 That PR hasn't been updated in a month and has changes requested, so let's just copy those factory changes over to this PR and merge this. I don't think there shouldn't be any merge conflicts, but even if there are, #181 can handle them.

sumeet-bansal:
Actually @StormFireFox1 this probably affects a bunch of other tests too. We could just set the timestamp of the account creation activity to like a month ago to ensure that activity always happens first.

StormFireFox1:
I figured yesterday explicitly setting account creation earlier is safer. I'll readjust all this tonight.

PR #188, opened 2021-05-19
Refactor merch store options for single option type and allow hidden items to have 0 options

### LINKED ISSUE #185: Update merch store schema to optimize single "option group" items
The item option requirement has been eased to support only 1 "option group" (e.g. size, color, shape, etc) per item. A design shown below (courtesy of @steets250) reflects this change:

image

Additionally, merch collection titles support text colors, which can be customized and would need to be stored in the backend:

image

The following schema changes have been planned:
```ts
interface MerchandiseCollectionModel {
  uuid: Uuid;
  title: string;
  titleTextColor: string; // hex color used for the title text (each collection title has its own text color)
  description: string;
  archived: boolean;
  items: MerchandiseItemModel[];
}

interface MerchandiseItemModel {
  uuid: Uuid;
  itemName: string;
  collection: MerchandiseCollectionModel;
  picture: string;
  description: string;
  monthlyLimit: number;
  lifetimeLimit: number;
  hidden: boolean;
  option/optionGroup/optionCategory: string | null; // exact name TBD, but is title of options, (e.g. the string "size", the string "color"; null if no options (since we only have 1 option per item)
  options: MerchandiseItemOptionModel[];
}

interface MerchandiseItemOptionModel {
  uuid: Uuid;
  item: MerchandiseItemModel;
  quantity: number;
  price: number;
  discountPercentage: number;
  value: string; // the option value, e.g. 'S', 'M', 'L' if this.item.optionGroup === 'size'
  position: number; // 0, 1, 2, etc. (for determining sort order of displaying options on frontend, i.e. XS (0) < S (1) < M (2) < L (3)< XL (4)etc)
  orders: OrderItemModel[];
}
```

shravanhariharan2:
Closes #185 

Functionality added:
* Merch collections now have a `themeColorHex: string` field signifying a UI theme for a particular collection
* Merch items can now have 0 options associated with it only if the item is hidden
* Merch items now have a `hasVariantsEnabled: boolean` field signifying whether or not the item is capable of holding multiple options. I stayed away from naming it `hasOptions` since the item would still have an option, but it would be a singleton option
* Merch items with `hasVariantsEnabled: false` cannot have more than 1 option
* Merch item options' `metadata` field has been updated to have a constraint of having type `MerchItemOptionMetadata`, which consists of three fields:
```ts
{
  type: string;
  value: string;
  position: number;
}
``` 
I've made the decision to define these fields in a type because it's been established that that's how our store will operate on the `metadata` field, and so that there's explicit types for them instead of leaving them as type `object`. There isn't a separate table for that field because we may find out that the metadata field will need to be adapted differently in the short term (since we'd only really know once the frontend starts coding the UI components/functionality), so I wanted to keep it not _too_ rigid but also not super flexible for clarity sake.
* Merch items cannot have options where `MerchItemOption.metadata.type` is different among options. This enforces the "single option type" constraint mentioned in the PR/issue title


sumeet-bansal:
Since this PR's editing the seeds as well I'll merge #163 this weekend.

migrations/0027-specify-merch-item-variants.ts
```diff
@@ -0,0 +1,14 @@
...
+export class SpecifyMerchItemVariants1621463589936 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn('MerchandiseItems', new TableColumn({
+      name: 'hasVariants',
+      type: 'boolean',
+    }));
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.dropColumn('MerchandiseItems', 'hasVariants');
```
sumeet-bansal, line 12: I think `variantsEnabled` or something along those lines would be clearer.

models/MerchandiseItemOptionModel.ts
```diff
@@ -24,19 +24,20 @@ export class MerchandiseItemOptionModel extends BaseEntity {
   @Column('integer', { default: 0 })
   discountPercentage: number;
 
+  // Potential fields: type: string, value: string, position: number
```
sumeet-bansal, line 27: Potential fields? Isn't this either a NULL value or an object with all three fields?

services/MerchStoreService.ts
```diff
@@ -44,7 +46,7 @@ export default class MerchStoreService {
     const collection = await this.transactions.readOnly(async (txn) => Repositories
       .merchStoreCollection(txn)
       .findByUuid(uuid));
-    if (!collection) throw new NotFoundError('Collection not found');
+    if (!collection) throw new NotFoundError(NotFoundErrors.MERCH_COLLECTION);
```
sumeet-bansal, line 49:
```suggestion
    if (!collection) throw NotFoundErrors.MERCH_COLLECTION;
```

services/MerchStoreService.ts
```diff
@@ -100,32 +102,49 @@ export default class MerchStoreService {
...
       .findByUuid(uuid));
-    if (!item) throw new NotFoundError('Item not found');
+    if (!item) throw new NotFoundError(NotFoundErrors.MERCH_ITEM);
     return item.getPublicMerchItem();
   }
 
   public async createItem(item: MerchItem): Promise<MerchandiseItemModel> {
     return this.transactions.readWrite(async (txn) => {
+      if (this.hasMultipleOptionTypes(item.options)) throw new UserError(UserErrors.MULTIPLE_MERCH_OPTION_TYPES);
+      if (!item.hasVariants && item.options.length > 1) {
```
sumeet-bansal, line 112: This error should go first I think.

services/MerchStoreService.ts
```diff
@@ -100,32 +102,49 @@ export default class MerchStoreService {
...
+
       const merchItemRepository = Repositories.merchStoreItem(txn);
       const merchItem = MerchandiseItemModel.create({ ...item, collection });
       await merchItemRepository.upsertMerchItem(merchItem);
       return merchItemRepository.findByUuid(merchItem.uuid);
     });
   }
 
+  private hasMultipleOptionTypes(options: MerchItemOption[] | MerchandiseItemOptionModel[] | MerchItemOptionEdit[]) {
+    for (let i = 1; i < options.length; i += 1) {
```
sumeet-bansal, line 127: This errors if an item only has a single variant. An alternative could be taking the set of all the metadata types and returning based on set size.

services/MerchStoreService.ts
```diff
@@ -100,32 +102,49 @@ export default class MerchStoreService {
...
+  }
+
   public async editItem(uuid: Uuid, itemEdit: MerchItemEdit): Promise<MerchandiseItemModel> {
     return this.transactions.readWrite(async (txn) => {
       const merchItemRepository = Repositories.merchStoreItem(txn);
       const item = await merchItemRepository.findByUuid(uuid);
       if (!item) throw new NotFoundError();
-      const { options, collection: updatedCollection, ...changes } = itemEdit;
+      const { options, collection: updatedCollection, hasVariants, ...changes } = itemEdit;
+      if (this.hasMultipleOptionTypes(options)) throw new UserError(UserErrors.MULTIPLE_MERCH_OPTION_TYPES);
```
sumeet-bansal, line 141: What happens if the options in the edit all have some type and all the options in the model have some other type?

services/MerchStoreService.ts
```diff
@@ -100,32 +102,49 @@ export default class MerchStoreService {
...
+      if (!hasVariants && options.length > 1) throw new UserError(UserErrors.NO_ITEM_VARIANTS_BUT_MULTIPLE_OPTIONS);
```
sumeet-bansal, line 142: Similar thing here where we're only checking the options in the request not the final state. What if we have multiple variants but then get a request to disable variants?

services/MerchStoreService.ts
```diff
@@ -154,19 +173,24 @@ export default class MerchStoreService {
...
     });
   }
 
   public async createItemOption(item: Uuid, option: MerchItemOption): Promise<PublicMerchItemOption> {
     return this.transactions.readWrite(async (txn) => {
       const merchItem = await Repositories.merchStoreItem(txn).findByUuid(item);
-      if (!merchItem) throw new NotFoundError('Item not found');
+      if (!merchItem) throw new NotFoundError(NotFoundErrors.MERCH_ITEM);
+      if (!merchItem.hasVariants) throw new UserError(UserErrors.NO_ITEM_VARIANTS_ADD_OPTION);
+      const hasDifferentOptionType = merchItem.options && merchItem.options[0].metadata?.type !== option.metadata?.type;
```
sumeet-bansal, line 186: The portal should guarantee that all items have at least one option, so no need to check that `merchItem.options` exists and is not empty.

services/MerchStoreService.ts
```diff
@@ -185,7 +209,7 @@ export default class MerchStoreService {
     const order = await this.transactions.readOnly(async (txn) => Repositories
       .merchOrder(txn)
       .findByUuid(uuid));
-    if (!order) throw new NotFoundError();
+    if (!order) throw new NotFoundError(NotFoundErrors.MERCH_ORDER);
```
sumeet-bansal, line 212: Can we make these changes in another PR?

tests/data/MerchFactory.ts
```diff
@@ -18,6 +19,13 @@ export class MerchFactory {
     return FactoryUtils.create(n, MerchFactory.fakeOption);
   }
 
+  public static createOptionMetadataOfSameType(n: number): MerchItemOptionMetadata[] {
```
sumeet-bansal, line 22: Nit: `createOptionMetadata`, since this method would only create valid metadata.

tests/data/MerchFactory.ts
```diff
@@ -38,21 +46,35 @@ export class MerchFactory {
...
+    const hasVariants = FactoryUtils.getRandomBoolean();
+    const maxNumOptions = hasVariants ? 5 : 1;
     return MerchandiseItemModel.create({
       uuid: uuid(),
-      itemName: faker.random.hexaDecimal(10),
+      itemName: faker.datatype.hexaDecimal(10),
       description: faker.lorem.sentences(2),
-      options: MerchFactory.createOptions(FactoryUtils.getRandomNumber(1, 5)),
+      hasVariants,
+      options: MerchFactory.createOptions(FactoryUtils.getRandomNumber(1, maxNumOptions)),
```
sumeet-bansal, line 77: Shouldn't this also depend on `hasVariants`? Maybe instead of `maxNumOptions`,
```typescript
const numOptions = hasVariants ? FactoryUtils.getRandomNumber(1, 5) : 1;
```

tests/data/MerchFactory.ts
```diff
@@ -65,6 +87,14 @@ export class MerchFactory {
     });
   }
 
+  public static fakeOptionMetadata(): MerchItemOptionMetadata {
+    return {
+      type: faker.datatype.hexaDecimal(10),
+      value: faker.datatype.hexaDecimal(10),
+      position: FactoryUtils.getRandomNumber(1, 10),
```
sumeet-bansal, line 94: If we always choose a random number for `position`, wouldn't we end up creating options with the same positions?

tests/merch.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      .createMerch([merchCollection])
+      .write();
+
+    const response = await ControllerFactory.merchStore(conn).createMerchItemOption(
+      { uuid: merchItem.uuid },
+      { option: newOption },
+      admin,
+    );
+
+    expect(response.error).toBeNull();
```
sumeet-bansal, line 44: Let's call `getMerchItem` and verify both options are there.

tests/merch.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    await expect(
+      ControllerFactory.merchStore(conn).createMerchItemOption(
+        { uuid: merchItem.uuid },
+        { option: newOption },
+        admin,
+      ),
+    ).rejects.toThrow(UserErrors.NO_ITEM_VARIANTS_ADD_OPTION);
+  });
+
+  test('cannot be added to an item with different metadata types', async () => {
```
sumeet-bansal, line 75:
```suggestion
  test('cannot add different metadata types to an item', async () => {
```

types/ApiResponses.ts
```diff
@@ -152,14 +153,21 @@ export interface PublicMerchItem {
   description: string;
   monthlyLimit: number;
   lifetimeLimit: number;
+  hasVariants: boolean;
   options: PublicMerchItemOption[];
 }
 
+export interface PublicMerchItemOptionMetadata {
```
sumeet-bansal, line 160: This is the same as `MerchItemOptionMetadata`.

sumeet-bansal (changes requested):

models/MerchandiseItemOptionModel.ts
shravanhariharan2: Correct, I added that comment when `metadata` was still of type `object`, but now that it's strongly-ish typed (i.e. has a type but no schema associated with it), the comment is redundant

migrations/0026-add-merch-collection-theme.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+export class AddMerchCollectionTheme1621463276136 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn('MerchandiseCollections', new TableColumn({
+      name: 'themeColor',
+      type: 'varchar(255)',
+      isNullable: true,
```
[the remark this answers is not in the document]
shravanhariharan2: Looks like theme color is planned to be required on frontend, so whatever is sent to the backend should be whatever is stored

services/MerchStoreService.ts
shravanhariharan2: Doesn't this not throw a 404 error then and instead throw a 500 (the default error code defined in `errors/handler.ts`)?

tests/data/MerchFactory.ts
shravanhariharan2: I was thinking `position` could be a field that's solely managed by the frontend but stored in the backend, so that the frontend can order/display items however they like based on the position. Do you think we should be enforcing position order (i.e. it's always 0, 1, 2, 3, etc) and uniqueness (i.e. only one 1, only one 2, etc)?

migrations/0026-add-merch-collection-theme.ts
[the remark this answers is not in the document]
sumeet-bansal: So we can set it to required in the database then?

services/MerchStoreService.ts
shravanhariharan2: Sure we could - I initially refactored them because a lot of the error messages were duplicated, but it just seems weird having one service with canned error messages and others without. Plus this PR is pretty sizeable on its own.

services/MerchStoreService.ts
sumeet-bansal: The idea's that `NotFoundErrors.MERCH_COLLECTION` would be `new NotFoundException(message)`. Nicer caller syntax.

services/MerchStoreService.ts
shravanhariharan2: Doesn't the loop not get executed if `options.length === 1`? Which means the function returns `false` and the errors wouldn't be thrown?

tests/data/MerchFactory.ts
sumeet-bansal: I initially thought so since the database should always have "clean"/valid data, but now I don't think so since FEs can order by `position` even if there're gaps and duplicate numbers.

services/MerchStoreService.ts
sumeet-bansal: Ah yeah, you're right.

sumeet-bansal: I personally prefer using sets for things like this since it's simpler but the loop works too.

migrations/0026-add-merch-collection-theme.ts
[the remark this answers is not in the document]
shravanhariharan2: yup

services/MerchStoreService.ts
shravanhariharan2: Great point - there's 2 approaches I'm thinking for that:
1. Keep the check on line 141. After that, check if `itemEdit.options.length` is equal to `item.options.length` (i.e. every option is being edited) and throw if the `type` field is changed in the edit. If not every option is being edited and instead the `itemEdit.options.length` array is filled with bogus edits that make the above length check pass, then the `Option Not Found` error will be thrown instead in the code below
2. Remove the check on line 141, and after [this](https://github.com/acmucsd/membership-portal/blob/4a9d27d7f90f90d953d8d0111c452f27d1c03dff/services/MerchStoreService.ts#L137) line, we do:
```ts
item.options = [...updatedOptions, ...item.options.filter((option) => !updatedOptionsIds.has(option.uuid))];
if (this.hasMultipleOptionTypes(item.options)) throw new UserError(UserErrors.MULTIPLE_MERCH_OPTION_TYPES);
```

Approach 2 is a lot simpler but the only downside is reads/writes are potentially wasted in the `Promise.all` call, but since the service is transacted, we can just throw and everything will get reverted. But again it doesn't really make sense to write data and then validate after it's been written. Meanwhile, approach 1 doesn't waste reads/writes but is a bit more complex of a check. Thoughts / is there any other approach you're thinking could work?

services/MerchStoreService.ts
shravanhariharan2: I think sets make more sense in this case, but it also might get weird since `metadata` could be null/undefined, so we shouldn't add anything to the set if that's the case, but then we'd need to check if the set is of size 0 or 1 (i.e. either single variant or multi variant with 1 option type), which makes it pretty much as complicated as the loop but without constant space.

shravanhariharan2: So it might just be better to stick with the loop even though it's a little more archaic

sumeet-bansal: This function is only called for items with variants enabled, so we should be catching undefined `metadata` before calling this.

services/MerchStoreService.ts
sumeet-bansal: Why would we need to write before validating? We already have the current item and its options, we merge the edits, we validate the updated item, and then we write the updated item.

services/MerchStoreService.ts
shravanhariharan2: I don't think that's always the case - the function is ran if the check above it passes:
```ts
if (!item.hasVariantsEnabled && item.options.length > 1) {
  throw new UserError(UserErrors.VARIANTS_DISABLED_MULTIPLE_OPTIONS);
}
if (this.hasMultipleOptionTypes(item.options)) throw new UserError(UserErrors.MULTIPLE_MERCH_OPTION_TYPES);
```
so `item` could have `hasVariantsEnabled = false` and only 1 option (thus having potentially empty metadata), or it could have `hasVariantsEnabled = true` and any number of options.

sumeet-bansal: We should only be calling `hasMultipleOptionTypes` for items with variants enabled, so
```typescript
if (item.hasVariantsEnabled && this.hasMultipleOptionTypes(item.options)) { ... }
```
should work, since if variants is enabled for an item its options should always have metadata.

services/MerchStoreService.ts
shravanhariharan2: I'll keep this in mind but will hold off till a separate, error refactor PR.

shravanhariharan2:
@sumeet-bansal does this PR introduce "incompatible API changes" (the addition of `hasVariantsEnabled` being required)? If so, does it warrant the bump to `/api/v3`? I feel like it isn't that big a change but it still might be incompatible since all clients would need to add that `hasVariantsEnabled` boolean to their merch item requests.

shravanhariharan2:
@sumeet-bansal shouldn't be any breaking changes now. Let me know if you see anything else

api/validators/MerchStoreRequests.ts
```diff
@@ -109,6 +130,9 @@ export class MerchItem implements IMerchItem {
   @Allow()
   hidden?: boolean;
 
+  @Allow()
+  hasVariantsEnabled?: boolean;
```
sumeet-bansal, line 134: Instead of setting the field as optional, can we specify a default through class-validator? If not, let's set a default in the controller.

api/validators/MerchStoreRequests.ts
```diff
@@ -16,13 +16,17 @@ import {
...
+  MerchItemOptionMetadata as IMerchItemOptionMetadata,
 } from '../../types';
 
 export class MerchCollection implements IMerchCollection {
   @IsDefined()
   @IsNotEmpty()
   title: string;
 
+  @Allow()
+  themeColorHex?: string;
```
sumeet-bansal, line 28: Should we validate that this is actually a hexadecimal value with expected length? Or just leave that to the FE like `position`. I think we should.

services/MerchStoreService.ts
```diff
@@ -100,47 +100,70 @@ export default class MerchStoreService {
...
       .merchStoreItem(txn)
       .findByUuid(uuid));
-    if (!item) throw new NotFoundError('Item not found');
+    if (!item) throw new NotFoundError('Merch item not found');
     return item.getPublicMerchItem();
   }
 
   public async createItem(item: MerchItem): Promise<MerchandiseItemModel> {
     return this.transactions.readWrite(async (txn) => {
+      item.hasVariantsEnabled ??= false;
```
sumeet-bansal, line 109: An assignment like this belongs in the controller I think, since it's request-related ("did the request send in a value for `hasVariantsEnabled`").

services/MerchStoreService.ts
```diff
@@ -100,47 +100,70 @@ export default class MerchStoreService {
...
       const merchItemRepository = Repositories.merchStoreItem(txn);
       const merchItem = MerchandiseItemModel.create({ ...item, collection });
       await merchItemRepository.upsertMerchItem(merchItem);
       return merchItemRepository.findByUuid(merchItem.uuid);
     });
   }
 
+  // verify that item.hasVariantsEnabled is consistent with the item options length and option type consistency
+  // as defined by the merch store functionality
+  private verifyVariantAndOptionConsistency(item: MerchItem | MerchandiseItemModel) {
```
sumeet-bansal, line 124: Where is this defined besides right here in the code? Also this can be static, and can the parameter be only type?

```suggestion
  /**
   * Verify that items have valid options. An item with variants disabled cannot have multiple 
   * options, and an item with variants enabled cannot have multiple option types.
   */
   private static verifyItemHasValidOptions(item: MerchandiseItemModel) {
```

services/MerchStoreService.ts
```diff
@@ -100,47 +100,70 @@ export default class MerchStoreService {
...
+    if (!item.hasVariantsEnabled && item.options.length > 1) {
+      throw new UserError('Items with variants disabled cannot have multiple options');
+    }
+    if (item.hasVariantsEnabled && this.hasMultipleOptionTypes(item.options)) {
+      throw new UserError('Item cannot have multiple option types');
```
sumeet-bansal, line 129: To match the other error messages.
```suggestion
      throw new UserError('Merch items with variants disabled cannot have multiple options');
    }
    if (item.hasVariantsEnabled && this.hasMultipleOptionTypes(item.options)) {
      throw new UserError('Merch items cannot have multiple option types');
```

services/MerchStoreService.ts
```diff
@@ -100,47 +100,70 @@ export default class MerchStoreService {
...
+    }
+  }
+
+  private hasMultipleOptionTypes(options: MerchItemOption[]) {
```
sumeet-bansal, line 133: Nit: this can be static.

services/MerchStoreService.ts
```diff
@@ -100,47 +100,70 @@ export default class MerchStoreService {
...
-        const option = await merchItemOptionRepository.findByUuid(optionUpdate.uuid);
-        if (!option) throw new NotFoundError('Item option not found');
-        // 'quantity' is incremented instead of directly set to avoid concurrency issues with orders
-        // e.g. there's 10 of an item and someone adds 5 to stock while someone else orders 1
-        // so the merch store admin sets quantity to 15 but the true quantity is 14
-        if (optionUpdate.quantity) optionUpdate.quantity += option.quantity;
-        return merchItemOptionRepository.upsertMerchItemOption(option, optionUpdate);
-      }));
+      if (options) {
+        const updatedOptionMap = new Map(options.map((option) => [option.uuid, option]));
```
sumeet-bansal, line 146: Nit: for maps, variable names should include what the keys are, e.g. `updatedOptionsByUuid`.

services/MerchStoreService.ts
```diff
@@ -100,47 +100,70 @@ export default class MerchStoreService {
...
+        item.options.map((option) => {
+          if (!updatedOptionMap.has(option.uuid)) return;
+          const updatedOption = updatedOptionMap.get(option.uuid);
+          // 'quantity' is incremented instead of directly set to avoid concurrency issues with orders
+          // e.g. there's 10 of an item and someone adds 5 to stock while someone else orders 1
+          // so the merch store admin sets quantity to 15 but the true quantity is 14
+          if (updatedOption.quantity) updatedOption.quantity += option.quantity;
```
sumeet-bansal, line 153: Can we rename this property to `addQuantity` in the updated option type? I think it's clearer that way.

services/MerchStoreService.ts
```diff
@@ -100,47 +100,70 @@ export default class MerchStoreService {
...
+          return MerchandiseItemOptionModel.merge(option, updatedOption);
+        });
+      }
 
-      const updatedOptionsIds = new Set(updatedOptions.map((option) => option.uuid));
-      item.options = [...updatedOptions, ...item.options.filter((option) => !updatedOptionsIds.has(option.uuid))];
+      // ensure the hasVariantsEnabled <--> options.length invariant is still consistent
+      if (changes.hasVariantsEnabled !== undefined) item.hasVariantsEnabled = changes.hasVariantsEnabled;
+      this.verifyVariantAndOptionConsistency(item);
```
sumeet-bansal, line 160: Let's just merge the item here and validate the new version.

```suggestion
      this.verifyItemHasValidOptions(MerchandiseItemModel.merge(item, changes));
```

services/MerchStoreService.ts
```diff
@@ -154,19 +177,24 @@ export default class MerchStoreService {
...
     });
   }
 
   public async createItemOption(item: Uuid, option: MerchItemOption): Promise<PublicMerchItemOption> {
     return this.transactions.readWrite(async (txn) => {
       const merchItem = await Repositories.merchStoreItem(txn).findByUuid(item);
-      if (!merchItem) throw new NotFoundError('Item not found');
+      if (!merchItem) throw new NotFoundError('Merch item not found');
+      if (!merchItem.hasVariantsEnabled) throw new UserError('Cannot add option to items with variants disabled');
+      const hasDifferentOptionType = merchItem.options[0].metadata?.type !== option.metadata?.type;
```
sumeet-bansal, line 190: `metadata` should always be present right? So do we need the question marks?

services/MerchStoreService.ts
```diff
@@ -154,19 +177,24 @@ export default class MerchStoreService {
...
+      if (hasDifferentOptionType) throw new UserError('Item cannot have multiple option types');
```
sumeet-bansal, line 191:
```suggestion
      if (hasDifferentOptionType) throw new UserError('Merch item cannot have multiple option types');
```

services/MerchStoreService.ts
```diff
@@ -154,19 +177,24 @@ export default class MerchStoreService {
...
+
       const merchItemOptionRepository = Repositories.merchStoreItemOption(txn);
       const createdOption = MerchandiseItemOptionModel.create({ ...option, item: merchItem });
       await merchItemOptionRepository.upsertMerchItemOption(createdOption);
-      return merchItemOptionRepository.findByUuid(createdOption.uuid);
+      const upsertedOption = await merchItemOptionRepository.findByUuid(createdOption.uuid);
```
sumeet-bansal, line 196: Doesn't `upsertMerchItemOption` return `upsertedOption`?

tests/data/MerchFactory.ts
```diff
@@ -15,7 +16,21 @@ export class MerchFactory {
...
   public static createOptions(n: number): MerchandiseItemOptionModel[] {
-    return FactoryUtils.create(n, MerchFactory.fakeOption);
+    const options = FactoryUtils.create(n, MerchFactory.fakeOption);
+    if (n === 1) return options;
+
+    const optionMetadata = this.createOptionMetadata(n);
+    return options.map((option, i) => {
+      option.metadata = optionMetadata[i];
+      return option;
+    });
```
sumeet-bansal, line 26: Instead of creating a matching array, we can add metadata to each option.
```suggestion
    const type = faker.datatype.hexaDecimal(10);
    return options.map((o) => {
      o.metadata = MerchFactory.optionMetadataWith({ type });
      return o;
    });
```

tests/data/MerchFactory.ts
```diff
@@ -38,22 +53,37 @@ export class MerchFactory {
...
+      };
+    });
+  }
+
   public static fakeCollection(substitute?: Partial<MerchandiseCollectionModel>): MerchandiseCollectionModel {
     const fake = MerchandiseCollectionModel.create({
       uuid: uuid(),
       title: faker.datatype.hexaDecimal(10),
       description: faker.lorem.sentences(2),
+      themeColorHex: faker.internet.color(),
```
sumeet-bansal, line 72: We should specify somewhere that it's plain hex so people don't try sending it with the pound sign or `0x` or something.

tests/data/MerchFactory.ts
```diff
@@ -68,6 +98,14 @@ export class MerchFactory {
     return MerchandiseItemOptionModel.merge(fake, substitute);
   }
 
+  public static fakeOptionMetadata(): MerchItemOptionMetadata {
```
sumeet-bansal, line 101: Should we just always pass in `type` here? Seems much simpler that way.

tests/merch.test.ts
```diff
@@ -0,0 +1,303 @@
...
+
+    const merchItemResponse = await ControllerFactory.merchStore(conn).getOneMerchItem({ uuid: item.uuid }, admin);
+
+    expect(merchItemResponse.item).toStrictEqual({
+      collection: merchItemResponse.item.collection,
+      ...item.getPublicMerchItem(),
+      ...merchItemEdits,
+    });
+  });
+  test('succeeds when basic option fields are updated', async () => {
```
sumeet-bansal, line 57: Newline nit

```suggestion
  });

  test('succeeds when basic option fields are updated', async () => {
```

tests/merch.test.ts
```diff
@@ -0,0 +1,303 @@
...
+
+    expect(merchItemResponse.item.options).toEqual(
+      expect.arrayContaining([option1.getPublicMerchItemOption(), option2.getPublicMerchItemOption()]),
+    );
+  });
+  test('succeeds when updated item option metadata types are changed but are still consistent', async () => {
+    const conn = await DatabaseConnection.get();
+    const [admin] = UserFactory.with({ accessType: UserAccessType.ADMIN });
+    const [metadata1, metadata2, metadata3] = MerchFactory.createOptionMetadata(3);
+    const options = MerchFactory.optionsWith(
```
sumeet-bansal, line 121: Can't we call `MerchFactory.fakeOptions(3)` and let the factory handle the metadata? If not then we should.

tests/merch.test.ts
```diff
@@ -0,0 +1,303 @@
...
+      items: [item],
+    });
+
+    await new PortalState()
+      .createUsers([admin])
+      .createMerch([collection])
+      .write();
+
+    // change every option's type to a different but consistent one
+    options.forEach((option) => { option.metadata.type = 'new type'; });
```
sumeet-bansal, line 140: Let's do a `map` instead of an assignment in a for-each here.

api/validators/MerchStoreRequests.ts
shravanhariharan2: Agreed

services/MerchStoreService.ts
shravanhariharan2: I don't believe we can use just 1 parameter since the function is called for both `MerchItem` and `MerchandiseItemModel` objects, unless we define a common interface for the two

services/MerchStoreService.ts
shravanhariharan2: Not necessarily - I think we're missing an edge case for if we have a merch item with no variants is created but then its option is deleted with `deleteItemOption()`. When `createItemOption` is called to add a new item option (with no metadata, since `hasVariantsEnabled = false`), then line 189 prevents it from being added. I don't think the zero-option behavior is defined currently (i.e. should we let that happen? or not?), so I can ask Steven about that, or let me know what you think as well

services/MerchStoreService.ts
shravanhariharan2: Yes but we generally return the "public" version of the objects i.e. `PublicMerchItemOption`

shravanhariharan2: Wait never mind either way we don't need that call whoops, my b

tests/merch.test.ts
shravanhariharan2: Good catch, yes we can use `createOptions(n)` instead.

tests/merch.test.ts
shravanhariharan2: Do you mean something like this?
```ts
options = options.map((option) => ({ ...option, metadata: { type: 'new type' }}));
```
I feel like this is a little less cleaner than using `forEach` since it's a simple assigning vs. reconstructing an object. We'd still need the `options` array for the assertion below so I don't think we can map to anything else.

api/validators/MerchStoreRequests.ts
shravanhariharan2: Looked into class-validator and it looks like the convention's to keep it read-only for validation. It should be good to just have a default set in the controller and specify this behavior in the API docs.

tests/data/MerchFactory.ts
shravanhariharan2: It actually does include the pound sign, but yes I'll include it in the Postman docs

tests/data/MerchFactory.ts
shravanhariharan2: I feel like it's fine as is, if someone needs metadata with a specific type it's cause they need multiple options, so `createOptionMetadata` can be used there

tests/merch.test.ts
sumeet-bansal: The thing with `forEach` is that it takes a consumer of a single option and does something with the option, but the consumer shouldn't alter the option itself. If we're actually altering options, then we should do it within a `map` and not a `forEach`.

services/MerchStoreService.ts
```diff
@@ -107,47 +107,69 @@ export default class MerchStoreService {
...
+      if (options) {
+        const updatedOptionsByUuid = new Map(options.map((option) => [option.uuid, option]));
+        item.options.map((option) => {
+          if (!updatedOptionsByUuid.has(option.uuid)) return;
+          const updatedOption = updatedOptionsByUuid.get(option.uuid);
+          // 'quantity' is incremented instead of directly set to avoid concurrency issues with orders
+          // e.g. there's 10 of an item and someone adds 5 to stock while someone else orders 1
+          // so the merch store admin sets quantity to 15 but the true quantity is 14
+          if (updatedOption.addQuantity) updatedOption.addQuantity += option.quantity;
+          return MerchandiseItemOptionModel.merge(option, updatedOption);
```
sumeet-bansal, line 162: Does this work? We assign the new quantity to `updatedOption.addQuantity` and then merge the option and its update, so I think `addQuantity` would be ignored by `MerchandiseItemOptionModel` since the model doesn't have an `addQuantity` field.

Maybe instead we do something like
```suggestion
          updatedOption.addQuantity ??= 0; // or move this line to the controller?
          return MerchandiseItemOptionModel.merge(option, {
            ...updatedOption,
            quantity: option.quantity + updatedOption.addQuantity,
          });
```

services/MerchStoreService.ts
```diff
@@ -107,47 +107,69 @@ export default class MerchStoreService {
...
+        });
+      }
 
-      const updatedOptionsIds = new Set(updatedOptions.map((option) => option.uuid));
-      item.options = [...updatedOptions, ...item.options.filter((option) => !updatedOptionsIds.has(option.uuid))];
+      MerchStoreService.verifyItemHasValidOptions(MerchandiseItemModel.merge(item, changes));
```
sumeet-bansal, line 166: Nit: can we move `verifyItemHasValidOptions` to below this method?

services/MerchStoreService.ts
sumeet-bansal: Then we should throw in `deleteItemOption` if we're trying to delete the only option for an item, since that leaves the item in a bad state. Items should always have at least one option IMO.

tests/data/MerchFactory.ts
```diff
@@ -15,7 +16,21 @@ export class MerchFactory {
   }
 
   public static createOptions(n: number): MerchandiseItemOptionModel[] {
-    return FactoryUtils.create(n, MerchFactory.fakeOption);
+    const options = FactoryUtils.create(n, MerchFactory.fakeOption);
+    if (n === 1) return options;
+
+    const type = faker.datatype.hexaDecimal(10);
+    return options.map((o) => {
+      [o.metadata] = MerchFactory.optionMetadataWith({ type });
```
sumeet-bansal, line 24: What're the brackets for?

tests/data/MerchFactory.ts
```diff
@@ -38,22 +53,37 @@ export class MerchFactory {
...
   }
 
+  public static optionMetadataWith(...substitutes: Partial<MerchItemOptionMetadata>[]): MerchItemOptionMetadata[] {
+    return substitutes.map((sub) => {
+      const metadata = MerchFactory.fakeOptionMetadata();
+      return {
+        type: sub.type ?? metadata.type,
+        value: sub.value ?? metadata.value,
+        position: sub.position ?? metadata.position,
+      };
```
sumeet-bansal, line 63:
```suggestion
      return { ...metadata, ...sub };
```
This should do the same thing I think.

tests/data/PortalState.ts
```diff
@@ -57,6 +57,9 @@ export class PortalState {
     for (let u = 0; u < users.length; u += 1) {
       const user = users[u];
       user.email = user.email.toLowerCase();
+      if (user.points) user.credits = user.points * 100;
+      else if (user.credits) user.points = Math.floor(user.credits / 100);
```
sumeet-bansal, line 61: What's this line for? Why would a user have credits but not points?

sumeet-bansal: Ah I see. Maybe set `points: 50` instead in the test?

tests/merch.test.ts
```diff
@@ -88,3 +92,277 @@ describe('collection editing', () => {
...
+      .createUsers([admin])
+      .createMerch([collection])
+      .write();
+
+    option1.price -= 1000;
+    option1.quantity -= 5;
+    option1.discountPercentage = 15;
+    option2.price += 500;
+    option2.quantity += 5;
+    option1.discountPercentage = 0;
```
sumeet-bansal, line 178:
```suggestion
    option2.discountPercentage = 0;
```

tests/merch.test.ts
```diff
@@ -88,3 +92,277 @@ describe('collection editing', () => {
...
+
+    await ControllerFactory.merchStore(conn).editMerchItem(
+      { uuid: item.uuid },
+      {
+        merchandise: {
+          options: [option1, option2],
```
sumeet-bansal, line 184: Shouldn't we be passing in `addQuantity` not `finalQuantity` here?

tests/merch.test.ts
```diff
@@ -88,3 +92,277 @@ describe('collection editing', () => {
...
+    await new PortalState()
+      .createUsers([admin])
+      .createMerch([merchCollection])
+      .write();
+
+    await ControllerFactory
+      .merchStore(conn)
+      .createMerchItemOption({ uuid: merchItem.uuid }, { option: newOption }, admin);
+
+    const merchItemResponse = await ControllerFactory
```
sumeet-bansal, line 305: Nit: we can create a controller once within each test instead of using `ControllerFactory` every time.

api/validators/MerchStoreRequests.ts
```diff
@@ -46,6 +54,19 @@ export class MerchCollectionEdit implements IMerchCollectionEdit {
...
+  @IsDefined()
+  @IsNotEmpty()
+  type: string;
+
+  @IsDefined()
+  @IsNotEmpty()
+  value: string;
+
+  @Allow()
+  position?: number;
```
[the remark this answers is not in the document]
shravanhariharan2: I don't think we did, but yeah might as well make it required so that we don't have to validate it's existence every time a new option is created

services/MerchStoreService.ts
shravanhariharan2: You're right that it doesn't currently work, but isn't the fix to just reverse the assignment? i.e.
```ts
if (updatedOption.addQuantity) option.quantity+= updatedOption.addQuantity;
```
e.g. let's say there's 10 of an option in stock, and we add 5 more to it's stock, then `addQuantity` is 5 and and `quantity` is 10 so we set `quantity = 15` instead of `addQuantity = 15`.

I think the variable naming is slightly confusing currently since `updatedOption` isn't really an updated option, it just a partial object that contains some of the updates that need to be applied to the currently existing option. `option` is the object we need to update, and we should make that clearer in it's naming, so I'll rename the variables accordingly

services/MerchStoreService.ts
shravanhariharan2: I think it's being used in `createItem` as well, so that's why it's above. I think it's fine to leave it up there since otherwise, the reader has to scroll a bit to find the function when reading it in `createItem`

tests/data/MerchFactory.ts
shravanhariharan2: `optionMetadataWith` returns an array (e.g. `T[]`), the brackets make it so that `o.metadata` is of type `T`.

tests/data/PortalState.ts
shravanhariharan2: Yeah, I have it in case a developer sets credits in a test and doesn't set points. I think it may cause silent issues if the points <-> credits invariant is broken when credits are set without points, so I personally think it's good to have this assignment to save debugging time. That way we can also set credits at the same scale as the cost of an item option instead of having to calculate the scale each time (even more beneficial if the scale gets changed in the future).

services/MerchStoreService.ts
shravanhariharan2: I think one case where it should be fine to have 0 options is when it's hidden - I'm thinking within the admin UI, an admin should be able to freely edit options without having to be constrained to always keeping 1 option even if the item's hidden

services/MerchStoreService.ts
sumeet-bansal: That works too, but I didn't want to edit the original `option` object so I moved that operation inside the model merge.

tests/data/PortalState.ts
sumeet-bansal: 👍🏽

shravanhariharan2:
Well this PR got a little out of hand... sorry about that @sumeet-bansal. Didn't mean to add new validation logic for every revision of this PR, but I figured I wanted to add as much as possible before adding any new features to the store (e.g. order cart validation, fulfillment pickup date, etc.) so that all existing functionality would be airtight. Also there seemed to be a few hidden bugs in the merch factory / portal state code that took some time to figure out. Will do my best to keep future store PRs much, much shorter (in time and length) than this

shravanhariharan2:
Will fix the failing test tomorrow - for some reason it's not failing locally

shravanhariharan2:
what the fuck i literally added a print statement and it worked

sumeet-bansal (approved):
Main code looks good to go! Messaged you about the tests on Discord.

shravanhariharan2:
The hidden bugs were mainly:
* `BaseEntity.merge()` merging two arrays instead of replacing the base array with the changes array: https://github.com/acmucsd/membership-portal/pull/188/files#diff-402ac24d3b82f657bcb0c8a835fd7236167077bc1bb25b32189196724f9ebaeaR63-R70
* `PortalState.orderMerch()` not decrementing the quantities of the options ordered in the DB: https://github.com/acmucsd/membership-portal/pull/188/files#diff-c6064403a5d830d9a64406cd8314752c38454dfa7fa23c65e29f47e5ad4bc99fR141-R154
I was surprised why the latter was only caught now, since `sample.test.ts` asserts the option quantities matched. But yeah, those did take a bit of time to unravel

PR #189, opened 2021-05-26
Handling Items in Archived Collections

michl1001:


repositories/MerchStoreRepository.ts
```diff
@@ -80,11 +90,7 @@ export class MerchItemOptionRepository extends BaseRepository<MerchandiseItemOpt
     await qb
       .update()
       .set({ discountPercentage })
-      .where(`item IN ${qb.subQuery()
-        .select('merch.uuid')
-        .from(MerchandiseItemModel, 'merch')
-        .where('merch.collection = :collection')
-        .getQuery()}`)
+      .where('merch.collection = :collection')
```
sumeet-bansal, line 93: This subquery exists because item options don't have a `collection` property. Items have the `collection` property and item options have an `item` property. So, if we want to update all the item options in a collection, we need to get all the items in the collection first and then update all the item options where `item` is in that list of items. Another way of thinking about this is that we're doing a JOIN on items and item options and then filtering by `collection`.

You can verify this for yourself by writing a quick test that calls this method, and see what happens. The test should fail because we throw some sort of invalid SQL error.

tests/merch.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    const itemOpt = MerchFactory.fakeOption();
+    const [item] = MerchFactory.itemsWith({
+      options: [itemOpt],
+    });
+    const [collection] = MerchFactory.collectionsWith({
+      items: [item],
+      archived: false,
+    });
+    item.collection = collection;
+    itemOpt.item = item;
```
sumeet-bansal, line 31: @shravanhariharan2 didn't occur to me earlier but maybe we should set `item.collection` and `option.item` in the `MerchFactory` methods.

services/MerchStoreService.ts
```diff
@@ -72,11 +72,18 @@ export default class MerchStoreService {
...
         await Repositories
           .merchStoreItemOption(txn)
           .updateMerchItemOptionsInCollection(uuid, discountPercentage);
+      }
+      if (changes.archived !== undefined) {
+        await Repositories
+          .merchStoreItem(txn)
+          .updateMerchItemInCollection(uuid, true);
+      }
+      if (changes.discountPercentage !== undefined || changes.archived !== undefined) {
```
shravanhariharan2, line 86: If an admin wants to un-archive a collection, do we still want to set all items to be hidden?

services/MerchStoreService.ts
```diff
@@ -214,6 +221,15 @@ export default class MerchStoreService {
         throw new NotFoundError(`Missing: ${missingItems}`);
       }
 
+      // Checks that hidden items were not ordered
+      const hiddenItem = Array.from(itemOptions.values())
+        .filter((o) => o.item.hidden)
+        .map((o) => o.uuid);
+
```
shravanhariharan2, line 228: nit: rename to `hiddenItems`, since there could be more than 1 hidden items

services/MerchStoreService.ts
```diff
@@ -214,6 +221,15 @@ export default class MerchStoreService {
...
+      if (hiddenItem.length !== 0) {
+        throw new ForbiddenError(`Not allowed to order: ${hiddenItem}`);
+      }
+
```
shravanhariharan2, line 232: Let's give a bit more detail to why the user is not allowed to order the items - something like:
```ts
throw new ForbiddenError(`The following items are no longer available: ${hiddenItems}`)
```

tests/merch.test.ts
```diff
@@ -0,0 +1,63 @@
...
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('collection editing', () => {
+  test('Set collection to archive', async () => {
+    const conn = await DatabaseConnection.get();
+    const itemOpt = MerchFactory.fakeOption();
+    const [item] = MerchFactory.itemsWith({
```
shravanhariharan2, line 23: nit: rename `itemOpt` -> `itemOption`: we should avoid shorthand for clarity and consistency

tests/merch.test.ts
```diff
@@ -0,0 +1,63 @@
...
+});
+
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('collection editing', () => {
+  test('Set collection to archive', async () => {
+    const conn = await DatabaseConnection.get();
```
shravanhariharan2, line 21: This test case description doesn't really describe what we are asserting. We're essentially making that sure that hidden items cannot be ordered from, so let's make that clear in the name.

shravanhariharan2: Also, I think this test is asserting two different things, and can probably be split into two test cases related to archiving collections to keep each of the cases as simple and clear as possible. One can be for checking if getting an archived collection does or does not throw an error for admins and non-admins, and the other can be for checking if hidden items can be ordered from.

tests/merch.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        title: collection.title,
+        description: collection.description,
+        archived: true,
+      },
+    };
+
+    await merchStore.editMerchCollection({ uuid: collection.uuid }, editCollection, admin);
+
+    const error = `Not allowed to order: ${[itemOpt.uuid]}`;
+
```
shravanhariharan2, line 54: nit: can inline this error since it's only used once

tests/merch.test.ts
shravanhariharan2: I'm not sure how necessary it would be since all the read operations in our repositories do joins on the parent. @michl1001 does the test case fail if the parents aren't set here?

services/MerchStoreService.ts
shravanhariharan2: Wait I just realized, doesn't the code block above already check if hidden items were ordered? It seems to be filtering only non-hidden items and returning the items that weren't non-hidden (i.e. were hidden). Referring to this block: https://github.com/acmucsd/membership-portal/blob/4a9d27d7f90f90d953d8d0111c452f27d1c03dff/services/MerchStoreService.ts#L208

tests/merch.test.ts
sumeet-bansal: Those are nullable fields, so we need to set them.

services/MerchStoreService.ts
sumeet-bansal: `ForbiddenError` is usually reserved for permissions the user doesn't have, like accessing an admin-only route. This seems more like a `UserError` since the user placed an invalid order.

services/MerchStoreService.ts
```diff
@@ -72,11 +72,18 @@ export default class MerchStoreService {
       const currentCollection = await merchCollectionRepository.findByUuid(uuid);
       if (!currentCollection) throw new NotFoundError('Collection not found');
       let updatedCollection = await merchCollectionRepository.upsertMerchCollection(currentCollection, changes);
-      if (changes.discountPercentage) {
+      if (changes.discountPercentage !== undefined) {
```
sumeet-bansal, line 75: @shravanhariharan2 I think we talked about this at some point but we should probably move towards explicit optional/maybe types.

repositories/MerchStoreRepository.ts
```diff
@@ -53,6 +53,16 @@ export class MerchItemRepository extends BaseRepository<MerchandiseItemModel> {
     return this.repository.save(item);
   }
 
+  public async updateMerchItemInCollection(collection: string, hidden: boolean): Promise<void> {
```
sumeet-bansal, line 56: Nit: `updateMerchItemsInCollection` (plural). Also, instead of passing in a boolean that corresponds to some property, we should pass in some `changes` object that can be `{ hidden: true }`, like the other update methods.

sumeet-bansal: This is more flexible and also makes it more obvious what the caller is doing.

services/MerchStoreService.ts
michl1001: That code block only runs if the query doesn't find all options in the originalOrder so it doesn't actually run in this situation

michl1001: My understanding is that it warns the user of getting non-existent items

repositories/MerchStoreRepository.ts
```diff
@@ -53,11 +54,15 @@ export class MerchItemRepository extends BaseRepository<MerchandiseItemModel> {
     return this.repository.save(item);
   }
 
-  public async updateMerchItemInCollection(collection: string, hidden: boolean): Promise<void> {
+  public async updateMerchItemsInCollection(collection: string, changes: any): Promise<void> {
```
sumeet-bansal, line 57: We know `changes` should only contain fields from `MerchandiseItemModel` so we can declare it as `changes: Partial<MerchandiseItemModel>`. All the other methods should do something similar with `changes: Partial<T>`.

repositories/MerchStoreRepository.ts
```diff
@@ -53,11 +54,15 @@ export class MerchItemRepository extends BaseRepository<MerchandiseItemModel> {
...
+    if (Object.keys(changes).length === 0) {
+      throw new UserError('No changes sent');
+    }
```
sumeet-bansal, line 60: I appreciate that you're being defensive here but we're the ones calling this method so if no changes are sent, then that's entirely on us, so returning a UserError (4xx) is a bit disingenuous. TBH I think it's fine to remove this check since every time we call it we're manually setting the keys in `changes`. If we ever get to the point where we're passing in changes from the request (e.g. user sends a bunch of fields to change for all the merch items in the collection), then I think we can add this back.

repositories/MerchStoreRepository.ts
```diff
@@ -80,8 +95,7 @@ export class MerchItemOptionRepository extends BaseRepository<MerchandiseItemOpt
     await qb
       .update()
       .set({ discountPercentage })
-      .where(`item IN ${qb.subQuery()
-        .select('merch.uuid')
+      .where(`item IN ${qb.subQuery().select('merch.uuid')
         .from(MerchandiseItemModel, 'merch')
```
sumeet-bansal, line 99: Nit: reverting this change. I usually indent subqueries like this so it's easier to read.
```suggestion
      .where(`item IN ${qb.subQuery()
        .select('merch.uuid')
        .from(MerchandiseItemModel, 'merch')
```

tests/merch.test.ts
michl1001: The test case doesn't fail without them but I don't think the code driectly accesses those values

repositories/MerchStoreRepository.ts
michl1001: Got it, should be removed

services/MerchStoreService.ts
shravanhariharan2: What do you mean by explicitly optional/maybe? Do you mean having an `Optional<T>` type, and have `discountPercentage` be `Optional<number>`, where we can just do `if (changes.discountPercentage.isNull())` (or something similar)?

sumeet-bansal: Yes.

tests/merch.test.ts
```diff
@@ -0,0 +1,100 @@
...
+});
+
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('collection editing', () => {
+  test('Appropriate Permission Check for Getting Archived Items', async () => {
+    const conn = await DatabaseConnection.get();
```
shravanhariharan2, line 21: nit: test descriptions should describe some behavior that the test is asserting. 'Appropriate Permission Check for Getting Archived Items' doesn't really describe a behavior - maybe something like 'users cannot view archived collections, but admins can'?

tests/merch.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const merchStore = ControllerFactory.merchStore(conn);
+
+    const editCollection = {
+      collection: {
+        title: collection.title,
+        description: collection.description,
+        archived: true,
+      },
+    };
```
shravanhariharan2, line 47: In our API, when you edit objects generally you only need to specify the fields that are changed. This matches the PATCH paradigm of HTTP requests which we predominantly use in our API, as opposed to PUT (you can read about the differences in [this article](https://www.baeldung.com/http-put-patch-difference-spring)). So, simply:
```ts
const collectionEdit = {
  collection: {
    archived: true,
  },
};
```
should suffice.

shravanhariharan2: also nit: `editCollection` sounds like a method name since it implies an action is performed (ie. a collection is edited). Variable names should mostly be nouns, so I think `collectionEdit` makes the intention clearer.

tests/merch.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    await expect(merchStore.getOneMerchCollection({ uuid: collection.uuid }, user)).rejects.toThrow(ForbiddenError);
+
+    const result = await merchStore.getOneMerchCollection({ uuid: collection.uuid }, admin);
+    expect(result.collection.uuid).toEqual(collection.uuid);
+  });
+
+  test('Ordering Archived Items is not allowed', async () => {
+    const conn = await DatabaseConnection.get();
+    const itemOption = MerchFactory.fakeOption();
```
shravanhariharan2, line 59: nit: there aren't necessarily "archived items" - there are "hidden" items that were set as hidden by a collection being archived, so "ordering hidden items" more accurately matches the behavior than "ordering archived items" 

also nit: capital casing is not needed

tests/merch.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        { option: itemOption.uuid, quantity: 1 },
+      ],
+    }, user)).rejects.toThrow(`Not allowed to order: ${[itemOption.uuid]}`);
+
+    await expect(merchStore.placeMerchOrder({
+      order: [
+        { option: itemOption.uuid, quantity: 1 },
+      ],
+    }, admin)).rejects.toThrow(`Not allowed to order: ${[itemOption.uuid]}`);
+  });
```
shravanhariharan2, line 99: I don't think you need to check if both users and admins can't order hidden items, since the merch store requirements don't specify that there is a different behavior. So there's no reason that ordering between user/admin should be different in the first place

tests/merch.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const merchStore = ControllerFactory.merchStore(conn);
+
+    const editCollection = {
+      collection: {
+        title: collection.title,
+        description: collection.description,
+        archived: true,
+      },
+    };
```
shravanhariharan2, line 84: Same thing with only specifying fields that are changed.

shravanhariharan2 (commented):
Mostly small nitpicks and convention matching, otherwise looks good

shravanhariharan2:
Also, can you rename the PR title to something that describes what the PR adds/fixes (ie. setting an archived collection's items to hidden + tests)?

tests/merch.test.ts
```diff
@@ -0,0 +1,90 @@
...
+});
+
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('collection editing', () => {
+  test('Only Admin Can Getting Archived Items', async () => {
+    const conn = await DatabaseConnection.get();
```
shravanhariharan2, line 21:
```suggestion

describe('merch collections', () => {
  test('collections can only be archived by admins', async () => {
    const conn = await DatabaseConnection.get();
```

repositories/MerchStoreRepository.ts
shravanhariharan2: @michl1001 what Sumeet's saying is essentially to change the type of `changes` from `any` to `Partial<MerchandiseItemModel>` since we know the type of it. Can you take care of this here?

shravanhariharan2 (approved):
LGTM, but can you address the typing change of the `changes` field in `MerchStoreRepository::updateMerchItemsInCollection` before merging?

tests/merch.test.ts
michl1001: For the typing change, am I just making it a Partial type of the MerchItem?

As for the description, I don't think we really test who can archive collections. I changed the test description and can push it with the typing change

PR #197, opened 2021-09-18
Lint file on save in VSCode

sumeet-bansal:


StormFireFox1 (approved):
Pretty much a copy of my `settings.json`.

While we're at it, we could add `.editorconfig` to set automatic indentation for editors that don't have LSP's enabled.

shravanhariharan2 (approved):
LGTM, just make sure to add `build/` directory to the `files.exclude`

PR #198, opened 2021-09-19
Refactor tests 

sumeet-bansal:
- Made the tests a little more readable IMO by creating variables for all the objects being passed into controller calls.
- Removed a bunch of factory calls. The tests use the items and options created by the factory instead of creating metadata and options and items.
- Removed the `createXXX` and `XXXwith` methods from the factories, wherever possible. They weren't really needed and just cluttered the factory. Everything's done through `fakeXXX` now.
- Don't need to wrap everything in arrays for `PortalState` anymore for some functions, e.g. `createUsers`, `createEvents`.
- Added some extra methods to `PortalState` like `createMerchItem` and `createMerchOption` so we don't have to wrap those in collections and items in the tests.
- Left the non-merch tests alone for the most part, just updated their factory and `PortalState` calls.

### DIFF
--- a/services/MerchStoreService.ts
+++ b/services/MerchStoreService.ts
@@ -143,9 +143,8 @@ export default class MerchStoreService {
 
   private static allOptionsHaveValidMetadata(options: MerchItemOption[]): boolean {
     return options
-      .every((o) => !!o.price && o.price >= 0
-      && !!o.quantity && o.quantity >= 0
-      && !!o.metadata?.position && !!o.metadata?.position && o.metadata?.value);
+      .every((o) => !!o.metadata
+      && !!o.metadata?.type && !!o.metadata?.position && o.metadata?.value);
   }
 
   private static hasMultipleOptionTypes(options: MerchItemOption[]): boolean {
--- a/services/MerchStoreService.ts
+++ b/services/MerchStoreService.ts
@@ -4,7 +4,6 @@ import { NotFoundError, ForbiddenError } from 'routing-controllers';
 import { EntityManager } from 'typeorm';
 import { difference, flatten, intersection } from 'underscore';
 import * as moment from 'moment';
-import { MerchItemOption } from '../api/validators/MerchStoreRequests';
 import { MerchandiseItemOptionModel } from '../models/MerchandiseItemOptionModel';
 import {
   Uuid,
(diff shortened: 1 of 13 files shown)


tests/data/MerchFactory.ts
```diff
@@ -60,34 +15,36 @@ export class MerchFactory {
...
-    // so fakeModel.options array would have both substitute and fake items if items
-    // are created in the MerchandiseItemModel.create() call.
-    if (!fakeModel.options) fakeModel.options = MerchFactory.createOptions(numOptions);
-    // explicitly set the option.item field for all item's options since it is required to be non-null
-    fakeModel.options.map((option) => ({ ...option, item: fakeModel }));
-    return fakeModel;
+
+    // merging arrays returns a union of fake.items and substitute.items so only create
+    // fake.options if the substitute doesn't provide any
+    if (!substitute?.options) {
```
shravanhariharan2, line 41:
```suggestion

    // merging arrays returns a union of fake.options and substitute.options so only create
    // fake.options if the substitute doesn't provide any
    if (!substitute?.options) {
```

tests/data/MerchFactory.ts
```diff
@@ -101,12 +58,31 @@ export class MerchFactory {
...
+    const fake = {
       type: faker.datatype.hexaDecimal(10),
       value: faker.datatype.hexaDecimal(10),
       position: FactoryUtils.getRandomNumber(1, 10),
     };
+    const sub = substitute ?? {};
+    return {
+      ...fake,
+      ...sub,
+    };
```
shravanhariharan2, line 77: 🤯

shravanhariharan2 (approved):
Big fan of the factory param changes and the merch factory rewrite. Some minor comments

services/MerchStoreService.ts
```diff
@@ -133,12 +133,21 @@ export default class MerchStoreService {
...
     }
   }
 
-  private static hasMultipleOptionTypes(options: MerchItemOption[]) {
+  private static allOptionsHaveValidMetadata(options: MerchItemOption[]): boolean {
+    return options
+      .every((o) => !!o.metadata
+      && !!o.metadata?.type && !!o.metadata?.position && o.metadata?.value);
+  }
+
```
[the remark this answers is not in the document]
sumeet-bansal: Yeah good point, first check should be enough then.

PR #202, opened 2021-09-22
Implements rudimentary verification route

michl1001:
Implements verification route for orders (no issue???)

### DIFF
--- a/api/controllers/MerchStoreController.ts
+++ b/api/controllers/MerchStoreController.ts
@@ -26,6 +26,7 @@ import {
   GetOneMerchOrderResponse,
   GetAllMerchOrdersResponse,
   PlaceMerchOrderResponse,
+  VerifyMerchOrderResponse,
   EditMerchOrderResponse,
   CreateMerchItemOptionResponse,
   DeleteMerchItemOptionResponse,
@@ -40,6 +41,7 @@ import {
   CreateMerchItemRequest,
   EditMerchItemRequest,
   PlaceMerchOrderRequest,
+  VerifyMerchOrderRequest,
   FulfillMerchOrderRequest,
   CreateMerchItemOptionRequest,
 } from '../validators/MerchStoreRequests';
@@ -175,6 +177,21 @@ export class MerchStoreController {
     if (originalOrder.length !== numUniqueUuids) throw new BadRequestError('There are duplicate items in this order');
     const order = await this.merchStoreService.placeOrder(originalOrder, user);
     return { error: null, order };
+  }
+
+  @Post('/order/verification')
+  async verifyMerchOrder(@Body() verifyOrderRequest: VerifyMerchOrderRequest,
+    @AuthenticatedUser() user: UserModel): Promise<VerifyMerchOrderResponse> {
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
+    const originalOrder = verifyOrderRequest.order.filter((oi) => oi.quantity > 0);
+    const orderIsEmpty = originalOrder.reduce((x, n) => x + n.quantity, 0) === 0;
+    if (orderIsEmpty) throw new UserError('There are no items in this order');
+    const numUniqueUuids = (new Set(originalOrder.map((oi) => oi.option))).size;
+    if (originalOrder.length !== numUniqueUuids) throw new BadRequestError('There are duplicate items in this order');
+
+    const totalCost = await this.merchStoreService.verifyOrder(originalOrder,user);
+
+    return { error: null, totalCost };
   }
 
   @Patch('/order')
--- a/models/OrderModel.ts
+++ b/models/OrderModel.ts
@@ -8,7 +8,7 @@ import {
   ManyToOne,
   JoinColumn,
 } from 'typeorm';
-import { Uuid, PublicOrder } from '../types';
+import { Uuid, PublicOrder, MerchItemOptionAndQuantity } from '../types';
 import { UserModel } from './UserModel';
 import { OrderItemModel } from './OrderItemModel';
 
@@ -40,5 +40,11 @@ export class OrderModel extends BaseEntity {
       orderedAt: this.orderedAt,
       items: this.items.map((oi) => oi.getPublicOrderItem()),
     };
+  }
+  public getItemsAndQuantity(): MerchItemOptionAndQuantity[] {
+    
+
+    
+    return []
   }
 }
--- a/services/MerchStoreService.ts
+++ b/services/MerchStoreService.ts
@@ -291,6 +291,68 @@ export default class MerchStoreService {
       await user.reload();
       const merchItemOptionRepository = Repositories.merchStoreItemOption(txn);
       const itemOptions = await merchItemOptionRepository.batchFindByUuid(originalOrder.map((oi) => oi.option));
+
+      const totalCost = await this.verifyOrder(originalOrder,user);
+
+      const merchOrderRepository = Repositories.merchOrder(txn);
+
+      // if all checks pass, the order is placed
+      const createdOrder = await merchOrderRepository.createMerchOrder(OrderModel.create({
+        user,
+        totalCost,
+        items: flatten(originalOrder.map((optionAndQuantity) => {
+          const option = itemOptions.get(optionAndQuantity.option);
+          const quantityRequested = optionAndQuantity.quantity;
+          return Array(quantityRequested).fill(OrderItemModel.create({
+            option,
+            salePriceAtPurchase: option.getPrice(),
+            discountPercentageAtPurchase: option.discountPercentage,
+          }));
+        })),
+      }));
+
+      const activityRepository = Repositories.activity(txn);
+      await activityRepository.logActivity({
+        user,
+        type: ActivityType.ORDER_MERCHANDISE,
+        description: `Order ${createdOrder.uuid}`,
+      });
+
+      await Promise.all(originalOrder.map(async (optionAndQuantity) => {
+        const option = itemOptions.get(optionAndQuantity.option);
+        const updatedQuantity = option.quantity - optionAndQuantity.quantity;
+        return merchItemOptionRepository.upsertMerchItemOption(option, { quantity: updatedQuantity });
+      }));
+
+      const userRepository = Repositories.user(txn);
+      userRepository.upsertUser(user, { credits: user.credits - totalCost });
+
+      return [createdOrder, itemOptions];
+    });
+
+    const orderConfirmation = {
+      items: originalOrder.map((oi) => {
+        const option = merchItemOptions.get(oi.option);
+        const { item } = option;
+        return {
+          ...item,
+          quantityRequested: oi.quantity,
+          salePrice: option.getPrice(),
+          total: oi.quantity * option.getPrice(),
+        };
+      }),
+      totalCost: order.totalCost,
+    };
+    this.emailService.sendOrderConfirmation(user.email, user.firstName, orderConfirmation);
+
+    return order.getPublicOrder();
+  }
+
+  public async verifyOrder(originalOrder: MerchItemOptionAndQuantity[], user: UserModel): Promise<number>{
+    const [verified,totalCost] = await this.transactions.readWrite(async (txn) => {
+      await user.reload();
+      const merchItemOptionRepository = Repositories.merchStoreItemOption(txn);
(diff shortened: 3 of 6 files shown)


michl1001:
Right now, the service method just returns the totalCost (outside of errors for invalid orders, this is the main significant output I could spot), let me know if there should be other thing in the response

api/controllers/MerchStoreController.ts
```diff
@@ -177,6 +179,21 @@ export class MerchStoreController {
...
+  async verifyMerchOrder(@Body() verifyOrderRequest: VerifyMerchOrderRequest,
+    @AuthenticatedUser() user: UserModel): Promise<ApiResponse> {
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
+    const originalOrder = verifyOrderRequest.order.filter((oi) => oi.quantity > 0);
+    const orderIsEmpty = originalOrder.reduce((x, n) => x + n.quantity, 0) === 0;
+    if (orderIsEmpty) throw new UserError('There are no items in this order');
+    const numUniqueUuids = (new Set(originalOrder.map((oi) => oi.option))).size;
+    if (originalOrder.length !== numUniqueUuids) throw new BadRequestError('There are duplicate items in this order');
+
+    await this.merchStoreService.verifyOrder(originalOrder, user);
```
shravanhariharan2, line 192: nit: you can probably create a private function `validateMerchOrderRequest` or something similar that takes in an order and does this validation. `placeOrder()` can use this method as well so that there isn't logic duplication happening

models/OrderModel.ts
```diff
@@ -41,4 +41,8 @@ export class OrderModel extends BaseEntity {
       items: this.items.map((oi) => oi.getPublicOrderItem()),
     };
   }
+
+  public getItemsAndQuantity(): MerchItemOptionAndQuantity[] {
+    return [];
+  }
```
shravanhariharan2, line 47: What's this method for?

services/MerchStoreService.ts
```diff
@@ -291,6 +291,68 @@ export default class MerchStoreService {
...
+    this.emailService.sendOrderConfirmation(user.email, user.firstName, orderConfirmation);
+
+    return order.getPublicOrder();
+  }
+
+  public async verifyOrder(originalOrder: MerchItemOptionAndQuantity[], user: UserModel): Promise<number> {
+    const totalOrderCost = await this.transactions.readWrite(async (txn) => {
+      await user.reload();
+      const merchItemOptionRepository = Repositories.merchStoreItemOption(txn);
+      const itemOptions = await merchItemOptionRepository.batchFindByUuid(originalOrder.map((oi) => oi.option));
```
shravanhariharan2, line 355: nit: you can just do `return this.transactions.readWrite(async (txn))...` instead of awaiting the call and returning the same value returned by the async call, since we're returning a `Promise` anyway.

services/MerchStoreService.ts
```diff
@@ -291,6 +291,68 @@ export default class MerchStoreService {
       await user.reload();
       const merchItemOptionRepository = Repositories.merchStoreItemOption(txn);
       const itemOptions = await merchItemOptionRepository.batchFindByUuid(originalOrder.map((oi) => oi.option));
+
+      const totalCost = await this.verifyOrder(originalOrder, user);
+
+      const merchOrderRepository = Repositories.merchOrder(txn);
```
shravanhariharan2, line 297: The naming of `verifyOrder` makes it slightly confusing as to why a number is returned from the method, since verification implies that an input is either valid or not. So it might make sense to rename this method to `verifyOrderAndReturnTotalCost`, but that might be a bit overkill. @sumeet-bansal any thoughts on this? We could also have a separate method `calculateTotalCostOfOrder` or something similar and use that within the verify method and the placeOrder method. Or just document what the return value means in a doc comment.

api/controllers/MerchStoreController.ts
```diff
@@ -14,6 +14,7 @@ import {
 import PermissionsService from '../../services/PermissionsService';
 import { UserAuthentication } from '../middleware/UserAuthentication';
 import {
+  ApiResponse,
   GetOneMerchCollectionResponse,
```
shravanhariharan2, line 18: nit: I don't think we should be importing `ApiResponse`, since the class was meant to be a general response type that other responses extend, and shouldn't be exposed to controller classes. You can create a new response type that just extends `ApiResponse` without any body, i.e.
```ts
export interface VerifyMerchOrderResponse extends ApiResponse {}
```

models/OrderModel.ts
michl1001: OOOPS, my bad, this was for a prior implementation that I realized we didn't need, I'll delete it

sumeet-bansal:
Can you update the PR description with some info about this PR, e.g. why we're making this change and a high-level overview of what the changes are?

services/MerchStoreService.ts
```diff
@@ -291,6 +291,68 @@ export default class MerchStoreService {
...
+        };
+      }),
+      totalCost: order.totalCost,
+    };
+    this.emailService.sendOrderConfirmation(user.email, user.firstName, orderConfirmation);
+
+    return order.getPublicOrder();
+  }
+
+  public async verifyOrder(originalOrder: MerchItemOptionAndQuantity[], user: UserModel): Promise<number> {
```
sumeet-bansal, line 351: `placeOrder` creates a transaction and calls `verifyOrder` from within that transaction. `verifyOrder` also creates a transaction though. Not sure if `verifyOrder` uses the `placeOrder` transaction or if we end up in a deadlock or what happens (tests pass though so I assume it's the former), but we should have two `verifyOrder` methods I think: one that's called from the controller and creates a transaction, and another that accepts a transaction as a param and is called by `placeOrder` and the other `verifyOrder`.

services/MerchStoreService.ts
```diff
@@ -291,6 +291,68 @@ export default class MerchStoreService {
...
+      await Promise.all(originalOrder.map(async (optionAndQuantity) => {
+        const option = itemOptions.get(optionAndQuantity.option);
+        const updatedQuantity = option.quantity - optionAndQuantity.quantity;
+        return merchItemOptionRepository.upsertMerchItemOption(option, { quantity: updatedQuantity });
+      }));
+
+      const userRepository = Repositories.user(txn);
+      userRepository.upsertUser(user, { credits: user.credits - totalCost });
+
+      return [createdOrder, itemOptions];
```
sumeet-bansal, line 330: I know I wrote this part but this is terrible. We should probably have some proper return type for this.

michl1001: Is there a good description for the return type here?

shravanhariharan2: Maybe `MerchOrderAndItemOptions`? You can add this type to `types/internal/index.ts`.

api/controllers/MerchStoreController.ts
```diff
@@ -177,6 +179,21 @@ export class MerchStoreController {
     return { error: null, order };
   }
 
+  @Post('/order/verification')
+  async verifyMerchOrder(@Body() verifyOrderRequest: VerifyMerchOrderRequest,
```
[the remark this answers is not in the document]
michl1001: We changed it to its own type instead of `ApiResponse`, Don't know if having the boolean would be needed though since an error would be thrown in any event there is a valid order

services/MerchStoreService.ts
michl1001: Are there any good resources on how to pass transactions as parameters? Haven't been able to find something that applies to this situation

sumeet-bansal: What we're trying to do here is restructure the code so that we don't try opening a transaction within a transaction. What's going on right now is
```typescript
// the controller calls this
async placeOrder() {
  this.transactions.readWrite(txn -> {
    verifyOrder();
  });
}

// the controller and placeOrder call this
async verifyOrder() {
  this.transactions.readWrite(txn -> {
    // do something
  });
}
```
`placeOrder` creates a transaction and calls `verifyOrder` from within that transaction. `verifyOrder` also creates a transaction though, so we try opening a transaction within a transaction. Best case, the code's a little confusing to read but the library we use to call our database handles it intelligently and doesn't try opening up another transaction (which seems to be the case); worst case, something gets stuck and prevents the backend from crashing and from processing other requests.

Instead, we should have two `verifyOrder` methods:
```typescript
// the controller calls this
async placeOrder() {
  this.transactions.readWrite(txn -> {
    verifyOrderInternal(txn);
  });
}

// the controller calls this
async verifyOrder() {
  this.transactions.readWrite(txn -> {
    verifyOrderInternal(txn);
  });
}

// placeOrder and verifyOrder call this
async verifyOrderInternal(txn: TransactionalEntityManager) {
 // do something
}
```
Here's an example from our repo that passes the transaction as a method parameter: [`TransactionsManager`](https://github.com/acmucsd/membership-portal/blob/master/repositories/index.ts#L64). Going off that, I think `TransactionalEntityManager` is the correct type, but you can look at the type definitions in your IDE to verify.

sumeet-bansal:
Also, @michl1001 if you merge master into this branch then you'll pick up #197 which adds some VSCode configuration to auto-lint a file whenever you save it.

api/controllers/MerchStoreController.ts
```diff
@@ -167,16 +170,22 @@ export class MerchStoreController {
...
+    @AuthenticatedUser() user: UserModel): Promise<VerifyMerchOrderResponse> {
+    const originalOrder = this.verifyMerchOrderRequest(verifyOrderRequest.order, user);
+
+    await this.merchStoreService.verifyOrder(originalOrder, user);
+
+    return { error: null };
+  }
```
shravanhariharan2, line 187: nit: I don't think the extra lines are needed, we can keep the same convention that the other routes have with spacing.

shravanhariharan2 (approved):
LGTM, one small nitpick, and make sure to merge the pickup event changes from master

api/controllers/MerchStoreController.ts
```diff
@@ -188,4 +197,16 @@ export class MerchStoreController {
     await this.merchStoreService.updateOrderItems(fulfillOrderRequest.items);
     return { error: null };
   }
+
+  private verifyMerchOrderRequest(orderRequest:MerchItemOptionAndQuantity[],
+    user:UserModel):MerchItemOptionAndQuantity[] {
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
```
sumeet-bansal, line 203: Can you move this permissions check out of this method and back to the endpoints? This method should only validate the request.

api/controllers/MerchStoreController.ts
```diff
@@ -188,4 +197,16 @@ export class MerchStoreController {
...
+    const originalOrder = orderRequest.filter((oi) => oi.quantity > 0);
+    const orderIsEmpty = originalOrder.reduce((x, n) => x + n.quantity, 0) === 0;
+    if (orderIsEmpty) throw new UserError('There are no items in this order');
+    const numUniqueUuids = (new Set(originalOrder.map((oi) => oi.option))).size;
+    if (originalOrder.length !== numUniqueUuids) throw new BadRequestError('There are duplicate items in this order');
```
sumeet-bansal, line 208: IMO this is more of a UserError than a BadRequestError. I think BadRequestError should be for invalid requests, and this is something that'd be a UserError since it's a valid request. @shravanhariharan2 yay/nay?
```suggestion
    if (originalOrder.length !== numUniqueUuids) throw new UserError('There are duplicate items in this order');
```

services/MerchStoreService.ts
```diff
@@ -400,6 +349,72 @@ export default class MerchStoreService {
     return order.getPublicOrder();
   }
 
+  private async verifyOrderWithTransaction(originalOrder: MerchItemOptionAndQuantity[],
```
sumeet-bansal, line 352: Nit: can you move this under `verifyOrder`?

services/MerchStoreService.ts
```diff
@@ -457,4 +472,12 @@ export default class MerchStoreService {
     }
     return requestedQuantitiesByMerchItem;
   }
+
+  private getTotalCost(order:MerchItemOptionAndQuantity[], itemOptions:Map<string, MerchandiseItemOptionModel>):number {
```
sumeet-bansal, line 476: Nit: this can be `private static totalCost(...)` since it only uses its parameters and no class variables.

sumeet-bansal (approved):

Issue #204, opened 2021-09-28
Implement deletion of order pickup events


shravanhariharan2:
Deletion of order pickup events should:
* Send an email to all users who have orders scheduled for that pickup event
* Set `OrderModel.pickupEvent = null` for all orders
* Set `OrderModel.status = OrderStatus.PICKUP_EVENT_CANCELLED` (when status field is implemented)
* Delete the event

Issue #203, opened 2021-09-25
Split merch.test.ts into merch-store.test.ts and merch-order.test.ts


shravanhariharan2:
For the same reason as https://github.com/acmucsd/membership-portal/issues/200. The test file is getting quite large with continued additions of store features, so splitting the file into two should improve readability.

PR #207, opened 2021-10-01
Implement order states for order cancellation, fulfillment, pickup misses, and pickup cancellation

### LINKED ISSUE #205: Implement order state system
- Add order states to the order model
- Rewrite order routes to utilize states correctly

shravanhariharan2:
Closes #205, #204, #203, #218, and #217

Details for this feature can be found in the Merch Store Specification on Notion: https://www.notion.so/acmucsd/Merch-Store-Specification-2021-d0d5fa4d16de472aa7e13534e060f8ec (under the "Orders" section of the API routes)


shravanhariharan2:
Tests still need to be written, but I'll publish this PR for now since it's pretty large and might need a bit of time to look at

services/MerchStoreService.ts
```diff
@@ -410,40 +414,176 @@ export default class MerchStoreService {
...
+      return orderRespository.upsertMerchOrder(order, { pickupEvent });
+    });
+  }
+
+  /**
+   * Maps an item's option to its price at purchase and quantity ordered by the user
+   * @param order order
+   * @returns map of item option to its price at purchase and quantity ordered by the user
+   */
+  private static getOptionPriceAndQuantitiesFromOrder(order: OrderModel): Map<string, OrderItemPriceAndQuantity> {
```
sumeet-bansal, line 487: Nit: if we're working with maps we should include the key in the name.
```suggestion
  private static getPriceAndQuantityByOption(order: OrderModel): Map<Uuid, OrderItemPriceAndQuantity> {
```

services/MerchStoreService.ts
```diff
@@ -410,40 +414,176 @@ export default class MerchStoreService {
...
+    const optionToPriceAndQuantity = new Map<string, OrderItemPriceAndQuantity>();
+    order.items.forEach((oi) => {
+      const { uuid } = oi.option;
+      if (optionToPriceAndQuantity.has(oi.option.uuid)) {
```
sumeet-bansal, line 491:
```suggestion
      if (optionToPriceAndQuantity.has(uuid)) {
```

services/MerchStoreService.ts
```diff
@@ -410,40 +414,176 @@ export default class MerchStoreService {
...
+  }
+
+  /**
+   * Maps an item's option to its price at purchase and quantity ordered by the user
+   * @param order order
+   * @returns map of item option to its price at purchase and quantity ordered by the user
+   */
+  private static getOptionPriceAndQuantitiesFromOrder(order: OrderModel): Map<string, OrderItemPriceAndQuantity> {
+    const optionToPriceAndQuantity = new Map<string, OrderItemPriceAndQuantity>();
+    order.items.forEach((oi) => {
```
sumeet-bansal, line 489: Does `forEach` process all the elements concurrently?

sumeet-bansal: Seems like it's blocking. IMO we should use for-each for pure functions and prefer for-loops if we're editing maps though. What do you think?

services/MerchStoreService.ts
```diff
@@ -410,40 +414,176 @@ export default class MerchStoreService {
...
+      const { uuid } = oi.option;
+      if (optionToPriceAndQuantity.has(oi.option.uuid)) {
+        optionToPriceAndQuantity.set(uuid, {
+          quantity: optionToPriceAndQuantity.get(uuid).quantity + 1,
```
sumeet-bansal, line 493: Nit: we can put `optionToPrice.get(uuid)` in a variable since we reference it twice.

services/MerchStoreService.ts
```diff
@@ -410,40 +414,176 @@ export default class MerchStoreService {
...
+   */
+  private static async buildOrderUpdateInfo(order: OrderModel, txn: EntityManager) {
+    // maps an item option to its price at purchase and quantity ordered by the user
+    const optionPricesAndQuantities = MerchStoreService.getOptionPriceAndQuantitiesFromOrder(order);
+    const itemOptionsOrdered = Array.from(optionPricesAndQuantities.keys());
+    const itemOptionByUuid = await Repositories
+      .merchStoreItemOption(txn)
+      .batchFindByUuid(itemOptionsOrdered);
+
+    const orderUpdateInfo = {
```
sumeet-bansal, line 524:
```suggestion
    return {
```

services/MerchStoreService.ts
```diff
@@ -502,16 +642,24 @@ export default class MerchStoreService {
...
     return this.transactions.readWrite(async (txn) => {
       const orderPickupEventRepository = Repositories.merchOrderPickupEvent(txn);
+      const orderRepository = Repositories.merchOrder(txn);
       const pickupEvent = await orderPickupEventRepository.findByUuid(uuid);
       if (pickupEvent.orders.length > 0) {
         throw new UserError('Cannot delete an order pickup event that has orders assigned to it');
       }
-      // Manually set all the orders' pickup events to null before deleting event
+      // email order cancellation for order, update order status, and set pickupEvent to null before deleting
+      pickupEvent.orders.forEach(async (order) => {
```
sumeet-bansal, line 657: If `forEach` is blocking then let's do these concurrently so we're not sending emails one at a time and blocking on each one.

api/controllers/MerchStoreController.ts
```diff
@@ -181,15 +187,32 @@ export class MerchStoreController {
...
+  @Patch('/order/:uuid')
+  async editMerchOrder(@Params() params: UuidParam,
+    @Body() editOrderRequest: EditMerchOrderRequest,
     @AuthenticatedUser() user: UserModel): Promise<EditMerchOrderResponse> {
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
+    const { status, pickupEvent } = editOrderRequest.order;
+    // status and pickup event are not allowed to be updated at the same time,
+    // since they are not logically related, and thus should never have to be updated at the same request
+    if (status && pickupEvent) throw new UserError('Only status or pickupEvent can be updated at once, not both');
+    // members are only allowed to cancel orders. store admins can perform any operation
```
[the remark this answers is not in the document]
shravanhariharan2: > Can an order be canceled and then re-placed? Should that be allowed? If it is, then we need to update quantities and all that.

No, this isn't allowed per store requirements.

> Should this endpoint be split into separate member and admin edits? e.g. DELETE /order/:uuid for members canceling orders and PATCH /order/:uuid for admins canceling orders or something?

I like the idea of splitting the giant `PATCH /order/:uuid` route into separate routes like `/cancellation` or `/missed` or `/fulfillment`. I think in the case with member vs. admin edit, it should be fine to keep them in the same route, since we do the same sort of split with other routes. Also I just realized I wrote some buggy ass code where if an admin cancells a user's order, then the admin gets refunded LMAO

api/controllers/MerchStoreController.ts
```diff
@@ -181,15 +187,32 @@ export class MerchStoreController {
...
+    if (status && status !== OrderStatus.CANCELLED && !PermissionsService.canEditMerchStore(user)) {
+      throw new ForbiddenError('Members can only cancel orders');
+    }
+    await this.merchStoreService.editOrder(params.uuid, editOrderRequest.order, user);
+    return { error: null };
+  }
+
+  @Post('/order/:uuid/fulfillment')
```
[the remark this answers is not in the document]
shravanhariharan2: I'm a fan of it too - personally I'm of the opinion of the 'convention' for REST API naming is whatever naming we establish (since I don't think there's a set of 'naming conventions' for APIs). So I don't think we should feel too bad about having verbs in the name - especially since it also distinguishes 'actions' (ie. reset password or fulfill orders) from 'resource operations' (ie. editing a field of a model).

api/validators/MerchStoreRequests.ts
```diff
@@ -300,3 +320,8 @@ export class EditOrderPickupEventRequest implements IEditOrderPickupEventRequest
   @IsDefined()
   pickupEvent: OrderPickupEventEdit;
 }
+
+export class GetOrderQueryParams implements IGetOrderQueryParams {
```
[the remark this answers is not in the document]
shravanhariharan2: The route is also for users to view their cancelled/fulfilled/pending/missed orders to view for themselves.

services/MerchStoreService.ts
shravanhariharan2: Yes, forEach is a blocking call - https://stackoverflow.com/questions/5050265/javascript-node-js-is-array-foreach-asynchronous

> IMO we should use for-each for pure functions and prefer for-loops if we're editing maps though. What do you think?
That makes sense to me - since forEach does by nature look more functional, we can use regular for-loops to emphasize that there's some state being changed within the iteration

services/MerchStoreService.ts
sumeet-bansal: Looked at the email template to see if we could just send the same email to all the orderers, and it seems not, but it looks like we don't specify a URL here? Or am I misreading the HTML?

https://github.com/acmucsd/membership-portal/blob/OrderStates/templates/orderPickupCancelled.ejs#L120

sumeet-bansal: And can we send the same email to all the orderers? e.g. "Hey your order got canceled" and then instead of showing the order in the email just link to it?

api/controllers/MerchStoreController.ts
```diff
@@ -181,15 +187,32 @@ export class MerchStoreController {
...
+  async fulfillMerchOrder(@Params() params: UuidParam, @Body() fulfillOrderRequest: FulfillMerchOrderRequest,
+    @AuthenticatedUser() user: UserModel): Promise<FulfillMerchOrderResponse> {
     if (!PermissionsService.canFulfillMerchOrders(user)) throw new ForbiddenError();
```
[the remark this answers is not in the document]
shravanhariharan2: I'm hesitant to add hardcoded roles into the backend now because there is an incoming roles rewrite where the ACM admins can create custom roles and define permission levels for each, so for now it might be alright to just have `ADMIN` roles handle this. 

@steets250 do we know have an estimate on wen the roles rewrite might take place? If it's a few months out, we might actually want to define one or more roles for store so that we don't have multiple people on full `ADMIN` accounts.

services/MerchStoreService.ts
```diff
@@ -410,40 +414,176 @@ export default class MerchStoreService {
...
+   * at least 2 days in the future.
+   *
+   * @param uuid order uuid
+   * @param orderEdit order edits
+   * @param user user requesting the edit
+   * @returns the edited order
+   */
+  public async editOrder(uuid: Uuid, orderEdit: MerchOrderEdit, user: UserModel): Promise<OrderModel> {
+    return this.transactions.readWrite(async (txn) => {
+      const orderRespository = Repositories.merchOrder(txn);
```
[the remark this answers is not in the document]
shravanhariharan2: good eyes

services/MerchStoreService.ts
shravanhariharan2: Good catch, I meant to ping Steven about what URL to use there but I never ended up doing that.

> And can we send the same email to all the orderers? e.g. "Hey your order got canceled" and check the portal to see which order? Kind of a dick move though, but it means we only send one email.

Haha interesting idea, I think it might be better UX if we send each person a separate email though so that they can see what specific items were in their cancelled order and so they each get their own URL to reschedule. Might need to look if there's some kind of rate limiting on Sendgrid's end to make sure mass mailing won't fail for some emails.

templates/orderPickupCancelled.ejs
```diff
@@ -0,0 +1,152 @@
...
+                                                                    <p><%= order.items[i].itemName %></p>
+                                                                    <p><small><%= order.items[i].salePrice %> x <%= order.items[i].quantityRequested %> = <%= order.items[i].total %></small></p>
+                                                <% } %>
+                                                <h1><%= order.totalCost %></h1>
+                                        <p style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-bottom: 15px;">Please reschedule a new pickup time for your order. If you are not able to schedule a pickup time
+                                          before the end of the quarter, then your order will be cancelled and fully refunded to your account.
+
+                                          You can reschedule here: https://
+                                        </p>
+                                        <p style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-bottom: 15px;">Regards, <br />ACM UCSD</p>
```
shravanhariharan2, line 122: @steets250 let me know what the URL for rescheduling an order should be on the frontend.

api/controllers/MerchStoreController.ts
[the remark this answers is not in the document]
steets250: Yeah, actually that's probably a good thing to add before we start adding other new portal features to the backend. I'll work on brainstorming roles + perms this week, and maybe we can further refine at the next work session.

templates/orderPickupCancelled.ejs
steets250: https://members.acmucsd.com/store/orders?uuid=${order.uuid}

api/validators/MerchStoreRequests.ts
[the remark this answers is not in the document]
sumeet-bansal: Frontend could have some pop-up like "are you really really sure". Also we should prob log picking up orders, canceling orders, and canceling events as activities.

shravanhariharan2:
Initial testing is done, @sumeet-bansal feel free to review the existing tests. TODOs still are:
* Implement `POST /order/cleanup` route for admins to refund all outstanding orders
* Log activities for (updating, cancelling, missing) pickup events, fulfilling partial/full orders
* Finalize email templates (will hand this off to @steets250 once this is merged into master, can use testing API for manually testing the emails)

migrations/0030-add-order-status-field.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationInterface, QueryRunner, TableColumn, TableIndex } from 'typeorm';
+
+const TABLE_NAME = 'Orders';
+
+export class AddOrderStatusField1633030219180 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn(TABLE_NAME, new TableColumn({
+      name: 'status',
+      type: 'varchar(255)',
+      default: '\'PLACED\'',
```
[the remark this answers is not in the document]
shravanhariharan2: image

The migration fails without the quotation marks, because PG thinks it's a column otherwise

tests/merch.order.test.ts
```diff
@@ -0,0 +1,721 @@
...
+    const affordableOption2 = MerchFactory.fakeOption({
+      quantity: 1,
+      price: 3000,
+      discountPercentage: 0,
+    });
+    const pickupEvent = MerchFactory.fakeFutureOrderPickupEvent();
+
+    await new PortalState()
+      .createUsers(member)
+      .createMerchItemOption(affordableOption1)
```
sumeet-bansal, line 187: Same here with `createMerchItemOptions`

tests/merch.order.test.ts
```diff
@@ -0,0 +1,721 @@
...
+    const placeMerchOrderRequest = {
+      order,
+      pickupEvent: pickupEvent.uuid,
+    };
+    const merchController = ControllerFactory.merchStore(conn, instance(emailService));
+    const placedOrderResponse = await merchController.placeMerchOrder(placeMerchOrderRequest, member);
+
+    // fulfill a single item in an order
+    const placedOrder = placedOrderResponse.order;
+    const uuidParams = { uuid: placedOrder.uuid };
```
sumeet-bansal, line 216: Nit: `uuidParams` is not really a helpful variable name. Might as well just do `params`, or maybe `orderParams`.

tests/merch.order.test.ts
```diff
@@ -0,0 +1,721 @@
...
+        },
+      ],
+    };
+    await merchController.fulfillMerchOrderItems(uuidParams, fulfillMerchOrderItemsRequest, admin);
+
+    const getOrderResponse = await merchController.getOneMerchOrder(uuidParams, member);
+    const partiallyFulfilledOrder = getOrderResponse.order;
+
+    // order status should not be FULFILLED because entire order isn't fulfilled
+    expect(partiallyFulfilledOrder.status).toEqual(OrderStatus.PLACED);
```
sumeet-bansal, line 235: PLACED doesn't seem like the correct state for this. Maybe PARTIALLY_FULFILLED? Like, what if an order is partially fulfilled, and then the user tries to get a full refund, and we say "ok your order's PLACED but not FULFILLED so sure here's your full refund even though you already got some item(s)." Let's add a test for getting refunds on a partially fulfilled order.

tests/merch.order.test.ts
```diff
@@ -0,0 +1,721 @@
...
+    expect(partiallyFulfilledOrder.items.some((item) => item.fulfilled)).toBeTruthy();
+    expect(partiallyFulfilledOrder.items.every((item) => item.fulfilled)).toBeFalsy();
+  });
+
+  test('admins can fulfill entire orders if every item of a member\'s order is fulfilled', async () => {
```
sumeet-bansal, line 240: Is the order correctly marked as fulfilled if the items are fulfilled in separate requests? Can we update this test to do that? I can't tell what'd happen just by looking at the service method (idk what happens in the loop where we fulfill order items when there's no update for an item since the map would presumably return null).

tests/merch.order.test.ts
```diff
@@ -0,0 +1,721 @@
...
+    const placedOrderResponse = await merchController.placeMerchOrder(placeMerchOrderRequest, member);
+
+    // attempt to fulfill the member's own order
+    const placedOrder = placedOrderResponse.order;
+    const uuidParams = { uuid: placedOrder.uuid };
+    const fulfillMerchOrderItemsRequest = {
+      items: [
+        {
+          uuid: placedOrder.items[0].uuid,
+          fulfilled: true,
```
sumeet-bansal, line 351: Do we really need this `fulfilled: true` in the API request? Since we never un-fulfill, I don't think so, right?

sumeet-bansal (commented):
Mostly LGTM.

Nit: I don't love the `merch.order.test.ts` naming convention. Maybe something like `merchOrder.test.ts` instead? Or `MerchOrders.test.ts`? Or `MerchOrderTests.ts` even?

Also I might be wrong on this but I think it's possible to get full refunds on partially fulfilled orders, which is not ideal.

api/controllers/MerchStoreController.ts
```diff
@@ -164,10 +170,11 @@ export class MerchStoreController {
   }
 
   @Get('/order')
```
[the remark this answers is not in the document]
shravanhariharan2: This would make more sense for this particular route, but wouldn't we want all routes under the same resource to have the same prefix? i.e. any operation on the OrderModel resource is currently under `/store/order`, but with this change we'd have `/store/orders` and /store/order`. 

With events we have `/event/past` and `/event/future` as singular even though they return many events, so do we want to keep the same convention here? I don't have a strong preference here

tests/controllers/ControllerFactory.ts
```diff
@@ -17,90 +17,49 @@ import EventService from '../../services/EventService';
...
+    return new LeaderboardController(userAccountService);
   }
 
-  public static merchStore(conn: Connection): MerchStoreController {
-    if (!ControllerFactory.merchStoreController) {
-      const merchStoreService = new MerchStoreService(conn.manager);
-      ControllerFactory.merchStoreController = new MerchStoreController(merchStoreService);
-    }
-    return ControllerFactory.merchStoreController;
+  public static merchStore(conn: Connection, emailService = new EmailService()): MerchStoreController {
```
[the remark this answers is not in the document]
shravanhariharan2: Wouldn't always passing a mock instance clutter up the test files a bit, since not every test would require `EmailService` usage? One alternative is that we make the param optional, and if its undefined, we create a mock instance in the factory.

tests/merch.order.test.ts
shravanhariharan2: Yeah I think `fulfillOrderItems` is a bit confusing in the way it's setup. Right now, we're requiring every `OrderItem` in an order to be passed to the controller, where the `fulfilled` boolean is either true or false, and if all item's `fulfilled = true` at the end, then the order is marked as fulfillled. I think this makes sense design wise since the UI would have a list of checkboxes, and the entire list can just be sent to the backend after the admin checks/unchecks fulfilled items. 

image

`fulfillOrderItems` has some extra complexity that was based on our old assumption that `OrderItem`s for different orders can be sent in the same request, and that not all `OrderItem`s are sent at the same time, so I'll update that method accordingly

tests/merch.order.test.ts
shravanhariharan2: I don't think we'd actually need `PARTIALLY_FULFILLED` since we can just refund all the items of an order where `fulfilled = false`, right? Since `PARTIALLY_FULFILLED` is just a state for if any item has been fulfilled already.

tests/merch.order.test.ts
shravanhariharan2: Update on the above - we actually are rolling with 'you can't unfulfill items' cause the above design would only display unfulfilled items. So `fulfillOrderItems` will take in only the items that it wants to fulfill, and users can fulfill their order across multiple requests. I'll add some more tests to test this.

tests/merch.order.test.ts
```diff
@@ -0,0 +1,721 @@
...
+    // cancel order
+    const { uuid } = placedOrderResponse.order;
+    await merchController.cancelMerchOrder({ uuid }, member);
+
+    // get order, making sure state was updated and user has been refunded
+    const cancelledOrderResponse = await merchController.getOneMerchOrder({ uuid }, member);
+    const cancelledOrder = cancelledOrderResponse.order;
+
+    expect(cancelledOrder.status).toEqual(OrderStatus.CANCELLED);
+    expect(member.credits).toEqual(10000);
```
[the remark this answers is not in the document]
shravanhariharan2: When `member` is passed into the controller call, its credits manually get updated in the activity repository by a `user.credits += credits` call, so the object does get modified. In the cases where `member` is never passed into a controller (like in `cancelAllPendingOrders`), then we can manually reload with `await member.reload()` - I think I'll  add this call in every test though so that it's clearer

shravanhariharan2:
Just added the 'pending order cleanup' route for admins to cancel all orders.

TODOs (for myself):
- Add activity logs for fulfilling, cancelling, marking as missed, and cancel-all-ing
- Add GET /order/pickup/past route (since admins would want to scroll through past pickup events to see if orders have been missed)
- (TBD) Add functionality for rescheduling a partially picked up order? e.g. a member can only fulfill a subset of items in 1 event so they should be able to reschedule their pickup to a future one.

shravanhariharan2:
Will create issues for each of the above, since this PR is getting a bit large. Still need to do some refactoring for code quality, then should be ready for review

tests/merch.order.test.ts
```diff
@@ -0,0 +1,721 @@
...
+      .createMerchItemOption(affordableOption)
+      .createOrderPickupEvents(pickupEvent)
+      .write();
+
+    const emailService = mock(EmailService);
+    when(emailService.sendOrderConfirmation(member.email, member.firstName, anything()))
+      .thenResolve();
+
+    // place order
+    const order = [
```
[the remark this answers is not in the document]
shravanhariharan2: We could, but in different tests we're using different quantities of different options or placing differently-sized orders, so it seems a bit inconsistent to have `order()` in some places but not in others, which is why I decided to leave the request building in each test. The comments above each block should help make the tests more digestible

api/validators/MerchStoreRequests.ts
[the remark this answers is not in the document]
shravanhariharan2: We will leave it in for now and wait for Steven's green light if we need to keep this route in or not since it's not 100% solidified yet based on admin designs. I've just implemented the `/order/cleanup` route, and we can easily delete this route later if we decide we don't need it

shravanhariharan2:
Actually got around to implementing activity logging + the route in this PR 😅. Should be the last large PR for store in the next few weeks

api/validators/MerchStoreRequests.ts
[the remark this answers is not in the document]
sumeet-bansal: Can we move the cleanup route to a separate PR?

api/controllers/MerchStoreController.ts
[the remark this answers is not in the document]
sumeet-bansal: I like `/orders` plural because it's obvious the route is for getting several orders, but your call.

tests/controllers/ControllerFactory.ts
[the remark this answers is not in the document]
sumeet-bansal: I think it's fine, let's keep the interface simple.

tests/merch.order.test.ts
sumeet-bansal: Yeah, but it's not exactly accurate and might be a foot gun if someone down the road thinks `PLACED` means `NONE_FULFILLED`. I think it's an important distinction between none, any, and all fulfilled.

api/controllers/MerchStoreController.ts
```diff
@@ -164,10 +170,11 @@ export class MerchStoreController {
...
-  async getAllMerchOrders(@AuthenticatedUser() user: UserModel): Promise<GetAllMerchOrdersResponse> {
+  async getAllMerchOrders(@QueryParams() filters: GetOrderQueryParams,
+    @AuthenticatedUser() user: UserModel): Promise<GetAllMerchOrdersResponse> {
     if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
     const canSeeAllOrders = PermissionsService.canSeeAllMerchOrders(user);
-    const orders = await this.merchStoreService.getAllOrders(user, canSeeAllOrders);
+    const orders = await this.merchStoreService.getAllOrders(user, filters.status, canSeeAllOrders);
```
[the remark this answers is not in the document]
sumeet-bansal: ^

api/controllers/MerchStoreController.ts
```diff
@@ -197,18 +204,57 @@ export class MerchStoreController {
...
   }
 
-  @Patch('/order')
-  async editMerchOrder(@Body() fulfillOrderRequest: FulfillMerchOrderRequest,
+  @Patch('/order/:uuid/pickup')
+  async editMerchOrderPickup(@Params() params: UuidParam,
+    @Body() editOrderRequest: EditMerchOrderPickupRequest,
     @AuthenticatedUser() user: UserModel): Promise<EditMerchOrderResponse> {
-    if (!PermissionsService.canFulfillMerchOrders(user)) throw new ForbiddenError();
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
```
sumeet-bansal, line 211:
```suggestion
    if (!PermissionsService.canManageMerchOrders(user)) throw new ForbiddenError();
```

api/controllers/MerchStoreController.ts
```diff
@@ -197,18 +204,57 @@ export class MerchStoreController {
...
+    await this.merchStoreService.editMerchOrderPickup(params.uuid, editOrderRequest.pickupEvent, user);
+    return { error: null };
+  }
+
+  @Post('/order/:uuid/cancel')
+  async cancelMerchOrder(@Params() params: UuidParam, @AuthenticatedUser() user: UserModel) {
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
```
sumeet-bansal, line 218:
```suggestion
    if (!PermissionsService.canManageMerchOrders(user)) throw new ForbiddenError();
```

api/controllers/MerchStoreController.ts
```diff
@@ -197,18 +204,57 @@ export class MerchStoreController {
...
     if (fulfillOrderRequest.items.length !== numUniqueUuids) {
       throw new BadRequestError('There are duplicate order items');
     }
-    await this.merchStoreService.updateOrderItems(fulfillOrderRequest.items);
+    await this.merchStoreService.fulfillOrderItems(fulfillOrderRequest.items, params.uuid, user);
     return { error: null };
   }
 
+  @Post('/order/cleanup')
+  async cancelAllPendingMerchOrders(@AuthenticatedUser() user: UserModel): Promise<CancelAllPendingOrdersResponse> {
```
sumeet-bansal, line 243: If this route isn't needed yet, can we make this PR a little smaller and put up another PR when it is needed?

services/MerchStoreService.ts
```diff
@@ -264,27 +262,20 @@ export default class MerchStoreService {
     return order.getPublicOrder();
   }
 
-  public async getAllOrders(user: UserModel, canSeeAllOrders = false): Promise<PublicOrder[]> {
+  public async getAllOrders(user: UserModel, status?: OrderStatus, canSeeAllOrders = false): Promise<PublicOrder[]> {
```
sumeet-bansal, line 265: Let's split this up into separate `getAllOrdersForUser` and `getAllOrdersForAllUsers` methods and then call the appropriate one from the controller (and do `getPublicOrder` in the controller).

services/MerchStoreService.ts
```diff
@@ -296,25 +287,23 @@ export default class MerchStoreService {
     const [order, merchItemOptions] = await this.transactions.readWrite(async (txn) => {
       const merchItemOptionRepository = Repositories.merchStoreItemOption(txn);
       const itemOptions = await merchItemOptionRepository.batchFindByUuid(originalOrder.map((oi) => oi.option));
-
-      await this.verifyOrderWithTransaction(originalOrder, user, txn);
-
-      const totalCost = MerchStoreService.totalCost(originalOrder, itemOptions);
-
-      const merchOrderRepository = Repositories.merchOrder(txn);
+      await this.validateOrderUnderTxn(originalOrder, user, txn);
```
sumeet-bansal, line 290:
```suggestion
      await this.validateOrderInTransaction(originalOrder, user, txn);
```

services/MerchStoreService.ts
```diff
@@ -359,19 +348,40 @@ export default class MerchStoreService {
...
+        end: MerchStoreService.dateToHumanReadableDateString(order.pickupEvent.end),
+      },
     };
     this.emailService.sendOrderConfirmation(user.email, user.firstName, orderConfirmation);
 
     return order.getPublicOrder();
   }
 
-  public async verifyOrder(originalOrder: MerchItemOptionAndQuantity[], user: UserModel): Promise<void> {
+  private static dateToHumanReadableDateString(date: Date): string {
```
sumeet-bansal, line 362:
```suggestion
  private static humanReadableDateString(date: Date): string {
```

services/MerchStoreService.ts
```diff
@@ -359,19 +348,40 @@ export default class MerchStoreService {
...
+   * Validates a merch order. An order is considered valid if all the below are true:
+   *  - all the ordered item options exist within the database
+   *  - the ordered item options were placed for non-hidden items
+   *  - the user wouldn't reach monthly or lifetime limits for any item if this order is placed
+   *  - the requested item options are in stock
+   *  - the user has enough credits to place the order
+   * @param originalOrder
+   * @param pickupEventUuid
+   * @param user
+   * @param txn
```
sumeet-bansal, line 382: These annotations aren't adding anything useful.
```suggestion
   * @param originalOrder
   * @param pickupEventUuid
   * @param user
   * @param txn
```

services/MerchStoreService.ts
```diff
@@ -359,19 +348,40 @@ export default class MerchStoreService {
...
+   */
+  private async validateOrderUnderTxn(originalOrder: MerchItemOptionAndQuantity[],
```
sumeet-bansal, line 384:
```suggestion
  private async validateOrderInTransaction(originalOrder: MerchItemOptionAndQuantity[],
```

services/MerchStoreService.ts
```diff
@@ -431,34 +441,236 @@ export default class MerchStoreService {
...
+    return this.transactions.readWrite(async (txn) => {
+      const orderRespository = Repositories.merchOrder(txn);
+      const order = await orderRespository.findByUuid(uuid);
+      if (!order) throw new NotFoundError('Order not found');
+      if (order.status !== OrderStatus.PLACED) {
+        throw new UserError('Cannot mark an order as missed if it\'s already been cancelled, missed, or fulfilled');
+      }
+      // compare with start date and not end date so that store admins can mark orders
+      // as missed during the event and not necessarily limited to after the event
+      // (e.g. in the case where the event ends early)
```
sumeet-bansal, line 487: was a little confused by the wording, I think this is saying the same thing?
```suggestion
      // as missed during and after the event
```

services/MerchStoreService.ts
```diff
@@ -431,34 +441,236 @@ export default class MerchStoreService {
...
+    return order.status === OrderStatus.FULFILLED || order.status === OrderStatus.CANCELLED;
+  }
+
+  /**
+   * Marks an order as missed. An order can be marked as missed only if it's previous status was PLACED,
+   * and its associated pickup event has already passed. Only admins can mark orders as missed
+   * @param uuid order uuid
+   * @returns updated order
+   */
+  public async markOrderAsMissed(uuid: Uuid, proxy: UserModel): Promise<OrderModel> {
```
sumeet-bansal, line 477: I think we talked a while back about not using `proxy` but maybe `admin` isn't the best alternative? It's better than `proxy` at least.
```suggestion
  public async markOrderAsMissed(uuid: Uuid, admin: UserModel): Promise<OrderModel> {
```

services/MerchStoreService.ts
```diff
@@ -431,34 +441,236 @@ export default class MerchStoreService {
...
+
+  private static isInactiveOrder(order: OrderModel): boolean {
+    return order.status === OrderStatus.FULFILLED || order.status === OrderStatus.CANCELLED;
+  }
+
+  /**
+   * Marks an order as missed. An order can be marked as missed only if it's previous status was PLACED,
+   * and its associated pickup event has already passed. Only admins can mark orders as missed
+   * @param uuid order uuid
+   * @returns updated order
```
sumeet-bansal, line 475: We can remove these annotations and rename the param `orderUuid`.

services/MerchStoreService.ts
```diff
@@ -431,34 +441,236 @@ export default class MerchStoreService {
...
+      const order = await orderRespository.findByUuid(uuid);
+      if (!order) throw new NotFoundError('Order not found');
+      if (!user.isAdmin() && order.user.uuid !== user.uuid) {
+        throw new ForbiddenError('Members cannot cancel other members\' orders');
+      }
+      if (MerchStoreService.isInactiveOrder(order)) throw new UserError('Cannot cancel an inactive order');
+      if (MerchStoreService.isLessThanTwoDaysBeforePickupEvent(order.pickupEvent)) {
+        throw new NotFoundError('Cannot cancel an order with a pickup date less than 2 days away');
+      }
+      const customer = order.user;
```
sumeet-bansal, line 524: Is this used anywhere?

services/MerchStoreService.ts
```diff
@@ -431,34 +441,236 @@ export default class MerchStoreService {
...
+  private async refundAndConfirmOrderCancellation(order: OrderModel, user: UserModel, txn: EntityManager) {
+    const orderRepository = Repositories.merchOrder(txn);
+    const upsertedOrder = await orderRepository.upsertMerchOrder(order, { status: OrderStatus.CANCELLED });
+
+    // refund only the items that haven't been fulfilled yet
+    const unfulfilledItems = order.items.filter((item) => !item.fulfilled);
+    const refundValue = unfulfilledItems.reduce((refund, item) => refund + item.salePriceAtPurchase, 0);
+    await MerchStoreService.refundUser(user, refundValue, txn);
+
+    // build email with only the unfulfilled items
```
sumeet-bansal, line 544: Nit: I think the code explains itself here.

services/MerchStoreService.ts
```diff
@@ -498,6 +710,25 @@ export default class MerchStoreService {
     }, 0);
   }
 
+  public async cancelAllPendingOrders(user: UserModel): Promise<void> {
```
sumeet-bansal, line 713: Should this cancel all pending orders or only the ones that were placed a long time ago?

sumeet-bansal (approved):
Didn't look at the tests, maybe Steven or someone else can take a look before merging?

services/MerchStoreService.ts
shravanhariharan2: This would cancel all pending orders, since there could be the case where someone orders for the last pickup event of a quarter and wasn't able to make it or wasn't able to fulfill all of it

services/MerchStoreService.ts
shravanhariharan2: Yeah, it's used in the activity logs to describe who's order has been cancelled (since admins can cancel other members orders in the case of 'cancel all pending orders')

services/MerchStoreService.ts
shravanhariharan2: I was hesitant to use `admin` since soon we'll have `STORE_MANAGER` and `STORE_DISTRIBUTOR` roles who aren't necessarily admins but can mark orders as missed. We can probably leave it as admin for now but change it later to something like `manager`? But yeah, something that conveys 'on behalf of' would be ideal.

api/controllers/MerchStoreController.ts
shravanhariharan2: Just confirmed with Steven, we will be rolling with /order/cleanup route and removing the status query param - frontend can do any filtering if needed since there's no current use case for it

api/controllers/MerchStoreController.ts
shravanhariharan2: Members can cancel their own orders, so they don't need 'manage merch order' permissions since that's something a store manager / store distributor would need

api/controllers/MerchStoreController.ts
shravanhariharan2: Members can reschedule their pickup events, so they don't need store manager / store distributor permissions for this route

shravanhariharan2:
This should be ready to merge now, addressed all of Sumeet's comments and ran through the tests with Steven to see what else needs to be added to get 100% store coverage

Issue #206, opened 2021-09-29
Add get cart route


steets250:
The frontend currently stores the user's cart as a mapping of item option UUIDs to quantities. In order to actually display the cart, the frontend then needs the data about the item options. We would like a `GET /cart` route where we can send an array of item option UUIDs, and receive the item option objects in return.

PR #209, opened 2021-10-06
Implement checkout cart route

michl1001:
Addresses #206.


services/MerchStoreService.ts
```diff
@@ -540,4 +540,11 @@ export default class MerchStoreService {
       await orderPickupEventRepository.deletePickupEvent(pickupEvent);
     });
   }
+
+  public async getCartItems(items: string[]): Promise<Map<string, MerchandiseItemOptionModel>> {
+    return this.transactions.readWrite(async (txn) => {
+      const merchItemRepository = Repositories.merchStoreItemOption(txn);
+      return merchItemRepository.batchFindByUuid(items);
```
sumeet-bansal, line 547: 1. How should we handle items in archived collections or hidden items?
2. How should we handle missing items? Throw or only return found items?

services/MerchStoreService.ts
```diff
@@ -540,4 +540,11 @@ export default class MerchStoreService {
       await orderPickupEventRepository.deletePickupEvent(pickupEvent);
     });
   }
+
+  public async getCartItems(items: string[]): Promise<Map<string, MerchandiseItemOptionModel>> {
+    return this.transactions.readWrite(async (txn) => {
```
sumeet-bansal, line 545:
```suggestion
    return this.transactions.readOnly(async (txn) => {
```

api/controllers/MerchStoreController.ts
```diff
@@ -235,4 +237,14 @@ export class MerchStoreController {
...
   }
+
+  @Get('/cart')
+  async getCart(@Body() getCartRequest:GetCartRequest, @AuthenticatedUser() user: UserModel): Promise<GetCartResponse> {
+    if (!PermissionsService.canAccessMerchStore(user)) {
+      throw new ForbiddenError();
+    }
+    const items = await this.merchStoreService.getCartItems(getCartRequest.items);
+
+    return { error: null, items };
```
shravanhariharan2, line 248:
```suggestion
  async getCart(@Body() getCartRequest:GetCartRequest, @AuthenticatedUser() user: UserModel): Promise<GetCartResponse> {
    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
    const items = await this.merchStoreService.getCartItems(getCartRequest.items);
    return { error: null, items };
```

api/validators/MerchStoreRequests.ts
```diff
@@ -308,3 +309,9 @@ export class EditOrderPickupEventRequest implements IEditOrderPickupEventRequest
   @IsDefined()
   pickupEvent: OrderPickupEventEdit;
 }
+
+export class GetCartRequest implements IGetCartRequest {
+  @IsDefined()
+  @IsNotEmpty()
+  items:string[];
+}
```
shravanhariharan2, line 317:
```suggestion
  @IsNotEmpty()
  items: string[];
}
```

services/MerchStoreService.ts
shravanhariharan2: I don't think we need to be returning items that members don't have permissions for, like Sumeet said. If the frontend wants to let the user know that some of their cart items have been archived/hidden, then frontend can diff the uuids sent and the uuids received and display something like "Some of the items in your cart have recently been archived."

types/ApiRequests.ts
```diff
@@ -263,3 +263,7 @@ export interface CreateOrderPickupEventRequest {
 export interface EditOrderPickupEventRequest {
   pickupEvent: OrderPickupEventEdit;
 }
+
+export interface GetCartRequest{
+  items:string[];
+}
```
shravanhariharan2, line 269:
```suggestion
export interface GetCartRequest {
  items: string[];
}
```

shravanhariharan2 (commented):
Functionality LGTM, small nitpick but make sure the whitespace of the variables/interfaces/etc. that you're declaring is consistent with what's already in the codebase (referring to the suggestions I left below)

services/MerchStoreService.ts
sumeet-bansal: > If the frontend wants to let the user know that some of their cart items have been archived/hidden, then frontend can diff the uuids sent and the uuids received and display something like "Some of the items in your cart have recently been archived."

We can include this diff in `GetCartResponse`.

api/validators/MerchStoreRequests.ts
sumeet-bansal: @shravanhariharan2 can we update the linter to include this?

services/MerchStoreService.ts
```diff
@@ -540,4 +540,14 @@ export default class MerchStoreService {
...
   }
+
+  public async getCartItems(items: string[]): Promise<Map<string, PublicMerchItemOption>> {
+    return this.transactions.readOnly(async (txn) => {
+      const merchItemRepository = Repositories.merchStoreItemOption(txn);
+      const itemModels = await merchItemRepository.batchFindByUuid(items);
+      const publicItems = new Map<string, PublicMerchItemOption>();
+      itemModels.forEach(((v, k) => publicItems.set(k, v.getPublicMerchItemOption(false))));
+      return publicItems;
+    });
```
shravanhariharan2, line 551: The naming of the repositories is misleading here - we are using the `MerchItemOption` repository but are naming it as `merchItem` or `publicItems`. Remember, what the frontend wants is a list of options (to display the option chosen) and their item (to display the item name and image), so let's rename the variables accordingly to match that.

Also, the reason we are returning a `Map<string, PublicMerchItemOption>` is so that we are returning the data in the same order as the frontend sent it, so that the cart is displayed in the same order each time. If we're building our map based on `itemModels`, then the ordering of the map will be different than the ordering of uuids passed in `items`. So what we'd need to do is do `items.forEach((v, k)...)` instead of `itemModels.forEach((v, k)...)`, since then the map will be in the same order as the input. 

Additionally, we can lower the complexity of building the map by using `Array.prototype.map()` along with the `Map()` constructor:
```ts
return new Map(itemsOptionModels.map((option) => [option.uuid, option.getPublicMerchItemOption(false)]);
```
Let me know if anything's confusing with the above

shravanhariharan2 (changes requested):

api/controllers/MerchStoreController.ts
```diff
@@ -235,4 +238,16 @@ export class MerchStoreController {
       editOrderPickupEventRequest.pickupEvent);
     return { error: null, pickupEvent: pickupEvent.getPublicOrderPickupEvent() };
   }
+
+  async getCart(@Body() getCartRequest:GetCartRequest, @AuthenticatedUser() user: UserModel): Promise<GetCartResponse> {
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
+    const items = await this.merchStoreService.getCartItems(getCartRequest.items);
```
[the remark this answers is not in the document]
michl1001: OOPS, route must have gotten deleted while I was editing earlier, my bad

services/MerchStoreService.ts
michl1001: Sounds good, just pushed my implementation (itemOptionModels is actually a map but I think the current implementation should work).

api/controllers/MerchStoreController.ts
```diff
@@ -235,4 +238,17 @@ export class MerchStoreController {
       editOrderPickupEventRequest.pickupEvent);
     return { error: null, pickupEvent: pickupEvent.getPublicOrderPickupEvent() };
   }
+
+  @Get('/store/cart')
+  async getCart(@Body() getCartRequest:GetCartRequest, @AuthenticatedUser() user: UserModel): Promise<GetCartResponse> {
```
steets250, line 243: Since this route is a `GET` request, I don't believe we're able to send a body along with the request. Can we either change it so that the UUIDs can be passed as a query parameter, or switch it to a `POST` request? Thanks!

michl1001: Don't think the request really works as a POST request, I think query parameters are likely more accurate - can make the change now

steets250:
@shravanhariharan2 can you look into switching from a map to just returning an array that matches the original order of the item option uuids given by frontend?

api/controllers/MerchStoreController.ts
```diff
@@ -235,4 +239,18 @@ export class MerchStoreController {
...
+  async getCart(@QueryParams() getCartRequest:GetCartRequest,
+    @AuthenticatedUser() user: UserModel): Promise<GetCartResponse> {
+    if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
+    const items = await this.merchStoreService.getCartItems(getCartRequest.items);
+    const foundItems = Array.from(items.values())
+      .map((o) => o.uuid);
+    if (foundItems !== getCartRequest.items) {
+      const missingItems = difference(getCartRequest.items, foundItems);
+      throw new NotFoundError(`The following items were not found: ${missingItems}`);
+    }
```
shravanhariharan2, line 253: Instead of returning the `missingItems` list within the string of the error message, we should instead return it in some kind of JSON so that the frontend can easily get the list of missing items without doing any string manipulation. But it seems like we can't pass in the JSON data separately as a constructor to `NotFoundError`. @sumeet-bansal what you recommend here? Since frontend would need to use the list of missing items to clear the missing items from its own local storage, it'd need to parse the list somehow

services/MerchStoreService.ts
```diff
@@ -540,4 +540,12 @@ export default class MerchStoreService {
       await orderPickupEventRepository.deletePickupEvent(pickupEvent);
     });
   }
+
+  public async getCartItems(items: string[]): Promise<Map<string, PublicMerchItemOption>> {
+    return this.transactions.readOnly(async (txn) => {
+      const merchItemOptionRepository = Repositories.merchStoreItemOption(txn);
+      const itemOptionModels = await merchItemOptionRepository.batchFindByUuid(items);
+      return new Map(items.map((option) => [option, itemOptionModels.get(option).getPublicMerchItemOption(false)]));
+    });
```
shravanhariharan2, line 549: Like we discussed in our earlier meeting, let's return an array instead of a map that maintains the same order as the items provided - that way the response is cleaner and the frontend can just take our response and display it without having to index through the map to get options.

shravanhariharan2:
Route and tests are up, currently running into a limitation with `routing-controllers` where if query params are passed in as an array of size 1, then `routing-controllers` doesn't know if its a string or string[], so an error is thrown. This is the exact error I'm running into: https://github.com/typestack/routing-controllers/issues/123

Need to see if there's any way we can deserialize the input manually, otherwise we could either just use a request body with GET, which is technically allowed but is not good practice (link: https://stackoverflow.com/questions/978061/http-get-with-request-body), or we could change the request type to POST or PATCH or something (but GET makes the most sense by far)

shravanhariharan2:
Decided to roll with the GET /cart with request body since POST /cart doesn't make much sense for this context. We can look into storing carts on the backend if needed rather than in the frontend if this solution is too unclean, but for MVP store release I'm thinking this should be alright.

This PR should be ready for review now

shravanhariharan2 (approved):
image

steets250:
143503360-aeab1301-58ba-4a4c-a109-ec0ddf8ae2d8

PR #210, opened 2021-10-09
Add tests for attendance endpoints

sumeet-bansal:


models/AttendanceModel.ts
```diff
@@ -40,8 +40,8 @@ export class AttendanceModel extends BaseEntity {
 
   public getPublicAttendance(): PublicAttendance {
     return {
-      user: this.user?.getPublicProfile(),
-      event: this.event?.getPublicEvent(),
+      user: this.user.getPublicProfile(),
```
sumeet-bansal, line 43: `PublicAttendance` requires `user` and `event`. Apparently the endpoints for getting attendances for event/current user are broken because `getPublicAttendance` threw.

sumeet-bansal: Love that we're catching silly things like this now. 🚀

tests/attendance.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      .attendEvents([staff, member], [event], true)
+      .write();
+
+    const attendanceController = ControllerFactory.attendance(conn);
+    const params = { uuid: event.uuid };
+
+    // throws permissions error for member
+    await expect(attendanceController.getAttendancesForEvent(params, member))
+      .rejects.toThrow();
+
```
shravanhariharan2, line 39: nit: we should probably check that specifically a `ForbiddenError` was thrown vs. any error, i.e.
```ts
await expect(attendanceController.getAttendancesForEvent(params, member))
      .rejects.toThrow(ForbiddenError);
```

sumeet-bansal: Ah yeah definitely, updated.

PR #214, opened 2021-11-02
Add Order Limit to Pickup Events

### LINKED ISSUE #208: Add order limit to pickup events
Pickup events should include an `orderLimit: number` field to hold the maximum number of orders to be fulfilled at the event. This will help with scheduling, so that users know that they can't select specific pickup events because they have been fully booked already.

StormFireFox1:
Closes #208.

While this currently does not address the case of "1 person with multiple orders at same event", it could be implemented if required.


shravanhariharan2:
I'll get to reviewing this during our backend kickback 🎉 🥴

repositories/MerchOrderRepository.ts
```diff
@@ -69,7 +69,9 @@ export class OrderPickupEventRepository extends BaseRepository<OrderPickupEventM
   public async getFuturePickupEvents(): Promise<OrderPickupEventModel[]> {
     return this.getBaseFindQuery()
       .where('"end" >= :now')
+      .andWhere(':orderCount <= "orderLimit"')
       .setParameter('now', new Date())
+      .setParameter('orderCount', 0)
       .getMany();
   }
```
shravanhariharan2, line 76: Any reason we are setting 0 as the `orderCount` here? Future pickup events can have orders scheduled for them, so we'd need to filter by the actual order count to get only the future pickup events that have not reached their capacity yet in the "place order" page on the frontend. However, we need to also be able to get all future pickup events regardless of capacity in the admin page for admins to scroll through, so let's remove the filter here and just have the frontend perform the filtering on the 'place order' side to only display future events with capacity.

services/MerchStoreService.ts
```diff
@@ -313,6 +314,13 @@ export default class MerchStoreService {
         throw new NotFoundError('Cannot pickup order at an event that starts in less than 2 days');
       }
 
+      // Verify that this order would not set the pickup event's order count
+      // over the order limit
+      const currentOrderCount = pickupEvent.orders.length;
+      if (currentOrderCount >= pickupEvent.orderLimit) {
+        throw new UserError('Cannot place order with a fully-booked pickup event');
+      }
```
shravanhariharan2, line 322: nit: Since this is a member-facing error message, let's use something more friendly/less technical like 'This merch pickup event is full! Please choose a different pickup event'

tests/merch.test.ts
```diff
@@ -564,4 +569,81 @@ describe('merch order pickup events', () => {
     expect(persistedPickupEvent.orders).toHaveLength(1);
     expect(persistedPickupEvent.orders[0]).toStrictEqual(persistedOrder);
   });
+
+  test('placing an order with a pickup event that has reached the order limit fails', async () => {
+    const conn = await DatabaseConnection.get();
+    const member = UserFactory.fake({
+      points: 100,
+    });
```
shravanhariharan2, line 577: nit: It should be fine to make this one-line since it's only one property we're passing

tests/merch.test.ts
```diff
@@ -564,4 +569,81 @@ describe('merch order pickup events', () => {
...
+    const item = MerchFactory.fakeItem({
+      hidden: false,
+      monthlyLimit: 100,
+    });
```
shravanhariharan2, line 581: nit: I don't believe we need to create this merch item, since we aren't using it / asserting on it anywhere in this test. `PortalState.createMerchItemOption()` should take care of creating the item and collection for that option, so we can exclude this factory call.

tests/merch.test.ts
```diff
@@ -564,4 +569,81 @@ describe('merch order pickup events', () => {
     expect(persistedPickupEvent.orders).toHaveLength(1);
     expect(persistedPickupEvent.orders[0]).toStrictEqual(persistedOrder);
   });
+
+  test('placing an order with a pickup event that has reached the order limit fails', async () => {
+    const conn = await DatabaseConnection.get();
+    const member = UserFactory.fake({
```
shravanhariharan2, line 575: Let's add some happy-path tests as well, like 'placing an order with a pickup event that has not reached its capacity succeeds', or 'updating a pickup event's order limit to a threshold above the current order count succeeds'

tests/merch.test.ts
```diff
@@ -564,4 +569,81 @@ describe('merch order pickup events', () => {
...
+    const item = MerchFactory.fakeItem({ hidden: false });
+    const option = MerchFactory.fakeOption({
+      item,
+      price: 10,
+    });
+    const pickupEvent = MerchFactory.fakeOrderPickupEvent({
+      start: moment().add(3, 'days').toDate(),
+      end: moment().add(4, 'days').toDate(),
+      orderLimit: 2,
+    });
```
shravanhariharan2, line 628: nit: there should be a factory method called `fakeFuturePickupEvent()` in the `OrderStates` branch that you can add to the factory so that you don't need to specify durations, since all we care about is that we can place orders on it:
```ts
  public static fakeFutureOrderPickupEvent(substitute?: Partial<OrderPickupEventModel>): OrderPickupEventModel {
    // guarantee that start and end are 2 days after current time so that orders can be placed
    const daysAhead = FactoryUtils.getRandomNumber(3, 10);
    const duration = FactoryUtils.getRandomNumber(30, 180, 30);
    const start = moment().add(daysAhead, 'days').toDate();
    const end = moment(start).add(duration, 'minutes').toDate();
    return MerchFactory.fakeOrderPickupEvent({ start, end, ...substitute });
  }
```

tests/merch.test.ts
```diff
@@ -564,4 +569,81 @@ describe('merch order pickup events', () => {
...
+
+    await new PortalState()
+      .createUsers(admin, member)
+      .createMerchItem(item)
+      .createMerchItemOption(option)
+      .createOrderPickupEvents(pickupEvent)
+      .orderMerch(member, [{ option, quantity: 1 }], pickupEvent)
+      .orderMerch(member, [{ option, quantity: 1 }], pickupEvent)
+      .write();
```
shravanhariharan2, line 637: nit: `orderMerch()` takes in an array of options and quantities, so I think you instead also do:
```ts
.orderMerch(member, [{ option, quantity: 1 }, { option, quantity: 1 }], pickupEvent)
```
or just:
```ts
.orderMerch(member, [{ option, quantity: 2 }], pickupEvent)
```

shravanhariharan2:
> While this currently does not address the case of "1 person with multiple orders at same event", it could be implemented if required.

What do you mean by this? It looks like to me that this should be handled just fine with the current implementation, since multiple separate orders do count towards the limit

tests/merch.test.ts
shravanhariharan2: nvm, we'd want to call it multiple times so that they're separate orders. No change needed 😛

StormFireFox1:
> It looks like to me that this should be handled just fine with the current implementation, since multiple separate orders do count towards the limit

I think @sumeet-bansal mentioned it in #208, but essentially if one person has multiple orders assigned to one pickup event, then their orders could potentially be "bunched up" into one bigger pickup. That way, we wouldn't have the extreme case of one person taking up an entire event's order limit with their own stuff (it would be weird if one would be the only person at a pickup event).

tests/merch.test.ts
StormFireFox1: Resolved in 075195b.

services/MerchStoreService.ts
StormFireFox1: Resolved in 895837e.

repositories/MerchOrderRepository.ts
StormFireFox1: Resolved in d149edc.

tests/merch.test.ts
StormFireFox1: Resolved in 895837e.

tests/merch.test.ts
StormFireFox1: Resolved in 895837e.

StormFireFox1: Apparently, removing the `fakeItem` call makes `fakeOption` generate an item that can potentially be hidden, which in turn causes unexpected failures in testing.

For now, I'll reverse this change, but perhaps we can adjust `fakeOption` to not hide items by default.

shravanhariharan2:
@StormFireFox1 was this ready for re-review?

tests/merch.test.ts
StormFireFox1: Resolved in `6c62606` with some issues. See comments.

StormFireFox1:
> was this ready for re-review?

Yes, it is now.

Also, I noticed an issue with our implementation of OrderPickupEvent, though it's possible I'm misinterpreting.

api/controllers/MerchStoreController.ts
```diff
@@ -233,6 +233,7 @@ export class MerchStoreController {
     if (!PermissionsService.canManagePickupEvents(user)) throw new ForbiddenError();
     const pickupEvent = await this.merchStoreService.editPickupEvent(params.uuid,
       editOrderPickupEventRequest.pickupEvent);
-    return { error: null, pickupEvent: pickupEvent.getPublicOrderPickupEvent() };
+    return { error: null,
+      pickupEvent: pickupEvent.getPublicOrderPickupEvent() };
```
StormFireFox1, line 237: Noticed that `merchStoreService` does not return a full fetch of `OrderModel[]` for the `order` field, causing `TypeError` to occur whenever we try to render the orders associated with a pickup event (ie `getPublicOrderPickupEvent(true)`. Because of this, an admin cannot directly return the orders present in a pickup event upon edit without needing another query.

Do we need to modify `getPublicOrderPickupEvent` to fix this? Am I misusing the method?

StormFireFox1:
As another additional comment, do we mock `emailService` in tests or not?

api/controllers/MerchStoreController.ts
shravanhariharan2: > Noticed that merchStoreService does not return a full fetch of OrderModel[] for the order field
I'm not 100% clear on what you mean by this - it looks like `editPickupEvent` calls `orderPickupRepository.findByUuid()`, which does a bunch of joins to get the orders, the orders' items, and the orders' users. If you use `getPublicOrderPickupEvent(true)`, do you still get a `TypeError`?

shravanhariharan2:
> As another additional comment, do we mock emailService in tests or not?

Yup - you can have a look at `auth.test.ts` to see how we do that

api/controllers/MerchStoreController.ts
StormFireFox1: Never mind on this; I had some scratch code that was ruining things, apparently.

shravanhariharan2:
@StormFireFox1 this is ready for review besides merging the order states changes into this PR for email mocking, right?

StormFireFox1:
Yeah, will merge `master` in today and get this ready for review again.

StormFireFox1:
@shravanhariharan2 Locked and loaded for review!

tests/merchStore.test.ts
```diff
@@ -406,3 +409,274 @@ describe('merch item options', () => {
...
   });
 });
+
+describe('merch order pickup events', () => {
+  test('GET /order/pickup/future returns pickup events that haven\'t ended yet', async () => {
+    const conn = await DatabaseConnection.get();
+    const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
+    const pastPickupEvent = MerchFactory.fakeOrderPickupEvent({
+      start: moment().subtract(2, 'hour').toDate(),
+      end: moment().subtract(1, 'hour').toDate(),
```
shravanhariharan2, line 419: We already have a few pickup event tests defined in `merchOrder.test.ts` - some of which overlap with the ones written here. Can we move these tests to `merchOrder.test.ts` and merge the duplicated tests? `merchOrder.test.ts` is meant for testing constructs related to merch orders (i.e. orders, pickup events, fulfillment) whereas `merchStore.test.ts` is meant for testing the store structures like collections, items, and options. Apologies if that was unclear, I can add some comments within those test files to make the distinction clearer

shravanhariharan2 (approved):
The logic LGTM, just a small comment on the tests

tests/merchStore.test.ts
StormFireFox1: This PR was made before your refactor, hence the duplication. I’ll merge the tests.

tests/data/MerchFactory.ts
```diff
@@ -91,6 +91,7 @@ export class MerchFactory {
       description: faker.lorem.sentences(2),
       start,
       end,
+      orderLimit: faker.datatype.number({ min: 1, max: 5 }),
       orders: [],
```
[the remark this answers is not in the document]
StormFireFox1: Resolved in `f4dc6a0`.

tests/merchStore.test.ts
StormFireFox1: Resolved in `c38ac89`.

PR #215, opened 2021-11-02
Extend event check-in time to 30 minutes before and after an event

### LINKED ISSUE #213: Add padding window to event checkin time
It's been requested that we allow a member to check in to an event anytime from 30 minutes before to 30 minutes after the posted time of an event. Now that events are back in person, we'd like to start enforcing that people have checked in to an event before they are allowed to enter, but if we can't start doing that until the event start time, it will lead to delays. In addition, there are cases where members may forget to check in or may not take the time to register on the portal until after the event, and we want to avoid the issue of them not having the ability to check in if that is the case. The time padding doesn't need to vary on a per-event basis, so it would be best to implement as a solely backend feature.

shravanhariharan2:
Closes #213 .

### DIFF
--- a/services/AttendanceService.ts
+++ b/services/AttendanceService.ts
@@ -38,7 +38,7 @@ export default class AttendanceService {
     return this.transactions.readWrite(async (txn) => {
       const event = await Repositories.event(txn).findByAttendanceCode(attendanceCode);
       if (!event) throw new NotFoundError('Oh no! That code didn\'t work.');
-      if (!event.isOngoing()) throw new UserError('You can only enter the attendance code during the event!');
+      if (!event.isAcceptingAttendances()) throw new UserError('This event is not currently accepting attendances!');
 
       const hasAlreadyAttended = await Repositories.attendance(txn).hasUserAttendedEvent(user, event);
       if (hasAlreadyAttended) throw new UserError('You have already attended this event');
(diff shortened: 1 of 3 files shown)


services/AttendanceService.ts
```diff
@@ -38,7 +38,7 @@ export default class AttendanceService {
     return this.transactions.readWrite(async (txn) => {
       const event = await Repositories.event(txn).findByAttendanceCode(attendanceCode);
       if (!event) throw new NotFoundError('Oh no! That code didn\'t work.');
-      if (!event.isOngoing()) throw new UserError('You can only enter the attendance code during the event!');
+      if (!event.isAcceptingAttendances()) throw new UserError('This event is not currently accepting attendances!');
```
shravanhariharan2, line 42: Wasn't too sure how to phrase the error to explain that 'you can only check in to events at most 30 minutes before and at least 30 minutes after the event' without being too long, @sumeet-bansal do you think this is preferred to simply 'this event is not currently accepting attendances'?

steets250: @shravanhariharan2 Would we be able to make this message a bit more specific, based on the current time compared to the event window? For example "This event hasn't started yet, please wait to check in." or "This event has ended and is no longer accepting attendances"? I'm definitely open to alternate wording as well.

sumeet-bansal: +1 on ^

shravanhariharan2:
trying to fix linting errors at a party using github editor be like

trulyronak (approved):
so TRUE

shravanhariharan2:
@sumeet-bansal this is ready for review now

sumeet-bansal (approved):

PR #220, opened 2021-11-21
Add store roles and permissions

### LINKED ISSUE #216: Add hardcoded user roles
- The new roles will be: `STORE_MANAGER`, `MERCH_DISTRIBUTOR` and `MARKETER`.
- `STORE_MANAGER` would have permissions to modify the store (add items/options, archive collections, set discount percentages, etc.) + handle merch distribution (fulfill/cancel orders, mark orders as missed, add/remove pickup events, etc)
- `MERCH_DISTRIBUTOR` would have just handle merch distribution (fulfilling + cancelling orders)
- `MARKETER` would have just add/edit event permissions

shravanhariharan2:
Closes #216 

API routes for assigning users roles will be created with the roles redesign, since this PR is just a temporary glue for logging which board members are fulfilling which orders and which board members are adding/deleting items, etc. Also since we'd want users to have multiple roles, instead of having 1 role that has an implicit hierarchy defined in the code


steets250:
Can a user have multiple access types? Like could someone have both the marketing role *and* a store distributor role?

shravanhariharan2:
For the initial store release, I'd like to leave it as 1 role per user for simplicity, since having multiple roles means more changes like adding a 'Roles' table and more edge case checks like making sure users can't be both have `BLOCKED` and `STANDARD` roles, etc.

If we're going to add more logic to implement roles, then I think we might as well do so according to the role redesign spec (e.g. rewriting permissions to be role-based, being able to add/remove roles, allowing permissions for some roles to grant roles with less permissions to members, etc), which isn't yet confirmed. That's why I'm keeping it very simple here, since our main goal is to track who is fulfilling/cancelling orders in the case that a fulfillment/cancellation goes south.

models/UserModel.ts
```diff
@@ -87,6 +87,22 @@ export class UserModel extends BaseEntity {
     return this.accessType === UserAccessType.STAFF;
   }
 
+  public hasStoreDistributorPermissions(): boolean {
```
sumeet-bansal, line 90: This should be `PermissionsService::canDistributeMerch` or something.

sumeet-bansal (approved):

models/UserModel.ts
michl1001: ^^^^ this

michl1001 (approved):
Looks good - Sumeet's comment is the only real thing I can see

Issue #231, opened 2021-12-15
Implement POST /order/pickup/:uuid/complete route


shravanhariharan2:
This route should mark all pending orders for a pickup event as either MISSED or PARTIALLY_FULFILLED. This eliminates the need to mark individual orders as missed, since realistically a store distributor should only need to worry about fulfilling orders and not marking specific orders as missed (this should be handled by a system).

With this route, the POST /order/:uuid/missed route is no longer needed, since this route would guarantee to cover all orders since every order would be associated with a pickup event (thus no order could be 'dangling' and need to be marked as missed manually).

shravanhariharan2:
Resolved via. #238

PR #238, opened 2021-12-24
Implement merch store order fulfillment changes

### LINKED ISSUE #225: Send confirmation emails on order fulfillment
Currently, only orders that are fully fulfilled have confirmation emails sent to them for users - this capability should be expanded to any type of fulfillment so that a member has confirmation after they pick up any part of their order at a pickup event, not just only if they get their full order

Edit: Wait, not even fully fulfilled orders have emails sent to them 🤔 . Updated the title accordingly

shravanhariharan2:
Closes #225 and #231

Make changes to APIs and services to match the following state machine for orders: 
image

Changes:
* Added ability to partially fulfill orders if an entire order cannot be fulfilled during a single pickup event
* Added email notifications for partial fulfillment and regular fulfillment
* Added POST /order/pickup/:uuid/complete route to 'complete' a pickup event, marking all unfulfilled orders as missed and requesting a reschedule for each pickup event


shravanhariharan2:
@steets250 what should we do in the case where the last pickup event of the quarter is completed and some orders are marked as missed? Normally we'd prompt for reschedule (by the changes in this PR), but since there's no more events, should we instead automatically cancel all pending orders after the last pickup event? If we don't do this, then people might be prompted to reschedule, but they can't actually reschedule to any future event, and thus need to wait for an admin to "cancel all" or to just cancel themselves. 

If we instead automatically cancel after the last pickup event, would we need the "cancel all orders" route anymore? If that's the case too, how do we differentiate the instance where we're at the end of the quarter and legitimately need to cancel all orders vs. we don't have any future pickup events on the portal yet but there are still more planned (ie. its in the middle of the quarter).

Let me know if you need any clarification with anything above

steets250:
@shravanhariharan2 I think in this case it's better to rely on the admin clicking "cancel all" versus doing so automatically. For something like the end of winter quarter where there's only a one week gap before school starts again and more pickup events will be added, the admins may opt to not cancel anything, where as if it's the summer and everyone is going to be gone for a while, they they'd opt to cancel all. I'd rather give membership that choice than having the backend make any assumptions about when orders should be cancelled.

sumeet-bansal:
Why can an order go from PARTIALLY_FULFILLED to PLACED? Once an order's PARTIALLY_FULFILLED it should only go to FULFILLED or CANCELED right? The "partially fulfilled but canceled" order is a weird situation to be in though because I don't think CANCELED adequately describes the state of that order.

shravanhariharan2:
> Why can an order go from PARTIALLY_FULFILLED to PLACED?

The PARTIALLY_FULFILLED state is defined where an order has been only partially fulfilled during a pickup event, and the user needs to reschedule for another pickup event to complete it. So when an admin marks only a subset of items of an order as fulfilled, the user will get an email asking them to schedule a pickup event to get the rest of their order. When the user does that scheduling, then the order state returns to PLACED, since PLACED defines an order that is ready to pickup. We could have this state be called READY_FOR_PICKUP or something similar, but all orders that are 'placed' are 'ready for pickup' since they always have a pickup event attached to them - otherwise, they are 'missed' or 'partially fulfilled' (i.e. they need to be rescheduled).

> The "partially fulfilled but canceled" order is a weird situation to be in though because I don't think CANCELED adequately describes the state of that order.

The "partially fulfilled but canceled" state is set when an order only gets partially fulfilled during a pickup event, and either the user decides to cancel the rest of their order (e.g. they cant make any of the remaining pickup events), or when an admin cancels the rest of the order (e.g. there are no more remaining pickup events for the year). So I think in both cases, the order is technically 'cancelled', since the entire order has not been received by the user.

shravanhariharan2:
@steets250 sounds good - in that case, since we are always sending reschedule emails if an order remains partially fulfilled after a pickup event, should it be fine to send one even if there's no future pickup events? Or do we want a separate, custom email like "Your order is partially fulfilled, but there's no more pickup events currently planned. Please visit <some url> to see updates to pickup events"? 

In the first case, the user is prompted to reschedule, sees that there's no pickup events to choose from, and then thinks "what the fuck", and then quits ACM (more probable though, they get confused for a few hours maximum before their order gets cancelled by an admin - but they'd have 2 emails, one for "reschedule your order" and another for "the remaining of your order has been cancelled"). But in the second case, we might have it where there still are future order pickup events planned (e.g. we are in the middle of the quarter), but they just aren't in the portal yet. 

Which one should we go with? I'm thinking the first case is alright since that case should barely happen anyway, if at all, and the user would probably be told during the pickup event that "we are going to cancel the rest of your order so you get a refund", so the user would know why they got 2 emails. But I'd like to run it by you and see what you think

steets250:
@shravanhariharan2 yeah, I think the first option is fine. It should be a super rare case that we somehow don't have enough in stock, since membership is going to pre-prepare orders to distribute. And in the case that an item is missing, they'll make it clear to the person at the pickup that they'll cancel it later on since it's the end of the quarter. I don't think any person is going to reach the "wtf" state, they'd be like.. oh yeah Shirley said ACM ran out of items and will refund my credits for later.

services/MerchStoreService.ts
```diff
@@ -28,7 +28,7 @@ import { OrderModel } from '../models/OrderModel';
 import { UserModel } from '../models/UserModel';
 import Repositories, { TransactionsManager } from '../repositories';
 import { MerchandiseCollectionModel } from '../models/MerchandiseCollectionModel';
-import EmailService from './EmailService';
+import EmailService, { OrderInfo, OrderPickupEventInfo } from './EmailService';
```
michl1001, line 31: nit: Why import the types from here? Should the types not be defined in the types file?

shravanhariharan2:
> When the user does that scheduling, then the order state returns to PLACED, since PLACED defines an order that is ready to pickup. We could have this state be called READY_FOR_PICKUP or something similar, but all orders that are 'placed' are 'ready for pickup' since they always have a pickup event attached to them - otherwise, they are 'missed' or 'partially fulfilled' (i.e. they need to be rescheduled).

@sumeet-bansal @steets250 what if we renamed PLACED to something like REQUESTED? @michl1001 suggested this idea, the reason being you're requesting an order to be fulfilled, and semantically, even if its partially fulfilled, it can still go back to being requested, since the order hasn't been entirely fulfilled, which makes more sense than going back to PLACED since you aren't "re-placing" the order.

steets250:
That sounds good to me, I think the logic behind the naming makes sense there, but I have less stake in the state names since frontend interacts with them a lot less than backend, so I'll leave it up to you all.

api/controllers/MerchStoreController.ts
```diff
@@ -325,6 +319,14 @@ export class MerchStoreController {
     return { error: null };
   }
 
+  @Post('/order/pickup/:uuid/complete')
+  async completePickupEvent(@Params() params: UuidParam, @AuthenticatedUser() user: UserModel):
+  Promise<CompleteOrderPickupEventResponse> {
+    if (!PermissionsService.canManagePickupEvents(user)) throw new ForbiddenError();
+    const ordersMarkedAsMissed = await this.merchStoreService.completeOrderPickupEvent(params.uuid);
```
michl1001, line 326: For readability and style, would we want the controller to have the higher level idea completeing a pickup event while the service would have a more low level specific function (i.e. the service would just have a funciton "markMissedPickUps") or is that not necessary/more confusing

michl1001 (approved):
Left a few comments relating to style, but over all the functionality looks correct

api/controllers/MerchStoreController.ts
shravanhariharan2: Good question - with controllers, their main role is to take an incoming request, validate it, send it to the service for executing some business logic that generates a response, then return that response to the client. In the case you mentioned, 'completing a pickup event' constitutes business logic, so the service would handle that.  The repository layer is in charge of doing any database-level operations, and does not need to know any business logic (e.g. instead of it having a `markMissedPickUps` function, it can have a `updateOrderStatus(status: OrderStatus)` function). Hopefully the responsibilities of each layer is a bit clearer

services/MerchStoreService.ts
shravanhariharan2: We had those types there since they were specifically used for the `EmailService` methods, but now that they're being accessed by multiple service classes, we should probably move them to a types file.

api/controllers/MerchStoreController.ts
michl1001: Ahhhh, Okay I see my misconception - in that case, this makes more sense 👍

services/MerchStoreService.ts
shravanhariharan2: Actually, I think since those types are only meant to be associated with `EmailService`, it makes more sense to leave it, since otherwise any class can import those types from `../types` and think that it's okay to use, when instead importing from './EmailService` makes it clearer that its meant to be used with `EmailService`.

PR #241, opened 2021-12-26
Resolve countItemOrders Bug

### LINKED ISSUE #240: GET /merch/order/verification Throwing 500 Error
When attempting to verify a merch order on the `jpan@ucsd.edu` testing account, the following response is given:
```
{
    "error": {
        "name": "TypeError",
        "message": "Cannot destructure property 'item' of 'itemOptions.get(...)' as it is undefined.",
        "httpCode": 500,
        "stack": "TypeError: Cannot destructure property 'item' of 'itemOptions.get(...)' as it is undefined.\n    at Function.countItemOrders (/app/build/services/MerchStoreService.js:713:21)\n    at MerchStoreService.<anonymous> (/app/build/services/MerchStoreService.js:448:65)\n    at Generator.next (<anonymous>)\n    at fulfilled (/app/build/services/MerchStoreService.js:17:58)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:93:5)"
    }
}
```

steets250:
The `countItemOrders` function made the assumption that the call `itemOptions.get(orderedItemOptions[i].option.uuid)` would always succeed and could be unwrapped to access the response's item property. However, this call returns false when an item that has previously been ordered isn't present in the current order's list of items. I added a check to see if the previously ordered item is present in the order being verified. If so, the count should be incremented. If not, we don't care, since we're only checking monthly/lifetime limits on the pending order.

I additionally increased the `jpan@ucsd.edu` testing account starting points, so that it wouldn't have a negative credit balance.

Resolves #240.

### DIFF
--- a/services/MerchStoreService.ts
+++ b/services/MerchStoreService.ts
@@ -705,8 +705,11 @@ export default class MerchStoreService {
     }
     const orderedItemOptions = flatten(orders.map((o) => o.items));
     for (let i = 0; i < orderedItemOptions.length; i += 1) {
-      const { item } = itemOptions.get(orderedItemOptions[i].option.uuid);
-      if (counts.has(item)) counts.set(item, counts.get(item) + 1);
+      const orderedOption = itemOptions.get(orderedItemOptions[i].option.uuid);
+      if (orderedOption) {
+        const { item } = orderedOption;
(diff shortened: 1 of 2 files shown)


shravanhariharan2 (approved):
Nice catch x2 👍🏾

PR #242, opened 2021-12-26
Allows separate route for getting all orders

### LINKED ISSUE #232: Modifying GET /order Route
The description for the `GET /order` route is currently - for a user: get all orders (and their ordered items) for that user. for an admin, get all orders (and their ordered items) by all users.

Can we modify this route so that the users orders are returned, and an admin would specifically have to indicate that they want all orders to be returned, instead of having this as the default behavior (for example, /order?all=true or something like that)? In the event that a board member who manages store wants to view their orders, I wouldn't want every single user's orders to be returned.

michl1001:
- Closes #232.
- Separated /order route to /order and /order/all
- Generalize GetMerchOrderResponse to work for both routes


api/controllers/MerchStoreController.ts
```diff
@@ -190,16 +190,18 @@ export class MerchStoreController {
     return { error: null, order };
   }
 
+  @Get('/orders/all')
+  async getAllMerchOrders(@AuthenticatedUser() user: UserModel): Promise<GetMerchOrdersResponse> {
+    if (!(PermissionsService.canAccessMerchStore(user)
```
shravanhariharan2, line 195: nit: can we rename `getAllMerchOrders` -> `getMerchOrdersForAllUsers` and `getUserMerchOrders` to `getMerchOrdersForCurrentUser`? Just to make it clearer the difference (ie. we can't do "get all orders for any user, only the one who is making the call")

api/controllers/MerchStoreController.ts
```diff
@@ -190,16 +190,18 @@ export class MerchStoreController {
...
+      && PermissionsService.canSeeAllMerchOrders(user))) throw new ForbiddenError();
+    const orders: OrderModel[] = await this.merchStoreService.getAllOrdersForAllUsers();
+    return { error: null, orders: orders.map((o) => o.getPublicOrder()) };
```
shravanhariharan2, line 198: nit: we don't need the explicit typing of `const orders: OrderModel[]` here, since we know the return type of `getAllOrdersForAllUsers()` is `OrderModel[]`

shravanhariharan2 (commented):
Can we add a test in `merch-order.test.ts` for the GET /orders/all route and the GET /orders? Just something quick to make sure when multiple users order something, we are able to get all users' orders with the 'all' route, and that the GET /orders route only returns for the specific users (you can have these two assertions in the same test).

shravanhariharan2 (approved):

PR #243, opened 2021-12-26
Adds checks to email address for store access

### LINKED ISSUE #230: Require @ucsd.edu or @acmucsd.org emails for Merch Store Checkout
Only users with verified emails at the above domains should be able to place an order on the merch store.

michl1001:
- Closes #230.
- Adds extra checks to PermissionService.canAccessMerchStore


shravanhariharan2:
Noticing that the tests are failing cause we aren't using @ucsd.edu or @acmucsd.org emails in our user generation 😛. What we can do is specify a provider in the `faker.internet.email()` call in `UserFactory.ts` as `'ucsd.edu'` (and specify the first and last names as the same ones generated for firstName and lastName, you can use some variables to store those beforehand)

services/PermissionsService.ts
```diff
@@ -41,7 +41,8 @@ export default class PermissionsService {
   }
 
   public static canAccessMerchStore(user: UserModel): boolean {
-    return user.state === UserState.ACTIVE;
+    return user.state === UserState.ACTIVE && (user.email.split('@')[1] === 'ucsd.edu'
+      || user.email.split('@')[1] === 'acmucsd.org');
   }
```
shravanhariharan2, line 46: nit: let's move `user.email.split('@')[1]` into a variable like `emailDomain` to improve readability

tests/data/UserFactory.ts
```diff
@@ -27,11 +27,13 @@ export class UserFactory {
   }
 
   public static fake(substitute?: Partial<UserModel>): UserModel {
+    const fName = faker.name.firstName();
+    const lName = faker.name.lastName();
     const fake = UserModel.create({
```
shravanhariharan2, line 32: nit: let's avoid variable naming shorthand and use `firstName` and `lastName`, even more so since we can simplify lines 34 and 35 to:
```ts
{
...
firstName,
lastName,
hash: UserFactory.PASSWORD_HASH,
...
}
```

shravanhariharan2 (commented):
Few comments:
* Looks like the `.env.example` file was deleted, can you add that file back?
* Looks like the `package-lock.json` diff is huge - there might've been something that was overwritten during an `npm install` or by your local node. Can you reset that file to only include the version bump that happened in #244?
* Can you add a basic merch store test to make sure someone without an @ucsd.edu or @acmucsd.edu email cannot access the merch store? (i.e. something like checking if GET /merch/collections fails for that user)

shravanhariharan2 (approved):

tests/merchOrder.test.ts
```diff
@@ -84,6 +84,65 @@ describe('merch orders', () => {
...
   });
 
+  test('members can only access store with a valid acm or ucsd email', async () => {
+    const conn = await DatabaseConnection.get();
+    const UCSDMember = UserFactory.fake({ credits: 10000 });
+    const ACMMember = UserFactory.fake({
+      email: 'gZhao@acmucsd.org',
+      credits: 10000,
+    });
+    const invalidMember = UserFactory.fake({
```
shravanhariharan2, line 94: Let's use a random, arbitrary email name with the @acmucsd.org extension instead of someone's name, e.g. `random@acmucsd.org'

shravanhariharan2: Also, technically the @acmucsd.org email is for ACM board members, not any ACM member, so a variable name like `ACMBoardMember` would be a better fit

tests/merchOrder.test.ts
```diff
@@ -84,6 +84,65 @@ describe('merch orders', () => {
...
+      .createMerchItemOptions(affordableOption1)
+      .createOrderPickupEvents(pickupEvent)
+      .write();
+
+    const emailService = mock(EmailService);
+    when(emailService.sendOrderConfirmation(ACMMember.email, ACMMember.firstName, anything()))
+      .thenResolve();
+    when(emailService.sendOrderConfirmation(UCSDMember.email, UCSDMember.firstName, anything()))
+      .thenResolve();
+
```
shravanhariharan2, line 117: For merch store access, we wouldn't even want non-UCSD students to be able to look at the items, let alone order them. So instead of checking that they can't place orders, can we check that they can't view collections? E.g. the `GET /collections` call throws `ForbiddenError`. That mimics what we're actually testing a bit better, plus it makes the test a lot more simpler

PR #244, opened 2021-12-27
Adds ability for users to change their email

### LINKED ISSUE #229: Add Ability to Change Email
Allow a user to change their account email, this should put the user back into the email verification account state.

trulyronak:
Closes #229 - needed since store will be requiring all users to have a ucsd email address to use the store, and we anticipate that users will need to change their email.

Adds new route:

- **POST** `/auth/modifyEmail`
    - Request Body:
        - basic authorization stuff
        - ``` { email: string } ```
    - Response Body:
        - errors 
            - usual 'couldn't find email jazz / couldn't send confirmation email jazz'

Note - in #229 i describe a case where this route might be problematic - worth reading that issue before reviewing this PR

### DIFF
--- a/api/controllers/AuthController.ts
+++ b/api/controllers/AuthController.ts
@@ -7,7 +7,7 @@ import {
   SendPasswordResetEmailResponse,
   VerifyEmailResponse,
   ResendEmailVerificationResponse,
-  ModifyEmailResponse,
+  EmailModificationResponse,
 } from 'types';
 import UserAccountService from '../../services/UserAccountService';
 import UserAuthService from '../../services/UserAuthService';
@@ -18,7 +18,7 @@ import {
   LoginRequest,
   RegistrationRequest,
   PasswordResetRequest,
-  ModifyEmailRequest,
+  EmailModificationRequest,
 } from '../validators/AuthControllerRequests';
 import { authActionMetadata } from '../../utils/AuthActionMetadata';
 import { OptionalUserAuthentication, UserAuthentication } from '../middleware/UserAuthentication';
@@ -72,11 +72,11 @@ export class AuthController {
   }
 
   @UseBefore(UserAuthentication)
-  @Post('/modifyEmail')
-  async modifyEmail(@Body() modifyEmailRequest: ModifyEmailRequest,
-    @AuthenticatedUser() user: UserModel): Promise<ModifyEmailResponse> {
-    await this.userAuthService.modifyEmail(user, modifyEmailRequest.email);
-    await this.emailService.sendEmailVerification(user.email, user.firstName, user.accessCode);
+  @Post('/emailModification')
+  async emailModification(@Body() emailModificationRequest: EmailModificationRequest,
+    @AuthenticatedUser() user: UserModel): Promise<EmailModificationResponse> {
+    const updatedUser = await this.userAuthService.modifyEmail(user, emailModificationRequest.email);
+    await this.emailService.sendEmailVerification(updatedUser.email, updatedUser.firstName, updatedUser.accessCode);
     return { error: null };
   }
 
--- a/api/controllers/AuthController.ts
+++ b/api/controllers/AuthController.ts
@@ -73,7 +73,7 @@ export class AuthController {
 
   @UseBefore(UserAuthentication)
   @Post('/emailModification')
-  async emailModification(@Body() emailModificationRequest: EmailModificationRequest,
+  async modifyEmail(@Body() emailModificationRequest: EmailModificationRequest,
     @AuthenticatedUser() user: UserModel): Promise<EmailModificationResponse> {
     const updatedUser = await this.userAuthService.modifyEmail(user, emailModificationRequest.email);
     await this.emailService.sendEmailVerification(updatedUser.email, updatedUser.firstName, updatedUser.accessCode);
--- a/api/controllers/AuthController.ts
+++ b/api/controllers/AuthController.ts
@@ -7,15 +7,21 @@ import {
   SendPasswordResetEmailResponse,
   VerifyEmailResponse,
   ResendEmailVerificationResponse,
+  ModifyEmailResponse,
 } from 'types';
 import UserAccountService from '../../services/UserAccountService';
 import UserAuthService from '../../services/UserAuthService';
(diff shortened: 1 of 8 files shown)


trulyronak:
> Thanks for contributing! If you've made changes to the API's functionality, please make sure to bump the package version—see [this guide to semantic versioning](https://semver.org/) for details—and document those changes as appropriate.

ran `npm version minor` to address this

shravanhariharan2:
A minor version bump is fine for this change, since there's no changes that aren't backwards-compatible

api/controllers/AuthController.ts
```diff
@@ -65,6 +71,15 @@ export class AuthController {
     return { error: null };
   }
 
+  @UseBefore(UserAuthentication)
+  @Post('/modifyEmail')
+  async modifyEmail(@Body() modifyEmailRequest: ModifyEmailRequest,
+    @AuthenticatedUser() user: UserModel): Promise<ModifyEmailResponse> {
```
shravanhariharan2, line 77: nit: can we rename this route to `/emailModification`? Even though it doesn't sound as clean as `/modifyEmail`, I think we should be consistent in API naming, since the other /auth routes are all nouns (like `/emailVerification` vs. `/verifyEmail` or `/passwordReset` vs. `/resetPassword` - we changed from verbs to nouns during the TS rewrite)

api/controllers/AuthController.ts
```diff
@@ -65,6 +71,15 @@ export class AuthController {
...
+    await this.userAuthService.modifyEmail(user, modifyEmailRequest.email);
+    await this.emailService.sendEmailVerification(user.email, user.firstName, user.accessCode);
+    return { error: null };
```
shravanhariharan2, line 80: nit: can we set the user's email as `toLowerCased`? (see `AuthController::resendEmailVerification()` for implementation). We've had issues in the past where a user accidentally capitalizes a character in their email, and thus wouldn't receive an email from us to that account since it doesn't exist, when in reality it was meant to be lowercased (basically, this allows users to typo the case)

services/UserAuthService.ts
```diff
@@ -42,6 +42,19 @@ export default class UserAuthService {
...
+      const userRepository = Repositories.user(txn);
+
+      await userRepository.upsertUser(user, {
+        email: proposedEmail,
+        state: UserState.PENDING,
+      });
+
+      await this.setAccessCode(proposedEmail);
+    });
+  }
```
shravanhariharan2, line 56: Let's return the updated `user` object in this method instead of return void. We need the updated `email` and `accessCode` fields in `sendEmailVerification`, so it's important that we're updating the `user` param accordingly. I think the line 49 call will update `user.email`, but the line 54 call wouldn't update the  `user` param's `accessCode` field since it's not being directly set (see `setAccessCode` - it's fetching a separate object and updating that one vs. directly updating the `user` param object passed in)

services/UserAuthService.ts
```diff
@@ -42,6 +42,19 @@ export default class UserAuthService {
...
   }
 
+  public async modifyEmail(user: UserModel, proposedEmail: string): Promise<void> {
+    return this.transactions.readWrite(async (txn) => {
+      const userRepository = Repositories.user(txn);
+
+      await userRepository.upsertUser(user, {
+        email: proposedEmail,
+        state: UserState.PENDING,
+      });
```
shravanhariharan2, line 52: We need to make sure that the user isn't setting their email to an existing email. Can you add a check here to make sure they aren't setting to a duplicate email?

shravanhariharan2 (changes requested):
Few comments on the setup, can you also add a quick test in `auth.test.ts` for checking that the user's email is properly updated and its state is properly set to `PENDING` when accessing this route?

services/UserAuthService.ts
```diff
@@ -42,6 +42,27 @@ export default class UserAuthService {
...
+    const updatedUser = await this.transactions.readWrite(async (txn) => {
+      const userRepository = Repositories.user(txn);
+
+      proposedEmail = proposedEmail.toLowerCase();
+
+      // ensure that the proposed email isn't used already
+      const foundUser = await userRepository.findByEmail(proposedEmail);
+      if (foundUser !== undefined) {
+        throw new BadRequestError('Email already in use');
+      }
```
shravanhariharan2, line 55: nit: we can shorten this a bit and use the same logic we use in `UserAuthService::registerUser` for checking duplicates:
```ts
const emailAlreadyUsed = !!(await userRepository.findByEmail(registration.email));
if (emailAlreadyUsed) throw new BadRequestError('Email already in use');
```

tests/auth.test.ts
```diff
@@ -224,6 +224,62 @@ describe('resending email verification', () => {
...
+    await authController.verifyEmail({ accessCode: member.accessCode });
+
+    member = await conn.manager.findOne(UserModel, { uuid: member.uuid });
+    expect(member.state).toEqual(UserState.ACTIVE);
+  });
+
+  test('user cannot change email to new email', async () => {
+    const conn = await DatabaseConnection.get();
+    let member = UserFactory.fake();
+    const otherMember = UserFactory.fake({ email: 'anotheremail@acmucsd.org' });
```
shravanhariharan2, line 261: nit: can we rename this test 'user cannot change email to an already existing email' for more clarity?

api/controllers/AuthController.ts
```diff
@@ -65,6 +71,15 @@ export class AuthController {
     return { error: null };
   }
 
+  @UseBefore(UserAuthentication)
+  @Post('/emailModification')
+  async emailModification(@Body() emailModificationRequest: EmailModificationRequest,
+    @AuthenticatedUser() user: UserModel): Promise<EmailModificationResponse> {
+    const updatedUser = await this.userAuthService.modifyEmail(user, emailModificationRequest.email);
```
shravanhariharan2, line 78: nit: we can leave the method name as `modifyEmail`, since we use nouns for REST endpoints but verbs for method names (since the verb in REST is provided by the request type (GET, POST, etc))

shravanhariharan2 (approved):
LGTM, few small comments but feel free to merge once they're addressed

PR #245, opened 2021-12-28
Add PublicOrderWithItems type to API

shravanhariharan2:
This change addresses this comment (https://github.com/acmucsd/membership-portal/commit/c1a55fc5961b41f5ba94222aa47ba3d5c457f255#r62577745) about a 500 error being thrown on GET /orders. The error was happening because the `PublicOrder` type assumes the join to OrderItems, MerchItemOptions, and MerchItems have been successfully made, but for the GET /orders route, we aren't doing that join.

Adding a `PublicOrderWithItems` type that assumes this join, and removing this assumption from the existing `PublicOrder` type (by removing the `items` field) allows GET /orders to not need that join (by using the `PublicOrder` type), while allowing other routes to still require that join (by using the `PublicOrderWithItems` type).


shravanhariharan2:
Closes #246

sumeet-bansal (approved):

PR #247, opened 2021-12-30
Modify GET /store/collection to order by creation

### LINKED ISSUE #239: Modify GET /store/collection to order by creation
Modify GET /store/collection to order by creation, so that frontend can display newly created collections first on the UI (eg. if there's a limited edition collection, or for listing spring collection ahead of winter collection, etc.)

trulyronak:
Closes #239.

Key Changes
- Adds a `createdAt` column to the `MerchandiseCollectionModel` (is there any reason we're not doing this by default?)
- Makes it so `getAllActiveCollections` and `getAllCollections` sort by the `createdAt` field in `DESC`

Note
- test naming is not my strong suit

### DIFF
--- a/migrations/0032-add-created-at-column-to-merch-collection.ts
+++ b/migrations/0032-add-created-at-column-to-merch-collection.ts
@@ -1,10 +1,8 @@
 import { MigrationInterface, QueryRunner } from 'typeorm';
 
 export class AddCreatedAtColumnToMerchCollection1640851101278 implements MigrationInterface {
-  name = 'AddCreatedAtColumnToMerchCollection1640851101278';
-
   public async up(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query('ALTER TABLE "MerchandiseCollections" ADD "createdAt" TIMESTAMP NOT NULL DEFAULT now()');
+    await queryRunner.query('ALTER TABLE "MerchandiseCollections" ADD "createdAt" TIMESTAMPTZ NOT NULL DEFAULT now()');
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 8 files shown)


migrations/0032-add-created-at-column-to-merch-collection.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddCreatedAtColumnToMerchCollection1640851101278 implements MigrationInterface {
+  name = 'AddCreatedAtColumnToMerchCollection1640851101278';
+
```
shravanhariharan2, line 5: nit: the `name` field here isn't necessary since it defaults to the class name, which is the same in this case. See https://github.com/typeorm/typeorm/blob/master/src/migration/MigrationInterface.ts#L10

migrations/0032-add-created-at-column-to-merch-collection.ts
```diff
@@ -0,0 +1,13 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query('ALTER TABLE "MerchandiseCollections" ADD "createdAt" TIMESTAMP NOT NULL DEFAULT now()');
+  }
+
```
shravanhariharan2, line 9: For dates in our database we use the `timestamptz` instead of the `timestamp` column type, so let's use that type as well for `createdAt` columns.

models/MerchandiseCollectionModel.ts
```diff
@@ -10,6 +10,9 @@ export class MerchandiseCollectionModel extends BaseEntity {
   @Column('varchar', { length: 255 })
   title: string;
 
+  @CreateDateColumn()
+  createdAt: Date;
+
```
shravanhariharan2, line 15: Regarding the `timestamptz` type, you can pass in a `type` property to the `@CreateDateColumn()` decorator as `timestamptz` to explicitly set the type appropriately.
```ts
  @CreateDateColumn({ type: 'timestamptz' })
  createdAt: Date;
```

shravanhariharan2 (approved):
Small comments on type consistency with date columns, good to merge once those are resolved

PR #249, opened 2022-01-04
Add user join for GET /order/pickup/future and /past routes

steets250:
There was a bug when a user was attempting to fetch their own order, in which `this.user` is null in the `OrderModel`, and an error is thrown. I removed the assumption of the user property being defined and it is working as expected. Let me know if this isn't the right remedy for the issue.

Here's the error printout:
```
TypeError: Cannot read properties of undefined (reading 'getPublicProfile')
    at OrderModel.getPublicOrder (/Users/steets250/Storage/GitHub/membership-portal/models/OrderModel.ts:50:23)
    at OrderModel.getPublicOrderWithItems (/Users/steets250/Storage/GitHub/membership-portal/models/OrderModel.ts:60:15)
    at /Users/steets250/Storage/GitHub/membership-portal/models/OrderPickupEventModel.ts:39:77
    at Array.map (<anonymous>)
    at OrderPickupEventModel.getPublicOrderPickupEvent (/Users/steets250/Storage/GitHub/membership-portal/models/OrderPickupEventModel.ts:39:56)
    at /Users/steets250/Storage/GitHub/membership-portal/api/controllers/MerchStoreController.ts:286:8
    at Array.map (<anonymous>)
    at MerchStoreController.<anonymous> (/Users/steets250/Storage/GitHub/membership-portal/api/controllers/MerchStoreController.ts:285:45)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/steets250/Storage/GitHub/membership-portal/api/controllers/MerchStoreController.ts:17:58)
```

### DIFF
--- a/models/OrderModel.ts
+++ b/models/OrderModel.ts
@@ -47,7 +47,7 @@ export class OrderModel extends BaseEntity {
   public getPublicOrder(): PublicOrder {
     return {
       uuid: this.uuid,
-      user: this.user?.getPublicProfile(),
+      user: this.user.getPublicProfile(),
       totalCost: this.totalCost,
       status: this.status,
       orderedAt: this.orderedAt,
--- a/models/OrderModel.ts
+++ b/models/OrderModel.ts
@@ -47,7 +47,7 @@ export class OrderModel extends BaseEntity {
   public getPublicOrder(): PublicOrder {
     return {
       uuid: this.uuid,
-      user: this.user.getPublicProfile(),
+      user: this.user?.getPublicProfile(),
       totalCost: this.totalCost,
       status: this.status,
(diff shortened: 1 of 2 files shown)


shravanhariharan2:
It looks like the error's happening on the GET /order/pickup/future route because only the pickup events and orders are joined, and not the user. Looking back at it now, wouldn't we always want the user to be joined with the order? Like if we're displaying a list of orders, I'm confused how we can even display an order without some user info because there's no like "order name" or anything, and the items aren't being returned here either.

If we do want to have the user always joined with order, then the fix here would be to add that join, rather than make `user` optional, which is why I ask.

shravanhariharan2 (approved):

PR #252, opened 2022-01-05
Properly filter hidden items from users

### LINKED ISSUE #250: Incorporating the Hidden Property into Items
When getting all collections or getting a collection, if an item is hidden and a non-admin is requesting, the item shouldn't be returned. However, it should be returned if the user is an admin, along with a `hidden` property so the admin knows whether or not the item has been hidden.

shravanhariharan2:
Closes #250

### DIFF
--- a/repositories/MerchStoreRepository.ts
+++ b/repositories/MerchStoreRepository.ts
@@ -1,4 +1,4 @@
-import { EntityRepository } from 'typeorm';
+import { EntityRepository, SelectQueryBuilder } from 'typeorm';
 import { MerchandiseItemOptionModel } from '../models/MerchandiseItemOptionModel';
 import { MerchandiseCollectionModel } from '../models/MerchandiseCollectionModel';
 import { MerchandiseItemModel } from '../models/MerchandiseItemModel';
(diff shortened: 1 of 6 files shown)


repositories/MerchStoreRepository.ts
```diff
@@ -41,6 +35,12 @@ export class MerchCollectionRepository extends BaseRepository<MerchandiseCollect
   public async deleteMerchCollection(collection: MerchandiseCollectionModel): Promise<void> {
     await this.repository.remove(collection);
   }
+
+  public getBaseFindManyQuery(): SelectQueryBuilder<MerchandiseCollectionModel> {
```
trulyronak, line 39: nit: since this code refactoring wasn't part of the initial issue, imo it'd be good to call it out in the PR (just so I'm not confused as to why this file was changed)

trulyronak (approved):
lgtm, assuming we already have tests that check to make sure the get items calls work for users with regular permissions (make sure that those tests don't use an admin to check if the right items return)

repositories/MerchStoreRepository.ts
shravanhariharan2: Good call, yeah this change was a quick one for refactoring common logic. Will call those out next time

shravanhariharan2:
> lgtm, assuming we already have tests that check to make sure the get items calls work for users with regular permissions (make sure that those tests don't use an admin to check if the right items return)

We make those assertions in these tests (ie. admins can get all items vs. user can only get visible items are both asserted in the same test)

PR #255, opened 2022-01-08
Allow decrementation of option quantities

shravanhariharan2:
Allows merch item option quantities to be decremented in addition to being incremented, in the event that the original quantity specified was typo'd, or if a set number of option stock needs to be set aside for some other event.

The one potential issue that could theoretically arise is if a member is at cart checkout after adding items to their cart, then an item's option gets decremented to below the quantity requested by the member, and then the member places their order - this will result in the order placement failing. However, even without this change, this case can still happen if other members also order that option before the first member places their order, so the addition of this change shouldn't cause any additional issues in the system. Plus, the only time decrementation would be used is for typo-fixing or for special cases in which stock needs to be reduced, so it is quite rare that this feature would be used in the first place, let alone being used concurrently with someone placing an order.

### DIFF
--- a/api/validators/MerchStoreRequests.ts
+++ b/api/validators/MerchStoreRequests.ts
@@ -9,6 +9,7 @@ import {
   IsHexColor,
   IsDateString,
   ArrayNotEmpty,
+  IsNumber,
 } from 'class-validator';
 import { Type } from 'class-transformer';
 import {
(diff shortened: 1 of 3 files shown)


steets250 (approved):
I can't vouch for the backend-specific code, but the modified logic looks solid overall. 👍

PR #259, opened 2022-01-11
Fix human readable string for email templates

StormFireFox1:
Currently, email templates print the date for the pickup event wrong (display date of week instead of day, and also display time in UTC).

Fix Moment output to display in PST and also display date of pickup event properly.


StormFireFox1:
@shravanhariharan2 Please `npm install moment-timezone` using your Node; mine for some reason breaks the package lockfile.

shravanhariharan2 (approved):

PR #263, opened 2022-01-12
Update merch store limit checks with cancelled orders

### LINKED ISSUE #262: Cancelling an order doesn't update remaining limits
The `getPastItemOrdersByUser` function doesn't check for order status when calculating the number of times the item has been purchased, so cancelled orders still cause the `monthlyRemaining` and `lifetimeRemaining` limits to decrease.

After purchasing an item with a limit of 1, I have 0 remaining. However, after cancelling that order, the remaining stays at 0 but will still allow me to purchase, so the remaining then has a value of -1.
Screen Shot 2022-01-11 at 12 37 07 PM

shravanhariharan2:
Closes #262 .


steets250 (approved):
Revised logic looks good to me.

steets250 (approved):
Awesome! This should be it for the bugs we've encountered with the user-facing side of store 😄

PR #265, opened 2022-01-13
Allow users with admin role to be displayed on leaderboard, except for acm@ucsd.edu

shravanhariharan2:
Since some users have ADMIN access types now, they weren't getting displayed on the leaderboard since we don't return any ADMIN accounts to the client in our leaderboard call. The intended behavior should be to only restrict the acm@ucsd.edu admin account


steets250 (approved):
LGTM 🥳

michl1001 (approved):
Looks good

PR #267, opened 2022-01-21
Add payload to POST /order/:uuid/fulfill

### LINKED ISSUE #266: FulfillMerchOrderResponse Update
For the `@Post('/order/:uuid/fulfill')` route, can we update `FulfillMerchOrderResponse` to include a copy of the order that was updated (including it's new status?). The logic for the frontend fulfillment page is getting pretty messy trying to track pending fulfillment changes and then update the pickup event orders array if fulfillment goes through successfully. It would be a lot easier to have the updated event returned (whether fully or partially fulfilled), so we can ensure that the data displayed matches what's stored in backend.

shravanhariharan2:
Closes #266


steets250 (approved):
Looks great, thanks for taking care of this so quickly 🚀

PR #268, opened 2022-01-21
Set minimum quantity to 1 for fake options

shravanhariharan2:
Currently, the MerchFactory allows for quantity = 0 to be set for an option. This means that, nondeterministically, some tests can fail because when they don't specify a quantity for an option, there is a chance it can equal 0, so when an order is attempted to be placed, an error will be thrown in the test because the item is out of stock.

This PR bumps the minimum quantity to 1 so that any option can be ordered from at least once without having to manually specify the quantity


steets250 (approved):
Logic is sounds to me.

PR #271, opened 2022-01-22
Fix monthly/lifetime limit counting bug

### LINKED ISSUE #269: User can bypass limits by buying different options

shravanhariharan2:
Closes #269 

There was a bug where a user could order different options of an item to bypass the monthly/lifetime limits set for that item. This bug was due to a logical bug in the counting of previous items ordered (in `MerchStoreService::countItemOrders`), where the count would only increment if an option wanting to be ordered now was previously ordered before. 

I've modified the logic of the method to increment counts for every item ordered rather than for only items whose specific options are being ordered now.

### DIFF
--- a/repositories/MerchOrderRepository.ts
+++ b/repositories/MerchOrderRepository.ts
@@ -40,10 +40,26 @@ export class MerchOrderRepository extends BaseRepository<OrderModel> {
   }
 
   /**
-   * Gets all orders for a given user. Returns the order joined with its pickup event.
+   * Gets all orders for a given user. Returns the order joined with its pickup event and user.
    */
   public async getAllOrdersForUser(user: UserModel): Promise<OrderModel[]> {
     return this.repository.find({ user });
+  }
+
+  /**
+   * Gets all orders for a given user. Returns the order joined with its pickup event, user,
+   * merch item options, and merch items.
+   */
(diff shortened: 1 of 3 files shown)


repositories/MerchOrderRepository.ts
```diff
@@ -40,12 +40,28 @@ export class MerchOrderRepository extends BaseRepository<OrderModel> {
...
+    return this.repository
+      .createQueryBuilder('order')
+      .leftJoinAndSelect('order.pickupEvent', 'orderPickupEvent')
+      .leftJoinAndSelect('order.items', 'orderItem')
+      .leftJoinAndSelect('order.user', 'user')
+      .leftJoinAndSelect('orderItem.option', 'option')
+      .leftJoinAndSelect('option.item', 'merchItem')
+      .where('order.user = :uuid', { uuid: user.uuid })
+      .getMany();
+  }
```
shravanhariharan2, line 63: This new query is needed so that the join happening in the following code is possible: 
https://github.com/acmucsd/membership-portal/pull/271/files#diff-8f5e873f32fd041a327e36c028853a65ced08b99cb09572d3d785ef193b0c212R804-R806

michl1001 (approved):
If I understand the query correct, this looks good.

PR #272, opened 2022-01-23
Add pickup event status field

### LINKED ISSUE #270: Add Pickup Event Status
A pickup event should be active by default and behave normally. Ending the pickup event should set its status to finished, and no more orders should be able to be fulfilled, and it shouldn't be able to be refinished.

shravanhariharan2:
Closes #270


steets250 (approved):
Looks good from a logic standpoint 👏

PR #274, opened 2022-01-24
Allow order pickup to be rescheduled for missed/cancelled events

### LINKED ISSUE #273: Unable to reschedule PICKUP_MISSED Order
Screen Shot 2022-01-24 at 1 19 28 PM
Screen Shot 2022-01-24 at 1 19 33 PM

shravanhariharan2:
Closes #273 

Previously we checked for reschedule/cancelling orders if the pickup event is less than 2 days away. The issue is, we do this check for orders with any status, so when the order is MISSED, PICKUP_CANCELLED, or PARTIALLY_FULFILLED and the pickup event has passed, the reschedule/cancel fails.

We want to change it so that only orders that are PLACED have this check in place, since otherwise, with any other state, the pickup event has already passed / is ongoing.


steets250 (approved):
LGTM 🚀

PR #286, opened 2022-02-08
Add guide for testing

shravanhariharan2:


.github/CONTRIBUTING.md
```diff
@@ -59,7 +59,180 @@ Modifying the schema requires not only direct edits to the data models (found in
...
+"Software engineering is the integral of coding over time"; to build a reliable system over several years, we rely on tests to verify that we've implemented changes correctly without breaking anything unexpected in the process, that we throw errors when appropriate, and that we've handled forgettable edge cases. 
+
+The tests we've defined can be found in the `tests` folder, and currently contain tests for the API layer. These tests are written with [Jest](https://jestjs.io/), with mocking done with (ts-mockito)[https://github.com/NagRock/ts-mockito] (a TypeScript version of the popular [Mockito](https://site.mockito.org/) library in Java). We don't test specific functions in the service or repository layer, since given the heirarchical structure of our layers, correctness at the API layer implies correctness in the service and repository layers.
```
michl1001, line 64: My main worry here is that there's a lot of jargon that might make this hard to read for newcomers

.github/CONTRIBUTING.md
```diff
@@ -59,7 +59,180 @@ Modifying the schema requires not only direct edits to the data models (found in
...
+
+#### Test Structure
+Every test consists of 3 main parts: setup, execution, and assertion. The setup step involves creating any data that is needed for that specific functionality to be tested. Below is an example merch store test that ensures members earn both points and credits for attending an event. The setup involves retrieving a `Connection` object via. the `DatabaseConnection` class (so we can write data to the database later on), and creating a fake member to check into a fake event. Every test will need to start by getting a `Connection` object, because the actual operations that take place on persistent data depend on it (see section on `ControllerFactory`).
```
michl1001, line 71: We highlight the 3 parts of a test but the following sections aren't very clear on what section corresponds to where - maybe we could add subtitles fot each part

(i.e. under setup there is factories, portal state)

michl1001: Also, mocking is more in the execution step, shouldn't it be in the setup with the factories?

michl1001 (approved):
The content is generally good and clear, the main issues I could see are organizational and wording - mostly thing that might make it easier for newcomers

.github/CONTRIBUTING.md
StormFireFox1, line 64: The link to ts-mockito is slightly wrong:
```suggestion
The tests we've defined can be found in the `tests` folder, and currently contain tests for the API layer. These tests are written with [Jest](https://jestjs.io/), with mocking done with [ts-mockito](https://github.com/NagRock/ts-mockito) (a TypeScript version of the popular [Mockito](https://site.mockito.org/) library in Java). We don't test specific functions in the service or repository layer, since given the heirarchical structure of our layers, correctness at the API layer implies correctness in the service and repository layers.
```

StormFireFox1: I think this is fine, personally; it's useful for those writing unit tests, which are likely already _somewhat_ knowledgeable with the backend by that point (and also they should've read everything above this section as well).

StormFireFox1 (approved):
Looks great! One little nit.

.github/CONTRIBUTING.md
shravanhariharan2: I personally think its fine too, especially since the layering architecture is covered above. So any new jargon is strictly related to testing, which is all explained in the below sections.

.github/CONTRIBUTING.md
shravanhariharan2: > We highlight the 3 parts of a test but the following sections aren't very clear on what section corresponds to where - maybe we could add subtitles fot each part

Good point, updated the sections with subtitles.

> Also, mocking is more in the execution step, shouldn't it be in the setup with the factories?

While I partially agree here, I think mocking is more directly tied to the execution step, since we need to know what services are being executed, mock those services, and pass the mocked instance into the `ControllerFactory` call, which is used primarily in the execution step.

PR #290, opened 2022-02-27
Migrate npm to yarn, plus small repo improvements

### LINKED ISSUE #284: Migrate package management to Yarn
Ideally, this would not only bring the package manager to be consistent with all TS projects in Dev Team, it'd also provide Workspaces as an abstraction to make publishing type bindings for the portal easier.

shravanhariharan2:
Closes #284 

Migrated npm to yarn, with additional small repo improvements. These include:
* Using multi-stage docker builds for the Docker image (reducing image size from ~1.8GB to ~300MB)
* Removed dead ormconfig.ts code for our old method of seeding
* Matching .env.example fields with the ones in my local .env


michl1001 (approved):
Nothing I can see that stands out as particularly wrong, but also not really able to decode the entire package.json file. The yml files look right

.circleci/config.yml
```diff
@@ -7,24 +7,24 @@ jobs:
     steps:
       - checkout
       - run:
-          command: npm ci
+          command: yarn
```
steets250, line 10: I think this should be `yarn install --frozen-lockfile` to prevent yarn from auto updated anything in `yarn.lock` when running ci testing.

StormFireFox1, line 10:
```suggestion
          command: yarn install --frozen-lockfile
```

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-*	@sumeet-bansal @shravanhariharan2
+*	@sumeet-bansal @shravanhariharan2 @michl1001
```
StormFireFox1, line 1: Truly a beautiful sight to see! 😀

Dockerfile
```diff
@@ -1,9 +1,13 @@
-FROM node:14.15.3
+FROM node:14.15.3 AS build
 
 WORKDIR /app
-COPY package.json package-lock.json ./
-RUN npm install
+COPY package.json yarn.lock ./
+RUN yarn
```
StormFireFox1, line 5:
```suggestion
RUN yarn install --frozen-lockfile
```

StormFireFox1 (changes requested):
Looks good, couple of nits.

StormFireFox1 (approved):
Perfect!

.circleci/config.yml
```diff
@@ -7,24 +7,24 @@ jobs:
...
     docker:
       - image: circleci/node:14.17.6
     steps:
       - checkout
       - restore_cache:
-          key: package-cache-{{ checksum "package-lock.json" }}
+          key: package-cache-{{ checksum "yarn.lock" }}
       - run:
-          command: npm install
+          command: yarn
```
steets250, line 21: Oop guess I forgot to mark all the instances of `yarn` but should these also be `yarn install --frozen-lockfile`? I see it was `nom install` instead of `npm ci` and wasn't sure if that was intentional.

PR #292, opened 2022-02-27
Set metrics recorder priority to higher than rest of middleware

### LINKED ISSUE #291: Fix Datadog API metrics issue
Currently, only the below API routes would have metrics posted on Datadog
image

This is due to an mis-prioritization between the MetricsRecorder middleware and the NotFoundHandler middleware. Both of these middlewares are ran after every request, but the default seems to be that NotFoundHandler is ran before MetricsRecorder. When an API route is successfully hit and the NotFoundHandler triggers (but not actually throwing a 'not found error'), the line `response.end()` in the handler stops any other middleware after it from running, so MetricsRecorder never gets to run. Only when a 'not found error' is thrown, the MetricsRecorder can actually run, since the `response.end()` line doesn't run in that case (because a JS error is thrown before that, presumably letting all the other middlewares finish before terminating the request).  

The fix is to set the priority of MetricsRecorder higher than NotFoundHandler, so that MetricsRecorder can always run after each request.

shravanhariharan2:
Closes #291

### DIFF
--- a/api/middleware/MetricsRecorder.ts
+++ b/api/middleware/MetricsRecorder.ts
@@ -2,7 +2,7 @@ import { ExpressMiddlewareInterface, Middleware } from 'routing-controllers';
 import * as express from 'express';
 import * as metrics from 'datadog-metrics';
 
-@Middleware({ type: 'after' })
+@Middleware({ type: 'after', priority: 1 })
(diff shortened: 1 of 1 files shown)


michl1001 (approved):
Looks good

sumeet-bansal:
nice

PR #293, opened 2022-03-08
Order error

michl1001:
Fixes a bug where someone ordering two item-options will not get rejected for going over the item limit

- new MerchItemWithQuantity type
- counting order items now uses uuid as a key


services/MerchStoreService.ts
```diff
@@ -827,13 +832,21 @@ export default class MerchStoreService {
...
   private static countItemRequestedQuantities(order: MerchItemOptionAndQuantity[],
-    itemOptions: Map<string, MerchandiseItemOptionModel>): Map<MerchandiseItemModel, number> {
-    const requestedQuantitiesByMerchItem = new Map<MerchandiseItemModel, number>();
+    itemOptions: Map<string, MerchandiseItemOptionModel>): Map<string, MerchItemWithQuantity> {
+    const requestedQuantitiesByMerchItem = new Map<string, MerchItemWithQuantity>();
     for (let i = 0; i < order.length; i += 1) {
-      const { item } = itemOptions.get(order[i].option);
+      const opt = itemOptions.get(order[i].option);
+
+      const { item } = opt;
```
shravanhariharan2, line 840: nit: let's avoid shorthand for clarity, so `option` vs. `opt`.

tests/merchOrder.test.ts
```diff
@@ -590,6 +590,44 @@ describe('merch orders', () => {
...
+    const item = MerchFactory.fakeItem({
+      options: [option1, option2],
+      monthlyLimit: 1,
+      lifetimeLimit: 1,
+    });
+    const orderPickupEvent = MerchFactory.fakeFutureOrderPickupEvent();
+    const member = UserFactory.fake({
+      accessType: UserAccessType.STANDARD,
+      credits: 1 * option1.price + 1 * option2.price,
+    });
```
shravanhariharan2, line 607: Why do we have `(1 * option1.price)` and `(1 * option2.price)` here? I think credits should be `(100 * option.price)`.

tests/merchOrder.test.ts
```diff
@@ -590,6 +590,44 @@ describe('merch orders', () => {
...
+
+    await new PortalState()
+      .createUsers(member)
+      .createOrderPickupEvents(orderPickupEvent)
+      .createMerchItem(item)
+      .write();
+
+    member.reload();
+
```
shravanhariharan2, line 616: Does this need to be `await`ed? I think `member.reload()` returns a `Promise`.

tests/merchOrder.test.ts
```diff
@@ -590,6 +590,44 @@ describe('merch orders', () => {
...
+    const merchController = ControllerFactory.merchStore(conn, instance(emailService));
+    const placeMerchOrderRequest = {
+      order: [{ option: option1.uuid, quantity: 1 }, { option: option2.uuid, quantity: 1 }],
+      pickupEvent: orderPickupEvent.uuid,
+    };
+    // console.log(placeMerchOrderRequest.order)
+    // await merchController.placeMerchOrder(placeMerchOrderRequest,member);
+    await expect(merchController.placeMerchOrder(placeMerchOrderRequest, member)).rejects.toThrowError(HttpError);
+  });
```
shravanhariharan2, line 629: nit: let's remove any commented-out code

shravanhariharan2, line 629: I'm still a bit skeptical on why an `HttpError` is being thrown here instead of a `UserError`. Do we know that the error being thrown is this exact error?
```ts
throw new UserError(`This order exceeds the lifetime limit for ${itemWithQuantity.item.itemName}`);
```

michl1001: I think there is some bug in Jest since when I waited for `UserError` threw an error when it calls the super constructor (effectively making it an HttpError as far as Jest understands)

tests/merchOrder.test.ts
michl1001: Are the option prices and user credits stored as different scales? The "1 *" was leftover from some experimentation and can be removed

michl1001: Also, just checked, in either case, the limit is checked before the credits so this shouldn't be an issue unless the limit fires first

tests/merchOrder.test.ts
michl1001: Just made a fix that looks for the specific error message so that should work better

tests/merchOrder.test.ts
shravanhariharan2: Oh my bad, yeah credits are the same as price so we can omit the 1.

shravanhariharan2 (approved):
LGTM, going to merge this before kickoff because we want to make sure people can't surpass the limits after we market store during kickoff.

Issue #288, opened 2022-02-15
getAllEvents is missing a return type


steets250:
https://github.com/acmucsd/membership-portal/blob/9310daf9f9dffd3584c0f017c96a209ed904be3e/api/controllers/EventController.ts#L113

nik-dange:
@dowhep is this closed? It looks like Shane merged something in related to it from last year

PR #295, opened 2022-03-08
Update return type for getAllEvents

shanekim28:
Updated return type of `getAllEvents` to utilize the `GetAllEventsResponse` DTO

Responds partly to #288 

Held off on updating the response to `cancelMerchOrder`. There was a discrepancy in the behaviours of `cancelMerchOrder`, `fulfillMerchOrderItems`, and `rescheduleOrderPickup`. I noticed `fulfillMerchOrderItems` returns a `FulfillMerchOrderResponse` DTO and modifies the respective order, but the other two routes do not return anything despite modifying the model. Not sure if this was intended, but may be worth looking into for consistency.

### DIFF
--- a/api/controllers/EventController.ts
+++ b/api/controllers/EventController.ts
@@ -111,10 +111,11 @@ export class EventController {
 
   @UseBefore(OptionalUserAuthentication)
   @Get()
-  async getAllEvents(@QueryParams() options: EventSearchOptions, @AuthenticatedUser() user: UserModel): Promise<GetAllEventsResponse> {
+  async getAllEvents(@QueryParams() options: EventSearchOptions, @AuthenticatedUser() user: UserModel):
+  Promise<GetAllEventsResponse> {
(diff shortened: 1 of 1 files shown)


shravanhariharan2:
Thanks for contribuing! Our type system for our controllers and services are currently a bit incomplete / inconsistent (e.g. missing return types for controllers, returning `Public` types vs. `Model` types in the services, to name a few), and we're looking to refactor these during our upcoming undertaking of exporting a types package. We're going to discuss these discrepancies with the rest of the development team to iron out where responses should be returned vs. where they shouldn't (i.e. figuring out where the frontend should or shouldn't be using the responses to their API calls), so you can expect that to be resolved within a few weeks.

For now, we can merge in this PR to patch the getAllEvents route, and we can handle the rest of the type discrepancies in future PRs relating to that package refactor.

shravanhariharan2 (approved):

PR #296, opened 2022-03-16
Make get single event authentication optional

farisashai:
disclaimer: i know basically nothing about back-end so i'm not sure how to run this code or if anything works. 

i need to use this endpoint on main website to get event data and we don't have any required log-in structure there so i want to remove authentication.

 I tried to model it off of the endpoint for getAllEvents so we can fetch individual events without authentication while still using the authentication if it was provided and necessary for attendance data. hopefully this wont break any existing functionality.

edit: tested locally and everything works

### DIFF
--- a/api/controllers/EventController.ts
+++ b/api/controllers/EventController.ts
@@ -80,11 +80,11 @@ export class EventController {
     return { error: null };
   }
 
-  @UseBefore(UserAuthentication)
+  @UseBefore(OptionalUserAuthentication)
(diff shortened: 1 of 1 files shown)


trulyronak (approved):
lgtm, we might wanna add tests later, but we need this route asap for main website events

michl1001 (approved):
The code looks fine syntax-wise, not as familiar with our middleware but nothing looks outof place to me

PR #297, opened 2022-03-21
Upgrade docker version to fix CircleCI image build failure

shravanhariharan2:
Fix: https://github.com/yarnpkg/yarn/issues/8389. We set the docker version to 20 instead of 19 to match our current docker version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,7 +86,9 @@ jobs:
       - run:
           name: Push application Docker image
           command: |
-              echo "test"
+            echo "$DOCKER_PASS" | docker login --username $DOCKER_USER --password-stdin
+            docker tag app "acmucsd/membership-portal-api:latest"
+            docker push "acmucsd/membership-portal-api:latest"
   deploy:
     machine:
       enabled: true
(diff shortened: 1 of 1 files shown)


trulyronak:
@shravanhariharan2 i'm not fully understanding why prior prs (https://github.com/acmucsd/membership-portal/pull/296) passed all tests but failed once merged; if there's a step that only occurs on the main branch, can we ensure that that test also occurs for prs?

shravanhariharan2:
@trulyronak CircleCI runs an `image` step on master to build a Docker image of the application before deploying it with the `deploy` step onto caprover. We didn't have this step on any other branch since the image would only be used for deployment, and adding an image step would increase build times substantially (image step takes anywhere from 5-10 minutes). Plus, building the image would only potentially fail on large configuration changes (like moving npm to yarn), so I'm not immediately seeing the value of adding this step to other branches (since it would be then ran on every commit pushed to that branch). 

I'm thinking specifically for configuration changes, we can temporarily add the image step to the branch to make sure everything still works, but I think the overhead for having this run even for feature changes isn't necessarily worth the extra build time. I can add an excerpt on testing the image step for configuration changes to our contributing guidelines.

trulyronak (approved):
sgtm

PR #298, opened 2022-03-27
Fix states for cancel all orders route

shravanhariharan2:
Fixes a bug where the POST /merch/order/cancel route was only cancelling orders that were PLACED, and not orders that could also be PICKUP_MISSED, PICKUP_CANCELLED, PARTIALLY_FULFILLED. Also improves an existing test for this route to make sure all order states properly transition to either the FULFILLED or CANCELLED state with the usage of this route.


michl1001 (approved):
The changes look fine, only worry was refunding but since the previous version already refunds correctly, should be fine, LGTM

shravanhariharan2:
I'll be manually flipping database values to fix all the existing states since new orders have been placed, so we won't be running this route until the end of Spring quarter.

PR #299, opened 2022-03-29
Add email template for cancel all orders confirmation

shravanhariharan2:
Adds a new template to let members know that their order was cancelled second-hand because they didn't pick up their merchandise by the end of the quarter, since otherwise they might be confused why their order was suddenly cancelled.

Example email:
image

### DIFF
--- a/repositories/MerchOrderRepository.ts
+++ b/repositories/MerchOrderRepository.ts
@@ -1,4 +1,4 @@
-import { EntityRepository, SelectQueryBuilder } from 'typeorm';
+import { EntityRepository, In, SelectQueryBuilder } from 'typeorm';
 import { OrderStatus, Uuid } from '../types';
 import { OrderModel } from '../models/OrderModel';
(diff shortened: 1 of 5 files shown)


shravanhariharan2:
@michl1001 thoughts on email wording as well?

michl1001:
Might be good to say "recent quarter" rather than "past quarter" for clarity

michl1001:
"You placed an order on...." might be clearer than "you had placed an order on..."

michl1001:
This is definitely optional but we could have something like: "Let us know at [email here] if there are any issues!"

shravanhariharan2:
> This is definitely optional but we could have something like: "Let us know at [email here] if there are any issues!"

Wasn't displayed in the picture I sent but we have this in the footer of the email already.

michl1001 (approved):
Looks good, but hard to really test from just looking - main worries include possible bugs with refunding system as it passes to the email but if it pasted prior tests, should be fine

PR #301, opened 2022-03-31
Remove PLACED from cancel all orders check

shravanhariharan2:
The PLACED order status does not need to be cancelled in this route, since orders can be placed for future events that do not need to be cancelled


shravanhariharan2:
the commits in this branch are messed up I'm going to recreate this PR

PR #302, opened 2022-03-31
Remove PLACED from cancel all orders check

shravanhariharan2:
Removes the PLACED state from the 'cancel all orders' check since orders that are PLACED shouldn't be cancelled (only orders that are PARTIALLY_FULFILLED, PICKUP_MISSED, or PICKUP_CANCELLED), since orders can exist for future pickup events when the cancel route gets ran.

### DIFF
--- a/services/MerchStoreService.ts
+++ b/services/MerchStoreService.ts
@@ -903,7 +903,6 @@ export default class MerchStoreService {
 
   private static pendingOrderStatuses(): OrderStatus[] {
     return [
-      OrderStatus.PLACED,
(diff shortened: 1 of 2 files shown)


michl1001 (approved):
Looks fine, main concern is if this route gets called while there still is a scheduled pick up event (to my understanding, this shouldn't happen anyways)

shravanhariharan2:
The existence of a scheduled pickup event is fine since we are removing PLACED from the orders check. If we hadn't removed PLACED, then this route would cancel orders for that pickup event as well, but now it wouldn't.

PR #305, opened 2022-04-13
Fix pickup completion check

### LINKED ISSUE #304: Unable to Finish Pickup Event after end
This is the error message shown when trying to finish a pickup event once the end time has passed. Can we modify the restriction such that the pickup event can be ended any time the day of the pickup event?

Screen_Shot_2022-04-11_at_5 02 41_PM

michl1001:
Closes #304.

Adds tests to ensure working functionality

### DIFF
--- a/services/MerchStoreService.ts
+++ b/services/MerchStoreService.ts
@@ -1018,7 +1018,8 @@ export default class MerchStoreService {
       if (!MerchStoreService.isActivePickupEvent(pickupEvent)) {
         throw new UserError('Cannot complete a pickup event that isn\'t currently active');
       }
-      if (!(MerchStoreService.isPickupEventHappeningToday(pickupEvent) || MerchStoreService.isPickupEventPast(pickupEvent))) {
+      if (!(MerchStoreService.isPickupEventHappeningToday(pickupEvent)
+        || MerchStoreService.isPickupEventPast(pickupEvent))) {
         throw new UserError('Cannot complete a pickup event that\'s not happening today or has past');
       }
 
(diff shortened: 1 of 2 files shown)


shravanhariharan2:
Instead of checking if the pickup is in the present or past, can we just check if its in the future? i.e. if a pickup event's start time is in the future, don't allow the event to be completed.

shravanhariharan2:
Also, can we rename this PR to match the naming convention from other PRs? (e.g. present tense declarative, like "Implement the ACM stock market"

services/MerchStoreService.ts
```diff
@@ -884,6 +884,14 @@ export default class MerchStoreService {
...
   }
 
+  private static isFuturePickupEvent(pickupEvent: OrderPickupEventModel): boolean {
+    return moment().isBefore(moment(pickupEvent.start), 'day');
+  }
+
+  private static isPastPickupEvent(pickupEvent: OrderPickupEventModel): boolean {
+    return moment().isAfter(moment(pickupEvent.start), 'day');
+  }
+
```
shravanhariharan2, line 894: If we specify the granularity `'day'`, then we are only comparing the days of the two rather than the actual times. So if the pickup event is at 4pm and it's 2pm right now, then `isFuturePickupEvent` would be false because today is not before today. So we don't want to specify the granularity of `day`. Can we add a test to check that we can't complete a pickup event on the same day if the time is before the pickup event?

also nit: I don't think isPastPickupEvent is being used so let's remove it

shravanhariharan2 (changes requested):

shravanhariharan2 (approved):

PR #309, opened 2022-05-18
Allow items to be fulfilled during and after a pickup event starts

### LINKED ISSUE #308: Reduce store deadlines
For merch pickup events + order fulfillment, can we loosen restrictions such that as long as the pickup event occurs on or before the current day, it's allowed to have orders fulfilled + marked as complete? We're running into an issue where admins are blocked if they forget to mark an order as fulfilled or try to mark a pickup event as complete after the end time has arrived.

shravanhariharan2:
Closes #308


steets250 (approved):
The condition update looks solid, didn't read through the test thoroughly though 🤫

PR #310, opened 2022-07-11
Update API Types

steets250:
The private user profile previously had the `string` type, but it's correct typing would be `UserAccessType`.


types/ApiResponses.ts
```diff
@@ -144,6 +146,7 @@ export interface PublicMerchCollection {
   title: string;
   themeColorHex?: string;
   description: string;
+  archived?: boolean;
   items: PublicMerchItem[];
```
shravanhariharan2, line 150: Is there a reason we want to expose the `archived` boolean here? For regular users [we don't return archived collections](https://github.com/acmucsd/membership-portal/blob/master/repositories/MerchStoreRepository.ts#L22), so `archived` would always be false here. Revealing archived collections IMO is store data that we wouldn't necessarily want users to be able to view, cause it shows past history about the store that they don't need to know about (even if it's not displayed on the UI, it'd be visible in DevTools)

steets250: The archived type is revealed to store admins, so they have the ability to still see what items have been archived in the event that they choose to unarchive them, or just for viewing purposes. We don't have a separate collection type for admins, so the property was included in the regular collection type, but is optional.

shravanhariharan2: Apologies for the late follow up lol, but to follow up with this behavior, when store admins view merch collections, they view the raw `MerchandiseCollectionModel[]` array, which contains the `archived` field. Only regular users actually see the `PublicMerchandiseCollection` objects, which wouldn't need to display the `archived` since they only see non-archived collections.

I'll take over this PR and remove the `archived` field addition here, but keep the rest of the changes

shravanhariharan2 (approved):

Issue #311, opened 2022-08-04
Implement resume uploads


shravanhariharan2:
Implement an API route that allows users to upload their resume. This route should upload the resume to S3 and store the URL link in the database

PR #312, opened 2022-08-04
Implement resume uploads

shanekim28:
Addreses #311
* Added resume url column
* Added a resume upload route in the Users controller
* Created new filetype enums
* Added constants for upload path


shravanhariharan2 (commented):
Nice work! Minor comments, should be ready to merge once the prod & testing env changes mentioned below are addressed

shravanhariharan2:
Also, let's add some tests to make sure s3 gets passed in the right config and that non-pdf files get properly rejected

config/index.ts
```diff
@@ -60,9 +60,11 @@ export const Config = {
     MAX_EVENT_COVER_FILE_SIZE: Number(process.env.MAX_EVENT_COVER_FILE_SIZE) * BYTES_PER_KILOBYTE,
     MAX_BANNER_FILE_SIZE: Number(process.env.MAX_BANNER_FILE_SIZE) * BYTES_PER_KILOBYTE,
     MAX_MERCH_PHOTO_FILE_SIZE: Number(process.env.MAX_MERCH_PHOTO_FILE_SIZE) * BYTES_PER_KILOBYTE,
+    MAX_RESUME_FILE_SIZE: Number(process.env.MAX_RESUME_FILE_SIZE) * BYTES_PER_KILOBYTE,
     PROFILE_PICTURE_UPLOAD_PATH: process.env.BASE_UPLOAD_PATH + process.env.PROFILE_PICTURE_UPLOAD_PATH,
     EVENT_COVER_UPLOAD_PATH: process.env.BASE_UPLOAD_PATH + process.env.EVENT_COVER_UPLOAD_PATH,
     BANNER_UPLOAD_PATH: process.env.BASE_UPLOAD_PATH + process.env.BANNER_UPLOAD_PATH,
+    RESUME_UPLOAD_PATH: process.env.BASE_UPLOAD_PATH + process.env.RESUME_UPLOAD_PATH,
     MERCH_PHOTO_UPLOAD_PATH: process.env.BASE_UPLOAD_PATH + process.env.MERCH_PHOTO_UPLOAD_PATH,
```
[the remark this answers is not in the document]
shanekim28: Caprover's been updated to include these environment variables. Will update .env.example on next push.

shanekim28:
@shravanhariharan2 The upload resume route does not throw an error when the resume is too big; it allows the resume to be uploaded regardless of file size. The file options use the .env file for max resume upload file size. When comparing the actual and max file sizes, the actual is indeed larger than the max, but it still allows it to upload.

Are there no automatic file size checks when the file is uploaded to the route?

shravanhariharan2:
> @shravanhariharan2 The upload resume route does not throw an error when the resume is too big; it allows the resume to be uploaded regardless of file size. The file options use the .env file for max resume upload file size. When comparing the actual and max file sizes, the actual is indeed larger than the max, but it still allows it to upload.
Are there no automatic file size checks when the file is uploaded to the route?

The reason why an error is not being thrown is because we are directly calling `userController.updateResume()` in our test code, which does not actually call the middleware. If you were to make an API request over Postman to this route, it would properly perform file size checks.

I'd love to say that there's an easy solution here, but to be honest we haven't figured out middleware testing in this codebase, so my inclination is to skip testing for file sizes and at least make sure it works manually, and we can discuss a middleware testing plan down the road and revisit this specific case then. What do you think?

shravanhariharan2:
Holding off on a full review until we have the resume upload requirements fully figured out: https://discord.com/channels/573028991527550986/791028414538186772/1022205693895376906

shanekim28:
> > @shravanhariharan2 The upload resume route does not throw an error when the resume is too big; it allows the resume to be uploaded regardless of file size. The file options use the .env file for max resume upload file size. When comparing the actual and max file sizes, the actual is indeed larger than the max, but it still allows it to upload.
> > Are there no automatic file size checks when the file is uploaded to the route?
> 
> The reason why an error is not being thrown is because we are directly calling `userController.updateResume()` in our test code, which does not actually call the middleware. If you were to make an API request over Postman to this route, it would properly perform file size checks.
> 
> I'd love to say that there's an easy solution here, but to be honest we haven't figured out middleware testing in this codebase, so my inclination is to skip testing for file sizes and at least make sure it works manually, and we can discuss a middleware testing plan down the road and revisit this specific case then. What do you think?

It feels a little dirty to leave it out, but I also think it's the best option until we have something figured out with the middleware. I'll update the tests asap

shanekim28:
TODO:
- [ ] Store resumes in folder with name of user's UUID in AWS bucket using user's file name instead of user's UUID
- [ ] Create opted-in column (is resume visible)
- [ ] Create upload date column

tests/resume.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    const resumeController = ControllerFactory.resume(conn, Mocks.storage(fileLocation));
+    const response = await resumeController.updateResume(resume, member);
+    expect(response.error).toBe(null);
+    expect(response.resume.url).toBe(fileLocation);
+  });
+
+  test('updating resume deletes all previous resumes', async () => {
+    const conn = await DatabaseConnection.get();
+    const member = UserFactory.fake();
```
shravanhariharan2, line 41: Can we verify that the storage mocks for deleting and uploading files properly get called here too?

shravanhariharan2 (changes requested):

api/controllers/ResumeController.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  @Post()
+  async updateResume(@UploadedFile('file',
+    { options: StorageService.getFileOptions(MediaType.RESUME) }) file: File,
+    @AuthenticatedUser() user: UserModel): Promise<UpdateResumeResponse> {
+    if (path.extname(file.originalname) !== '.pdf') throw new BadRequestError('Filetype must be \'.pdf\'');
+
+    const oldResumes = await this.resumeService.getUserResumes(user);
+    await this.storageService.deleteAtUrls(oldResumes.map((resume) => resume.url));
+
```
shravanhariharan2, line 30: I think instead of having a `deleteAtUrls`, we can just do a findOne query on resumes and delete that single resume. The reason being is we'd never really have the use case for deleting all other resumes for a user - for a "single resume per user" case, we just delete the previous one and upload the new one, and for a "multiple resumes per user" case, we would either just not delete any resumes, or delete single resumes within a category (e.g. if a user uploads a DEVOPS resume, and then another DEVOPS resume, we would delete the single previous DEVOPS resume.), depending on the requirements.

So I think it makes sense to keep it simple and just have a "deleteAtUrl" function that we're calling here, and not have a "deleteAtUrls" one that will always just run once.

shravanhariharan2 (commented):
Great progress! One comment on our `deleteAtUrls` case, and another comment from the past review on testing that the storage mocks properly get called.

tests/mocks/MockFactory.ts
```diff
@@ -0,0 +1,10 @@
+import { anything, instance, mock, when } from 'ts-mockito';
+import StorageService from '../../services/StorageService';
+
+export default class Mocks {
+  public static storage(fileLocation: string): StorageService {
+    const storageMock = mock(StorageService);
+    when(storageMock.uploadToFolder(anything(), anything(), anything(), anything())).thenResolve(fileLocation);
+    return instance(storageMock);
+  }
+}
```
shravanhariharan2, line 10: We also need to mock the other StorageService functions here, so that no actual S3 calls get made. Right now, only `uploadToFolder` is the only fn that's mocked, but we also need to mock `upload` and `delete`.

models/ResumeModel.ts
```diff
@@ -12,7 +12,7 @@ export class ResumeModel extends BaseEntity {
   @Index('resumes_by_user_index')
   user: UserModel;
 
-  @Column('boolean', { default: false, nullable: false })
+  @Column('boolean', { default: true, nullable: false })
   isResumeVisible: boolean;
```
shravanhariharan2, line 16: Should we make this default true instead of false? I'd imagine we'd only want a user to have their resume viewable by recruiters if they specify they want to, otherwise we should hide by default. Let me know if you have any opinions on this though

tests/data/FileFactory.ts
```diff
@@ -0,0 +1,40 @@
...
+    return {
+      fieldname: 'image',
+      originalname: 'image.jpg',
+      filename: 'image.jpg',
+      mimetype: 'image/jpg',
+      buffer,
+      encoding: '7bit',
+      size,
+      // the below fields are unused for in-memory multer storage but are still required by type def
+      stream: null,
```
shravanhariharan2, line 35: nit: let's move this comment to the `pdf()` fn since that is the first occurence of these fields in the file

tests/resume.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    await new PortalState()
+      .createUsers(member)
+      .write();
+
+    const resume = FileFactory.pdf(Config.file.MAX_RESUME_FILE_SIZE / 2);
+    const fileLocation = 'fake location';
+
+    const storageService = Mocks.storage(fileLocation);
+    const storageSvcInstance = instance(storageService);
+    const resumeController = ControllerFactory.resume(conn, storageSvcInstance);
```
shravanhariharan2, line 37: nit: we can inline `storageSvcInstance` since we aren't using it anywhere else:
```ts
const resumeController = ControllerFactory.resume(conn, instance(storageService));
```

tests/resume.test.ts
```diff
@@ -0,0 +1,94 @@
...
```
shravanhariharan2, line 58: nit: inline

tests/resume.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    expect(userResumes.length).toBe(1);
+    verify(storageService.deleteAtUrl(fileLocation)).called();
+    verify(storageService.uploadToFolder(resume, MediaType.RESUME, anything(), anything())).called();
+    verify(storageService.uploadToFolder(newResume, MediaType.RESUME, anything(), anything())).called();
+  });
+
+  test('wrong filetype', async () => {
+    const conn = await DatabaseConnection.get();
```
shravanhariharan2, line 80:
```suggestion
  test('uploading resumes with the wrong filetype throws an error', async () => {
    const conn = await DatabaseConnection.get();
```

shravanhariharan2 (approved):
Awesome work! Love the coverage with this test suite. Last few comments, you're welcome to merge once they've all been addressed

PR #315, opened 2022-08-04
Create Resume Opted Option

dowhep:
This creates a boolean field in the user info to choose whether to opt-in resume sharing for the user to the recruiters.


api/validators/UserControllerRequests.ts
```diff
@@ -29,6 +29,8 @@ export class UserPatches implements IUserPatches {
   @Allow()
   bio?: string;
 
+  resumeOptedIn?: boolean;
+
```
shravanhariharan2, line 33: When we say "opt-in", what are the users opting in to? Are they opting into recruiters viewing their resumes (i.e., only select accounts with RECRUITER types can see these, and opting out meaning no one can see their resume except themselves)? Or are they opting into their resumes being public? (i.e., all users can see their resumes, vs. opting out means only recruiters see the resumes)

shravanhariharan2: Also, let's rename to something like `isResumePublic` so that it's clearer what the resume is being opted into (depending on what it actually is being opted into)

models/UserModel.ts
```diff
@@ -103,6 +106,10 @@ export class UserModel extends BaseEntity {
     return this.accessType === UserAccessType.MERCH_STORE_DISTRIBUTOR;
   }
 
+  public isResumeOptedIn(): boolean {
+    return this.resumeOptedIn;
+  }
+
```
shravanhariharan2, line 112: Is there a reason this function is necessary? Can we just rename the field to be `isResumeOptedIn` or `isResumePublic` so that we can avoid defining this function?

types/ApiResponses.ts
```diff
@@ -301,6 +301,7 @@ export interface PublicProfile {
 export interface PrivateProfile extends PublicProfile {
   email: string,
   accessType: string,
+  resumeOptedIn: boolean,
   state: string,
   credits: number,
```
shravanhariharan2, line 306: Should we add this field to `PublicProfile` instead? `PublicProfile` is returned when a user views another user's profile, so depending on the behavior of "opting-in", we may or may not need to include this field in the public response, so that the frontend knows to display/not display the resume when another user views a profile

shravanhariharan2 (changes requested):
Good first PR! Mostly questions to clear up what "opting into resumes" actually means from a permissions level

dowhep:
Should we make the resume visibility a bit-flag int variable to allow the users to decide who is able to view the resume (recruiters, member public, and more) instead of having multiple long-named boolean variables? It might hinder the understandability of the code but I think if done well, it can help reduce the code count and increase performance without sacrificing too much clarity.

api/validators/UserControllerRequests.ts
dowhep: Opt-in in this case is referring to the ability of the recruiters to view the resume and consider the owner of the resume a potential recruit. This is necessary because there could be cases where a member puts their resume on the website without a need for the recruiters to see it, such as simply viewing their resume from other devices or already hired to a job.

models/UserModel.ts
dowhep: This function is not necessary and is removed.

types/ApiResponses.ts
dowhep: We don't actually want public users, in any case, to see whether an individual user is allowing the recruiters to see their resume as that is not necessary.

dowhep (commented):
Removed the unnecessary getter method in UserModel and clarified what Opted-In is (now changed to "resumeVisibleToRecruiter"

PR #316, opened 2022-08-11
Add social media links to user

meshachadoe:
Closes #[ISSUE-NUM].

### DIFF
--- a/migrations/0035-add-social-media-links.ts
+++ b/migrations/0035-add-social-media-links.ts
@@ -1,0 +1,39 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+const TABLE_NAME = 'Users';
+
+export class AddSocialMediaLinks1660118187298 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn(TABLE_NAME, new TableColumn({
+      name: 'linkedin',
+      type: 'varchar(255)',
+      isNullable: false,
+      default: '',
+    }));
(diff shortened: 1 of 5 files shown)


shravanhariharan2 (changes requested):
Good work! I think there might be a more efficient way to design our database schema to handle social media links though. There's two options that I see for supporting social media links:
1. Have a column in the Users table for each media link (what we have in this PR)
2. Create a `UserSocialMediaUrls` table with 4 columns: uuid, user, socialMediaType, and url.

I prefer option 2 for a few reasons. The first reason is that option 1 makes our database more sparse (i.e., lots of columns but not much data in those columns), which leads to extra storage needed and extra costs. Why is this the case? Well it's because of our use case - we have around 3000 users in our portal DB, and realistically, only a small percent of those users will have social media links attached to the portal. So for most of the users in our DB, these columns will be empty.

Additionally, every time we need to add a social media type (eg. lets say we add Github, or Twitter), then we'd need a new migration for each URL, which can add overhead to our setup.

With option 2, we use relationships to efficiently store URLs - a row only exists when a user adds a social media url. We can have `socialMediaType` be an enum of `(LINKEDIN, INSTAGRAM, FACEBOOK, PORTFOLIO)`, and any time we need to add a new type, we just modify the enum within the code, which is much more efficient since no database changes are necessary.

Feel free to follow up with any questions about the approach I described

meshachadoe:
Hey @shravanhariharan2, thanks for your suggestion! I also realized that what I was doing was ineffective but I struggled to come up with a better approach. This is really helpful and I'll work on implementing it.

PR #317, opened 2022-09-22
Implement user social media url functionality

meshachadoe:
Closes #[ISSUE-NUM].


api/validators/UserSocialMediaUrlControllerRequests.ts
```diff
@@ -0,0 +1,32 @@
...
+  UpdateUserSocialMediaUrlRequest as IUpdateUserSocialMediaUrlRequest,
+  SocialMediaUrl as ISocialMediaUrl,
+  SocialMediaType,
+} from '../../types';
+
+export class SocialMediaUrl implements ISocialMediaUrl {
+  @IsDefined()
+  @IsValidSocialMediaUrlType()
+  socialMediaType: SocialMediaType;
+
```
shravanhariharan2, line 15: nit: we can rename this field across the whole PR to just 'type', since it'd always be in the context of social media (e.g. either in the database table or in an object)

shravanhariharan2: nit: let's rename this object to `SocialMedia`

api/validators/UserSocialMediaUrlControllerRequests.ts
```diff
@@ -0,0 +1,32 @@
...
+  @IsNotEmpty()
+  url: string;
+}
+
+export class InsertSocialMediaUrlRequest implements IInsertUserSocialMediaUrlRequest {
+  @Type(() => SocialMediaUrl)
+  @ValidateNested()
+  @IsDefined()
+  socialMediaUrl: SocialMediaUrl;
+}
```
shravanhariharan2, line 26: nit: rename to `socialMedia`

migrations/0034-add-userSocialMediaUrl-table.ts
```diff
@@ -0,0 +1,48 @@
+import { MigrationInterface, QueryRunner, Table } from 'typeorm';
+
+export class AddUserSocialMediaUrlsTable1663760514239 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.createTable(new Table({
+      name: 'UserSocialMediaUrls',
+      columns: [
```
shravanhariharan2, line 7: nit: rename to 'UserSocialMedia'

migrations/0034-add-userSocialMediaUrl-table.ts
```diff
@@ -0,0 +1,48 @@
...
+          generationStrategy: 'uuid',
+        },
+        {
+          name: 'user',
+          type: 'uuid',
+        },
+        {
+          name: 'socialMediaType',
+          type: 'varchar(255)',
+        },
```
shravanhariharan2, line 22: nit: rename to 'type'

migrations/0034-add-userSocialMediaUrl-table.ts
```diff
@@ -0,0 +1,48 @@
...
+          referencedColumnNames: ['uuid'],
+          onDelete: 'CASCADE',
+        },
+      ],
+    }));
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.dropTable('UserSocialMediaUrls');
+  }
```
shravanhariharan2, line 47: nit: generally in migrations we have a `TABLE_NAME` constant that holds the table name, and we use that in the createTable and dropTable functions to avoid repetition

repositories/UserSocialMediaUrlRepository.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  public async isNewSocialMediaTypeForUser(user: UserModel, socialMediaType: SocialMediaType): Promise<boolean> {
+    const socialMediaUrls = await this.getSocialMediaUrlsForUser(user);
+    let isNewSocialMediaType = true;
+    socialMediaUrls.forEach((url) => {
+      if (url.socialMediaType === socialMediaType) isNewSocialMediaType = false;
+    });
+    return isNewSocialMediaType;
+  }
+
```
shravanhariharan2, line 31: One more efficient way to do this operation is to query the table where user = user parameter and socialMediaType = socialMediaType parameter, and return if the number of rows == 0

services/UserSocialMediaUrlService.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.transactions = new TransactionsManager(entityManager);
+  }
+
+  public async getSocialMediaUrlsForUser(user: UserModel): Promise<PublicUserSocialMediaUrl[]> {
+    const userSocialMediaUrl = await this.transactions.readOnly(async (txn) => {
+      const userSocialMediaUrlRepository = Repositories.userSocialMediaUrl(txn);
+      return userSocialMediaUrlRepository.getSocialMediaUrlsForUser(user);
+    });
+    return userSocialMediaUrl.map((fb) => fb.getPublicSocialMediaUrl());
+  }
```
shravanhariharan2, line 25: nit: rename 'fb' to 'sm' or something similar (fb refers to feedback here)

services/UserSocialMediaUrlService.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor(@InjectManager() entityManager: EntityManager) {
+    this.transactions = new TransactionsManager(entityManager);
+  }
+
+  public async getSocialMediaUrlsForUser(user: UserModel): Promise<PublicUserSocialMediaUrl[]> {
+    const userSocialMediaUrl = await this.transactions.readOnly(async (txn) => {
+      const userSocialMediaUrlRepository = Repositories.userSocialMediaUrl(txn);
+      return userSocialMediaUrlRepository.getSocialMediaUrlsForUser(user);
+    });
```
shravanhariharan2, line 23:
```suggestion
    const userSocialMediaUrl = await this.transactions.readOnly(async (txn) =>  
         Repositories.userSocialMediaUrl(txn).getSocialMediaUrlsForUser(user));
```

services/UserSocialMediaUrlService.ts
```diff
@@ -0,0 +1,49 @@
...
+    const addedSocialMediaUrl = await this.transactions.readWrite(async (txn) => {
+      await Repositories.userSocialMediaUrl(txn);
+      const isNewSocialMediaType = await Repositories.userSocialMediaUrl(txn)
+        .isNewSocialMediaTypeForUser(user, socialMediaUrl.socialMediaType);
+      if (!isNewSocialMediaType) throw new UserError('Social media URL of this type has been created for this user');
+      return Repositories.userSocialMediaUrl(txn)
+        .upsertSocialMediaUrl(UserSocialMediaUrlsModel.create({ ...socialMediaUrl, user }));
+    });
+    return addedSocialMediaUrl.getPublicSocialMediaUrl();
+  }
```
shravanhariharan2, line 37: nit: let's create the socialMediaRepository variable here since we're using it on more than one occasion

shravanhariharan2 (commented):
Great work! Lots of minor comments, let's add some tests to this code now to ensure our business logic is covered (e.g. making sure users can't create multiple social media rows of the same type, etc)

api/controllers/UserSocialMediaUrlController.ts
```diff
@@ -0,0 +1,44 @@
...
+import {
+  InsertSocialMediaUrlRequest,
+  UpdateSocialMediaUrlRequest,
+} from '../validators/UserSocialMediaUrlControllerRequests';
+
+@UseBefore(UserAuthentication)
+@JsonController('/socialMedia')
+export class UserSocialMediaUrlController {
+  private userSocialMediaUrlService: UserSocialMediaUrlService;
+
```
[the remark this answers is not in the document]
meshachadoe: I simplified it to just `/user/socialMedia` since I thought that only the user should be allowed to make a `POST`, `PATCH` and `DELETE` request to their own urls. There would not be a scenario where one user is trying to change another user's social media url. Since the `GET` request is now tied to the user, other users are still able to fetch another user's social media urls.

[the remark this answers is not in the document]
shravanhariharan2: Great point, this makes sense to me 👍🏾

migrations/0034-add-userSocialMedia-table.ts
```diff
@@ -0,0 +1,50 @@
+import { MigrationInterface, QueryRunner, Table } from 'typeorm';
+
+const TABLE_NAME = 'UserSocialMedias';
```
shravanhariharan2, line 3: nit: grammar with plurals (I just noticed this in this PR, let's have this table name be "UserSocialMedia" instead of "UserSocialMedias"
```suggestion
const TABLE_NAME = 'UserSocialMedia';
```

migrations/0034-add-userSocialMedia-table.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export class AddUserSocialMediasTable1663760514239 implements MigrationInterface {
```
shravanhariharan2, line 5:
```suggestion
export class AddUserSocialMediaTable1663760514239 implements MigrationInterface {
```

models/UserModel.ts
```diff
@@ -113,6 +117,7 @@ export class UserModel extends BaseEntity {
       major: this.major,
       bio: this.bio,
       points: this.points,
+      userSocialMedia: this.userSocialMedia,
     };
```
shravanhariharan2, line 121: This field here should be optional, since the join of Users <-> UserSocialMedia might not always happen (since the relationship is not eager - we're only joining on `getPublicProfile()` or `getFullUserProfile()`.

I left a similar comment on the Resumes PR which explains this further and how to make this field optional: https://github.com/acmucsd/membership-portal/pull/319#discussion_r1023404624

models/UserModel.ts
```diff
@@ -130,6 +135,7 @@ export class UserModel extends BaseEntity {
       bio: this.bio,
       points: this.points,
       credits: this.credits,
+      userSocialMedia: this.userSocialMedia,
```
shravanhariharan2, line 138: Same as above

models/UserSocialMediaModel.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Entity('UserSocialMedias')
+export class UserSocialMediaModel extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  uuid: Uuid;
+
+  @ManyToOne((type) => UserModel, (user) => user.feedback, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'user' })
+  @Index({ unique: true })
+  user: UserModel;
```
shravanhariharan2, line 13: Should this column have a unique constraint? I think if a user has multiple social media links on their profile, then there would be more than one `UserSocialMediaModel`s with the same user, which should be legal according to our business logic.

models/UserSocialMediaModel.ts
```diff
@@ -0,0 +1,29 @@
...
+  url: string;
+
+  public getPublicSocialMedia(): PublicUserSocialMedia {
+    return {
+      uuid: this.uuid,
+      user: this.user,
+      type: this.type,
+      url: this.url,
+    };
+  }
```
shravanhariharan2, line 28: Same comment as above with making `user` here optional. Since we aren't always joining UserSocialMedia <-> Users, then we need to make this field optional and set it only if it exists. Also, let's have `user` be of type `PublicProfile` rather than `UserModel`, since this is a public entity method (so call `getPublicProfile()`)

repositories/UserSocialMediaRepository.ts
```diff
@@ -0,0 +1,36 @@
...
+import { SocialMediaType, Uuid } from 'types';
+import { UserSocialMediaModel } from '../models/UserSocialMediaModel';
+import { UserModel } from '../models/UserModel';
+import { BaseRepository } from './BaseRepository';
+
+@EntityRepository(UserSocialMediaModel)
+export class UserSocialMediaRepository extends BaseRepository<UserSocialMediaModel> {
+  public async getSocialMediasForUser(user: UserModel): Promise<UserSocialMediaModel[]> {
+    return this.getBaseFindQuery().where({ user }).getMany();
+  }
```
shravanhariharan2, line 11: I'm not sure this method needs to do the JOIN on Users. Since this would only be called if the user is already provided (e.g. in the parameter), then the joins may be redundant since we already have the user data. E.g., in `getFullUserProfile()` we are calling this method and attaching it to a `user` object, so `user.socialMedias[0].user` would be a needless cycle.

services/UserSocialMediaService.ts
```diff
@@ -0,0 +1,55 @@
...
+      const userSocialMediaRepository = Repositories.userSocialMedia(txn);
+      const isNewSocialMediaType = await userSocialMediaRepository.isNewSocialMediaTypeForUser(user, socialMedia.type);
+      if (!isNewSocialMediaType) throw new UserError('Social media URL of this type has been created for this user');
+      return userSocialMediaRepository.upsertSocialMedia(UserSocialMediaModel.create({ ...socialMedia, user }));
+    });
+    return addedSocialMedia.getPublicSocialMedia();
+  }
+
+  public async updateSocialMediaByUuid(uuid: Uuid, url: string): Promise<PublicUserSocialMedia> {
+    const updatedSocialMedia = await this.transactions.readWrite(async (txn) => {
```
shravanhariharan2, line 36: Referring to the comment on passing in the `user` as a parameter, let's do the authentication check here on whether or not the user is allowed to update this specific social media.

services/UserSocialMediaService.ts
```diff
@@ -0,0 +1,55 @@
...
+      const userSocialMediaRepository = Repositories.userSocialMedia(txn);
+      const socialMedia = await userSocialMediaRepository.findByUuid(uuid);
+      if (!socialMedia) throw new NotFoundError('Social media URL not found');
+
+      return userSocialMediaRepository.upsertSocialMedia(socialMedia, { url });
+    });
+    return updatedSocialMedia.getPublicSocialMedia();
+  }
+
+  public async deleteSocialMediaByUuid(uuid: Uuid): Promise<PublicUserSocialMedia> {
```
shravanhariharan2, line 46: Same comment as above

types/Enums.ts
```diff
@@ -81,3 +81,10 @@ export enum OrderPickupEventStatus {
   CANCELLED = 'CANCELLED',
   COMPLETED = 'COMPLETED',
 }
+
+export enum SocialMediaType {
+  LINKEDIN = 'LINKEDIN',
+  INSTAGRAM = 'INSTAGRAM',
+  FACEBOOK = 'FACEBOOK',
+  PORTFOLIO = 'PORTFOLIO',
+}
```
shravanhariharan2, line 90: Have we confirmed with membership/faris/etc. that these are the 4 social media types we are starting with?

shravanhariharan2 (changes requested):
Nice progress, there's a few breaking things we need to check still like authentication with the POST and PATCH routes, but otherwise business logic makes sense.

Let's also add some simple tests to make sure all the inserts/updates and such are happening properly, and also that we can get the social media urls of a user with `getUser()` or `getCurrentUser()`

services/UserSocialMediaService.ts
```diff
@@ -0,0 +1,58 @@
...
+    const userSocialMedia = await this.transactions.readOnly(async (txn) => Repositories.userSocialMedia(txn)
+      .getSocialMediaForUser(user));
+    return userSocialMedia.map((sm) => sm.getPublicSocialMedia());
+  }
+
+  public async insertSocialMediaForUser(user: UserModel, socialMedia: SocialMedia) {
+    const addedSocialMedia = await this.transactions.readWrite(async (txn) => {
+      const userSocialMediaRepository = Repositories.userSocialMedia(txn);
+      const isNewSocialMediaType = await userSocialMediaRepository.isNewSocialMediaTypeForUser(user, socialMedia.type);
+      if (!isNewSocialMediaType) throw new UserError('Social media URL of this type has been created for this user');
```
shravanhariharan2, line 29: nit: Let's add an "already" here so that it explains why this case is an error (since otherwise, the error message here makes it sound like the social media entry has been created, even though an error is thrown)
```suggestion
      if (!isNewSocialMediaType) throw new UserError('Social media URL of this type has already been created for this user');
```

services/UserSocialMediaService.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public async updateSocialMediaByUuid(user: UserModel,
+    uuid: Uuid,
+    changes: Partial<UserSocialMediaModel>): Promise<PublicUserSocialMedia> {
+    const updatedSocialMedia = await this.transactions.readWrite(async (txn) => {
+      const userSocialMediaRepository = Repositories.userSocialMedia(txn);
+      const socialMedia = await userSocialMediaRepository.findByUuid(uuid);
+      if (!socialMedia) throw new NotFoundError('Social media URL not found');
+      if (user.uuid !== socialMedia.user.uuid) throw new ForbiddenError('User does not own social media entity');
```
shravanhariharan2, line 42: nit: Let's change the error message to something like
```suggestion
      if (user.uuid !== socialMedia.user.uuid) throw new ForbiddenError('User cannot update a social media URL of another user');
```
since the caller does not need to know details about entities or who owns what entity, as entities are more of an internal backend-specific concept.

services/UserSocialMediaService.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+    return updatedSocialMedia.getPublicSocialMedia();
+  }
+
+  public async deleteSocialMediaByUuid(user: UserModel, uuid: Uuid): Promise<PublicUserSocialMedia> {
+    const updatedSocialMedia = await this.transactions.readWrite(async (txn) => {
+      const userSocialMediaRepository = Repositories.userSocialMedia(txn);
+      const socialMedia = await userSocialMediaRepository.findByUuid(uuid);
+      if (!socialMedia) throw new NotFoundError('Social media URL not found');
+      if (user.uuid !== socialMedia.user.uuid) throw new ForbiddenError('User does not own social media entity');
```
shravanhariharan2, line 53: nit: same as above

types/ApiRequests.ts
```diff
@@ -74,6 +79,14 @@ export interface UpdateFeedbackStatusRequest {
   status: FeedbackStatus;
 }
 
+export interface InsertUserSocialMediaRequest {
+  socialMedia: SocialMedia;
+}
+
+export interface UpdateUserSocialMediaRequest {
+  socialMedia: Partial<SocialMedia>;
+}
```
shravanhariharan2, line 88: I think this should be a custom `SocialMediaPatches` type rather than `Partial<SocialMedia>`, since using `SocialMedia` as a partial means the `type` property is also allowed, which it shouldn't be, since users shouldn't be able to update the social media type of their URLs.

services/UserSocialMediaService.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  constructor(@InjectManager() entityManager: EntityManager) {
+    this.transactions = new TransactionsManager(entityManager);
+  }
+
+  public async getSocialMediaForUser(user: UserModel): Promise<PublicUserSocialMedia[]> {
+    const userSocialMedia = await this.transactions.readOnly(async (txn) => Repositories.userSocialMedia(txn)
+      .getSocialMediaForUser(user));
+    return userSocialMedia.map((sm) => sm.getPublicSocialMedia());
+  }
```
shravanhariharan2, line 23: nit: Instead of returning the `Public` type in the service, let's return the model directly, and map it to its `Public` type in the API layer. This pattern is more ideal since the `Public` type is a concept specific to the API layer (since it defines the structure of the responses of the API to the client), so it should be used only in the API layer. We do still have remnants of services returning the `Public` type in other places in the code, but it's something we should move away from and eventually remove from the service layer entirely in a separate refactor.

Let's make this change for each of the functions in this service and their corresponding API layer equivalents.

shravanhariharan2 (commented):
Great progress, again mostly nitpicks on things like error messaging and typing. Will give a review on tests once they are ready to go.

tests/userSocialMedia.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    await new PortalState()
+      .createUsers(member)
+      .write();
+
+    const userController = ControllerFactory.user(conn);
+
+    await userController.insertSocialMediaForUser({ socialMedia: userSocialMedia }, member);
+    const userResponse = await userController.getUser({ uuid: member.uuid }, member);
+
```
shravanhariharan2, line 34: In this block, to limit the number of controller calls to 1, let's use a `conn.manager.findOne()` call instead of the `getUser` call here, and maybe have 1 small unit test where we just check the `getUser()` route functionality. This would be the first time in the whole codebase that we're doing tests for specifically GET routes, but it's a convention we should set for future precedence to keep the 1-route-per-test standard.

tests/userSocialMedia.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const userSocialMedia = UserSocialMediaFactory.fake();
+
+    await new PortalState()
+      .createUsers(member)
+      .write();
+    const unauthorizedMember = UserFactory.fake();
+    await new PortalState()
+      .createUsers(unauthorizedMember)
+      .write();
+
```
shravanhariharan2, line 76: Here we don't actually need two separate PortalState calls, we can do both in 1 call:
```ts
    const member = UserFactory.fake();
    const unauthorizedMember = UserFactory.fake();
    const userSocialMedia = UserSocialMediaFactory.fake();

    await new PortalState()
      .createUsers(member, unauthorizedMember)
      .write();
```

tests/userSocialMedia.test.ts
```diff
@@ -0,0 +1,116 @@
...
```
shravanhariharan2, line 104: Same as above

tests/userSocialMedia.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    await userController.insertSocialMediaForUser({ socialMedia: userSocialMedia }, member);
+    const userResponse = await userController.getUser({ uuid: member.uuid }, member);
+
+    const errorMessage = 'User cannot update a social media URL of another user';
+
+    await expect(userController.updateSocialMediaForUser(
+      { uuid: userResponse.user.userSocialMedia[0].uuid },
+      { socialMedia: { url: faker.internet.url() } },
+      unauthorizedMember,
+    )).rejects.toThrow(errorMessage);
```
shravanhariharan2, line 87: nit: Let's avoid inlining the object creation here and keep it in a separate variable to improve the readability of this call

tests/userSocialMedia.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const userController = ControllerFactory.user(conn);
+    await userController.insertSocialMediaForUser({ socialMedia: userSocialMedia }, member);
+    const userResponse = await userController.getUser({ uuid: member.uuid }, member);
+
+    const errorMessage = 'User cannot delete a social media URL of another user';
+
+    await expect(userController.deleteSocialMediaForUser(
+      { uuid: userResponse.user.userSocialMedia[0].uuid },
+      unauthorizedMember,
+    )).rejects.toThrow(errorMessage);
```
shravanhariharan2, line 114: nit: same as above with inlining

shravanhariharan2 (commented):
Few key comments on ideal testing design, the cases themselves look great though. Let me know if there are any questions about the proposed design.

tests/data/PortalState.ts
```diff
@@ -114,6 +119,11 @@ export class PortalState {
     return this;
   }
 
+  public createUserSocialMedia(...userSocialMediaList: UserSocialMediaModel[]): PortalState {
+    this.userSocialMediaList = this.userSocialMediaList.concat(userSocialMediaList);
+    return this;
+  }
+
```
shravanhariharan2, line 126: The pattern we follow in `PortalState` with inserting data that is related to users is to specify the user in the first parameter, and then the list of social media after (e.g. [submitFeedback](https://github.com/acmucsd/membership-portal/blob/master/tests/data/PortalState.ts#L192), orderMerch, attendEvents, etc.). That way, you don't need to pass in the user model into the social media factory, and can instead just pass in the user model into this function and handle the relationship here.

tests/userSocialMedia.test.ts
```diff
@@ -0,0 +1,121 @@
...
+afterAll(async () => {
+  await DatabaseConnection.clear();
+  await DatabaseConnection.close();
+});
+
+describe('social media URL submission', () => {
+  test('properly persists on successful submission', async () => {
+    const conn = await DatabaseConnection.get();
+    let member = UserFactory.fake();
+    const userSocialMedia = UserSocialMediaFactory.fake({ user: member, type: SocialMediaType.FACEBOOK });
```
shravanhariharan2, line 25: nit: the type of social media doesn't matter here so we can omit specifying the type here and just let the factory randomly generate a type for us. I think the type only matters for the test below this one where we verify that two social medias of the same type can't be created, so for the other tests, let's also randomize the types there.

tests/userSocialMedia.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const errorMessage = 'User cannot update a social media URL of another user';
+    const uuidParams = { uuid: member.userSocialMedia[0].uuid };
+    const socialMediaParams = { socialMedia: { url: faker.internet.url() } };
+    await expect(userController.updateSocialMediaForUser(uuidParams, socialMediaParams, unauthorizedMember))
+      .rejects.toThrow(errorMessage);
+  });
+
+  test('is invalidated when invalid uuid is passed in', async () => {
+    const conn = await DatabaseConnection.get();
```
shravanhariharan2, line 85: nit: 'invalid uuid' here can mean a few different things (e.g. an improperly formatted one, a uuid v2 vs. a v4, etc.) so let's be more specific here and with the variable naming on what exactly we are trying to test. It seems like we are testing a uuid that is valid but that is not found, so we can rename this test to 'is invalidated when the entity for a uuid is not found', and rename `uuidParams` to something like `missingEntityUuid`

shravanhariharan2 (approved):
Great work! Code and tests seem to functionally work, I left a few small comments for code quality and matching our existing patterns, but you are good to merge once those comments are addressed. Thanks for your patience throughout this PR 👍🏾

PR #319, opened 2022-09-23
Allow admins to get all visible resumes

dowhep:
Create API routes for getting all opted in/visible resumes


repositories/ResumeRepository.ts
```diff
@@ -1,8 +1,22 @@
 import { ResumeModel } from 'models/ResumeModel';
```
shanekim28, line 1: I've found that VSCode's auto-importer fails when using absolute routes. Make sure you use the relative route like `../models/ResumeModel` whenever you're importing.

models/UserModel.ts
```diff
@@ -134,6 +134,7 @@ export class UserModel extends BaseEntity {
       bio: this.bio,
       points: this.points,
       credits: this.credits,
+      resumes: this.resumes,
```
shanekim28, line 137: Do we need this? Since all resumes should be handled through the controller, I feel like we don't need a field for this in the UserModel. The "Resumes" table and controller should handle all of the Resume viewing/updating for a single user

api/controllers/ResumeController.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(resumeService: ResumeService) {
+    this.resumeService = resumeService;
+  }
+
+  @Get()
+  async getAllVisibleResumes(@Params() params: UuidParam,
+  @AuthenticatedUser() currentUser: UserModel): Promise<GetResumesListResponse> {
+    const resumes = await this.resumeService.getVisibleResumes();
+    return { error: null, resumes };
+  }
```
shravanhariharan2, line 24: I think we should have some kind of permissions check here. e.g., only admins should be able to get all visible resumes, not any member.

repositories/ResumeRepository.ts
```diff
@@ -1,8 +1,22 @@
...
 import { EntityRepository } from 'typeorm';
 import { BaseRepository } from './BaseRepository';
+import { Uuid } from '../types';
+import { UserModel } from '../models/UserModel';
 
 @EntityRepository(ResumeModel)
 export class ResumeRepository extends BaseRepository<ResumeModel> {
+  public async findVisible(): Promise<ResumeModel[]> {
+    return this.repository.find({
```
shravanhariharan2, line 10: nit: let's name it something like `findVisibleResumes()`

repositories/ResumeRepository.ts
```diff
@@ -1,8 +1,22 @@
...
+    });
+  }
 
+  public async findByUuid(uuid: Uuid): Promise<ResumeModel> {
+    return this.repository.findOne({ uuid });
+  }
+
+  public async findByUser(user: UserModel): Promise<ResumeModel[]> {
+    return this.repository.find({ user });
+  }
```
shravanhariharan2, line 21: nit: let's rename to `findAllByUser` since this would return an array as opposed to a single object

services/ResumeService.ts
```diff
@@ -0,0 +1,23 @@
...
+  private transactions: TransactionsManager;
+
+  constructor(@InjectManager() entityManager: EntityManager) {
+    this.transactions = new TransactionsManager(entityManager);
+  }
+
+  public async getVisibleResumes() : Promise<ResumeModel[]> {
+    const resumes = await this.transactions.readOnly(async (txn) => Repositories
+    .resume(txn).findVisible());
+    if (!resumes) throw new NotFoundError('Resumes was not found');
```
shravanhariharan2, line 20: nit: formatting - line 19 should be tabbed inwards

shravanhariharan2: also, do we need to throw an error if no resumes were found? I imagine just returning an empty array should suffice, and thus no resumes would be displayed on the frontend.

tests/controllers/ControllerFactory.ts
```diff
@@ -15,6 +15,8 @@ import UserAuthService from '../../services/UserAuthService';
 import EmailService from '../../services/EmailService';
 import EventService from '../../services/EventService';
 import MerchStoreService from '../../services/MerchStoreService';
+import { ResumeController } from 'api/controllers/ResumeController';
+import ResumeService from 'services/ResumeService';
```
shravanhariharan2, line 19: nit: this should be a relative import

models/UserModel.ts
shravanhariharan2: I think this is fine - getting a user's own profile should return all the info regarding that user, which includes their resume(s). Resumes are tightly related to a user's profile (since that's one of the only places they're displayed), so I don't particularly see an issue with having them returned in the same API call.

Also, @dowhep I don't believe this code works here because we aren't [eager-loading](https://typeorm.io/eager-and-lazy-relations) resumes. This means that this field will be undefined every time we make this function call. Instead, we need to query the Resumes table for all resumes for that user, and attach it to this field. Then, we can return the fields as you specified.

shravanhariharan2 (changes requested):
Nice work - let's add some quick tests for asserting that getting all visible resumes returns only visible resumes, and that getting a user's full profile returns the resumes properly.

services/ResumeService.ts
dowhep: So far I think an empty array should suffice

services/ResumeService.ts
```diff
@@ -0,0 +1,24 @@
+import { ResumeModel } from 'models/ResumeModel';
+import { NotFoundError } from 'routing-controllers';
+import { Service } from 'typedi';
+import { EntityManager } from 'typeorm';
+import { InjectManager } from 'typeorm-typedi-extensions';
+import Repositories, { TransactionsManager } from '../repositories';
+// import { ResumeRepository } from '../repositories/ResumeRepository';
```
michl1001, line 7: nit: is there a reason we need this import?

tests/resume.test.ts
```diff
@@ -0,0 +1,56 @@
+import { ActivityScope, ActivityType, SubmitAttendanceForUsersRequest, UserAccessType } from '../types';
+import { ControllerFactory } from './controllers';
+import { DatabaseConnection, EventFactory, PortalState, UserFactory } from './data';
+
```
michl1001, line 4: Might be worth handling other common error cases? (request a user resume that isn't uploaded yet, what happens with reuploads)

repositories/ResumeRepository.ts
```diff
@@ -1,8 +1,22 @@
...
+import { ResumeModel } from '../models/ResumeModel';
 import { BaseRepository } from './BaseRepository';
+import { Uuid } from '../types';
+import { UserModel } from '../models/UserModel';
 
 @EntityRepository(ResumeModel)
 export class ResumeRepository extends BaseRepository<ResumeModel> {
+  public async findVisibleResumes(): Promise<ResumeModel[]> {
+    return this.repository.find({
+      isResumeVisible: true,
```
michl1001, line 11: Should we filter duplicates from one person?

michl1001 (commented):
Looks good for general logic - though had a few nitpick and questions

services/ResumeService.ts
shravanhariharan2: Yeah let's definitely remove any commented-out code

api/controllers/ResumeController.ts
```diff
@@ -0,0 +1,25 @@
...
+    this.resumeService = resumeService;
+  }
+
+  @Get()
+  async getAllVisibleResumes(@Params() params: UuidParam,
+    @AuthenticatedUser() currentUser: UserModel): Promise<GetResumesListResponse> {
+    const resumes = await this.resumeService.getVisibleResumes(PermissionsService.canSeeAllVisibleResumes(currentUser));
+    return { error: null, resumes };
+  }
+}
```
shravanhariharan2, line 25: We should throw if someone does not have permission to see all resumes (and if someone doesn't have permission to view stuff in general). If someone who doesn't have permission is calling this API route, this is fundamentally an error by the caller, and is something that we should throw a 401 Forbidden Error for. Rerturning an empty response makes it seem like they _do_ have permission, but that there just aren't any resumes, which is incorrect from a data point of view.

repositories/ResumeRepository.ts
shravanhariharan2: At some point we may support users having multiple resumes, and we also have the guarantee that there'll only be one resume per-person right now due to our resume deletion logic, so I think it's okay to not do any duplicate filtering here (also to keep API types consistent so that we wouldn't have to switch from returning Resume to returning Resume[] if we do need to support multiple resumes.

services/ResumeService.ts
```diff
@@ -0,0 +1,24 @@
...
+export default class ResumeService {
+  private transactions: TransactionsManager;
+
+  constructor(@InjectManager() entityManager: EntityManager) {
+    this.transactions = new TransactionsManager(entityManager);
+  }
+
+  public async getVisibleResumes(canSeeAllVisibleResumes = false) : Promise<ResumeModel[]> {
+    if (!canSeeAllVisibleResumes) return [];
+    const resumes = await this.transactions.readOnly(async (txn) => Repositories
```
shravanhariharan2, line 19: Related to my first comment, let's throw on permission error instead of returning an empty response

tests/resume.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    //   expect(attendanceResponse.attendances).toHaveLength(1);
+    //   expect(attendanceResponse.attendances[0].event).toStrictEqual(event.getPublicEvent());
+
+    //   const activityResponse = await userController.getUserActivityStream({ uuid: user.uuid }, admin);
+
+    //   expect(activityResponse.activity).toHaveLength(2);
+    //   expect(activityResponse.activity[1].pointsEarned).toEqual(event.pointValue);
+    //   expect(activityResponse.activity[1].type).toEqual(ActivityType.ATTEND_EVENT);
+    //   expect(activityResponse.activity[1].scope).toEqual(ActivityScope.PUBLIC);
+    // }
```
shravanhariharan2, line 53: Was there a reason this is commented out?

models/UserModel.ts
```diff
@@ -121,6 +122,7 @@ export class UserModel extends BaseEntity {
   }
 
   public getFullUserProfile(): PrivateProfile {
+    const resumes = await ResumeRepository.findAllByUser(this); // TODO: somehow get to userresume
     return {
```
shravanhariharan2, line 126: Here's my idea: let's not perform the resume lookup in `getFullUserProfile()`, and instead let's create a new function `async getResumes()` that does the lookup and populates the `user.resumes` field. Then, in the getCurrentUser() route, we can call this function and then return `user.getFullUserProfile()` in the API response.

The logic behind this is that there's only one route that would ever *need* to return resumes, which is the getCurrentUser() route. All the other calls to getFullUserProfile() don't actually need the resumes to be populated, which is why I'm hesitant to add an extra query to that function.

Let's still return `resumes: user.resumes` in the getFullUserProfile() return object so that the field is not undefined, but I don't necessarily think it needs to be populated, and if the caller needs to populate it, they can call `getResumes()` before it. Let's also leave a comment on the `getFullUserProfile()` function mentioning this.

Feel free to follow up if there's any questions with the approach

models/UserModel.ts
```diff
@@ -134,6 +136,7 @@ export class UserModel extends BaseEntity {
       bio: this.bio,
       points: this.points,
       credits: this.credits,
+      resumes,
     };
```
shravanhariharan2, line 140: nit: this would become `resumes: this.resumes`

shravanhariharan2 (commented):
Good progress, let me know if there's questions regarding my or Michael's comments 👍🏾

services/ResumeService.ts
dowhep: Was using it at some point and thought i needed it later, gonna remove it now

tests/resume.test.ts
dowhep: good idea, will look into that

api/controllers/ResumeController.ts
dowhep: Would you recommend handling the permission error in the controller in this case?

tests/resume.test.ts
dowhep: It was for reference for test cases, but I wanted to make sure the resume fetching works in getFullUserProfile before I started with the test cases

api/controllers/ResumeController.ts
shravanhariharan2: Yes, since permissions are related to whether or not we should process a request, permission logic should happen in the controller

models/UserModel.ts
dowhep: Is there a case where calling `getFullUserProfile` actually does not want to have the resume field populated even though it was populated with the new `getResume` function?

api/controllers/ResumeController.ts
```diff
@@ -1,11 +1,14 @@
...
+} from 'routing-controllers';
 import * as path from 'path';
+import PermissionsService from '../../services/PermissionsService';
 import StorageService from '../../services/StorageService';
 import { UserAuthentication } from '../middleware/UserAuthentication';
 import ResumeService from '../../services/ResumeService';
 import { UserModel } from '../../models/UserModel';
 import { AuthenticatedUser } from '../decorators/AuthenticatedUser';
-import { File, MediaType, UpdateResumeResponse } from '../../types';
+import { File, MediaType, GetResumesListResponse, UpdateResumeResponse } from '../../types';
```
shravanhariharan2, line 12: nit: let's call this `GetVisibleResumesResponse` to match the API name

models/ResumeModel.ts
```diff
@@ -13,7 +13,7 @@ export class ResumeModel extends BaseEntity {
   user: UserModel;
 
   @Column('boolean', { default: true, nullable: false })
-  isResumeVisible: boolean;
+  isResumeVisible: boolean = true;
```
shravanhariharan2, line 17: I think this default should be false, since the client should actually specify if it wants to share the resume publicly before doing so. That way, users wouldn't have their resumes publicly visible unless they specify (which they'd do during the upload of their resume, so it's not actually too much work on their end even if we set the default to false)

services/UserAccountService.ts
```diff
@@ -145,4 +146,13 @@ export default class UserAccountService {
       .user(txn)
       .getAllEmails());
   }
+
+  // this differs from calling UserModel directly bc it fills the resume field as well
+  public async getFullUserProfile(user: UserModel) : Promise<PrivateProfile> {
+    return this.transactions.readOnly(async (txn) => {
+      const userProfile = user.getFullUserProfile();
+      userProfile.resumes = await Repositories.resume(txn).findAllByUser(user);
+      return userProfile;
```
shravanhariharan2, line 155: Lovely

api/controllers/UserController.ts
```diff
@@ -25,9 +26,12 @@ export class UserController {
...
   private storageService: StorageService;
 
-  constructor(userAccountService: UserAccountService, storageService: StorageService) {
+  private resumeService: ResumeService;
+
+  constructor(userAccountService: UserAccountService, storageService: StorageService, resumeService: ResumeService) {
     this.userAccountService = userAccountService;
     this.storageService = storageService;
+    this.resumeService = resumeService;
   }
```
shravanhariharan2, line 35: nit: I think resumeService is unused in this controller, so let's remove it

tests/resume.test.ts
```diff
@@ -21,6 +21,35 @@ afterAll(async () => {
...
+      .createUsers(admin, member)
+      .write();
+
+    const resume = FileFactory.pdf(Config.file.MAX_RESUME_FILE_SIZE / 2);
+    const fileLocation = 'fake location';
+
+    const storageService = Mocks.storage(fileLocation);
+    const resumeController = ControllerFactory.resume(conn, instance(storageService));
+    const uploadedResume = (await resumeController.updateResume(resume, member)).resume;
+    delete uploadedResume.user;
```
shravanhariharan2, line 40: This resume upload operation can be moved into PortalState. The reason we have PortalState is to make it easy to persist fake data into the database.

Here, we are creating mocks, controllers, etc. to upload a resume but all we care about is that a resume is in the database - we've already validated that the 'upload resume' controller works as intended. So instead, let's modify PortalState to have a `createResumes` function that takes in a user and a list of resumes, and uploads resumes for that user. Note that we'd also need a `ResumeFactory` to create fake resumes.

The corresponding code should look something like this:
```ts
    const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
    const member = UserFactory.fake();
    const resume = ResumeFactory.fake({ userId: member.id });

    await new PortalState()
      .createUsers(admin, member)
      .createResumes(member, resume)
      .write();
    
    // throws permissions error for member
    expect(resumeController.getAllVisibleResumes(member))
      .rejects.toThrow(ForbiddenError);

    // get response resumes
    const response = await resumeController.getAllVisibleResumes(admin);
    expect(response.error).toBeNull();
    expect(response.resumes).toContainEqual(uploadedResume);
```

Let me know if there's any questions with what I mentioned!

shravanhariharan2 (commented):
Almost there! A few comments mostly pertaining to the tests, implementation looks great.

Also, can we delete the package-lock.json? Looks like `npm` was ran somewhere and the lockfile was committed

api/controllers/ResumeController.ts
dowhep: edited

models/ResumeModel.ts
```diff
@@ -13,7 +13,7 @@ export class ResumeModel extends BaseEntity {
   user: UserModel;
 
   @Column('boolean', { default: true, nullable: false })
-  isResumeVisible: boolean;
+  isResumeVisible: boolean = false;
```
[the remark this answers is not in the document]
shravanhariharan2: Also, this is not a change in your PR, but in `ResumeModel::getPublicResume()` we need to return the entire user, and not just their uuid. You can see in the figma below that we need to return some user data to the frontend (in this case, just first and last names):
image

Can we add this change into this PR? Instead of returning just `user: this.user.uuid`, let's return the user as a `PublicUser`. This means we also need to do a join when querying all visible resumes. Let's also test that we're returning the proper user object in our tests as well (just making sure the object is returned instead of just a uuid)

tests/resume.test.ts
```diff
@@ -23,6 +23,35 @@ afterAll(async () => {
...
+      .rejects.toThrow(ForbiddenError);
+
+    // get response resumes
+    const expected = resume;
+    delete expected.user;
+    const response = await resumeController.getAllVisibleResumes(admin);
+    expect(response.error).toBeNull();
+    expect(response.resumes).toContainEqual(expected);
+    expect(response.resumes).toHaveLength(1); // only the visible resume is shown
+  });
```
shravanhariharan2, line 52: Instead of calling `delete` on objects to strip down the fields we don't want to test, let's instead make it clear the fields we do want to test. This follows the testing convention we have, and also doesn't add confusion on why we are calling `delete` on some seemingly arbitrary field. We can do something like:
```ts
const response = await resumeController.getAllVisibleResumes(admin);
expect(response).toStrictEqual({
  uuid: resume.uuid,
  user: member.uuid,
  isResumeVisible: resume.isResumeVisible
  ... // fill out for rest of fields
})
```

types/ApiResponses.ts
```diff
@@ -319,6 +321,10 @@ export interface GetUserActivityStreamResponse extends ApiResponse {
   activity: PublicActivity[];
 }
 
+export interface GetVisibleResumesResponse extends ApiResponse {
+  resumes: ResumeModel[];
+}
```
shravanhariharan2, line 326: nit: this would need to be `PublicResume[]` after we change our API to return public resumes

api/controllers/ResumeController.ts
```diff
@@ -54,4 +57,11 @@ export class ResumeController {
     const resume = await this.resumeService.patchResume(params.uuid, patchResumeRequest.resume, user);
     return { error: null, resume: resume.getPublicResume() };
   }
+
+  @Get()
+  async getAllVisibleResumes(@AuthenticatedUser() user: UserModel): Promise<GetVisibleResumesResponse> {
+    if (!PermissionsService.canSeeAllVisibleResumes(user)) throw new ForbiddenError();
+    const resumes = (await this.resumeService.getVisibleResumes()).map((resume) => resume.getPublicResume());
+    return { error: null, resumes };
```
shravanhariharan2, line 65: nit: let's move the map operation to the return statement for easier readibility:
```suggestion
    const resumes = await this.resumeService.getVisibleResumes();
    return { error: null, resumes: resumes.map((resume) => resume.getPublicResume()) };
```

repositories/ResumeRepository.ts
```diff
@@ -1,9 +1,20 @@
 import { EntityRepository } from 'typeorm';
 import { ResumeModel } from '../models/ResumeModel';
 import { BaseRepository } from './BaseRepository';
+import { UserModel } from '../models/UserModel';
 
 @EntityRepository(ResumeModel)
 export class ResumeRepository extends BaseRepository<ResumeModel> {
+  public async findVisibleResumes(): Promise<ResumeModel[]> {
+    return this.repository.find({
+      isResumeVisible: true,
```
shravanhariharan2, line 10: Referring to the eager loading comment earlier, let's do the user join here

tests/resume.test.ts
```diff
@@ -23,42 +23,103 @@ afterAll(async () => {
...
+    expect(response.resumes[0]).toStrictEqual({
+      uuid: resume.uuid,
+      user: member.getPublicProfile(),
+      isResumeVisible: resume.isResumeVisible,
+      url: resume.url,
+      lastUpdated: resume.lastUpdated,
+    });
+  });
+
+  test('admins cannot get invisible resumes', async () => {
```
shravanhariharan2, line 58: nit: vocabulary
```suggestion

  test('admins cannot get hidden resumes', async () => {
```

tests/resume.test.ts
```diff
@@ -23,42 +23,103 @@ afterAll(async () => {
...
+    expect(response.error).toBeNull();
+    expect(response.resumes).toHaveLength(0);
+  });
+});
+
+describe('upload resume', () => {
+  test('authenticated user can upload resume', async () => {
+    const conn = await DatabaseConnection.get();
+    const member = UserFactory.fake();
+    await new PortalState().createUsers(member).write();
```
shravanhariharan2, line 80: nit: formatting
```suggestion
    await new PortalState()
      .createUsers(member)
      .write();
```

types/ApiResponses.ts
```diff
@@ -1,3 +1,4 @@
+import { ResumeModel } from 'models/ResumeModel';
```
shravanhariharan2, line 1: nit: this needs to be a relative URL instead of absolute, since our build system doesn't properly resolve absolute paths

types/ApiResponses.ts
```diff
@@ -303,6 +304,7 @@ export interface PrivateProfile extends PublicProfile {
   accessType: string,
   state: string,
   credits: number,
+  resumes: ResumeModel[],
```
shravanhariharan2, line 307: nit: this should also be `PublicResume[]`, since in almost every case we don't want to return a model object to the client since it might display fields that we didn't intend on showing (e.g. `id`, `UserModel`, etc). Even though this is a "private profile", we still return the "public" type because the term "public" in our types refers to anything the client interacts with

types/ApiResponses.ts
```diff
@@ -390,7 +396,7 @@ export interface CancelAllPendingOrdersResponse extends ApiResponse {}
 
 export interface PublicResume {
   uuid: Uuid;
-  user: Uuid;
+  user: PublicProfile;
```
shravanhariharan2, line 399: nit: let's make this optional since not all `PublicResumes` need to return a `user` object (e.g. in the `getFullUserProfile()` case, resumes are queried by the user to populate `user.resumes`, so we don't need `user.resumes.user`.

shravanhariharan2 (commented):
Mostly small comments, holding off on approval still since there's 1-2 breaking changes that need to be addressed, but should be good to go after changes are addressed

dowhep:
Somehow merging the master branch updates make this branch fail test for "connection failed"

models/ResumeModel.ts
```diff
@@ -24,7 +24,7 @@ export class ResumeModel extends BaseEntity {
   public getPublicResume(): PublicResume {
     return {
       uuid: this.uuid,
-      user: this.user.uuid,
+      user: this.user.getPublicProfile(),
```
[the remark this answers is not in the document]
shravanhariharan2: We still want this field to be optional, since `this.user` might not always be defined. So e.g. when a user tries to view their own resumes, which is an operation where the Resume <-> User join is not done, then this line of code will throw an error. We need to do something like:

```ts
const publicResume: PublicResume = {
  uuid: this.uuid,
  isResumeVisible: this.isResumeVisible,
  url: this.url,
  lastUpdated: this.lastUpdated
}
if (this.user) publicResume.user = this.user.getPublicProfile();
return publicResume
```

And `PublicResume` should have `user` as an optional field. Let me know if there's any confusion around this.

models/UserModel.ts
shravanhariharan2: Apologies for revisiting such an old comment, but re-visiting this code I don't think what I recommended here is actually good practice. I think having a field here un-explicitly always returning `null` is something very easy to look over, so instead, I'd rather us just outright not specify the field here and instead make `resumes` optional in `PrivateProfile`. That way, we can return this object without the `resumes` field, and we can continue to add it as we do in `UserAccountService.getFullUserProfile`. I think this way gets rid of any ambiguity the code has by always returning `null`.

We can also remove the comment above mentioning that "resumes is null by design".

types/ApiResponses.ts
```diff
@@ -305,6 +305,7 @@ export interface PrivateProfile extends PublicProfile {
   accessType: UserAccessType,
   state: string,
   credits: number,
+  resumes: PublicResume[],
```
shravanhariharan2, line 308: As per above comment, let's make this optional

shravanhariharan2 (commented):
Last few small code comments addressing optionality of fields. Otherwise LGTM, will approve after these are addressed

shravanhariharan2:
> Somehow merging the master branch updates make this branch fail test for "connection failed"

Hmm I've never actually seen this test error before - seems like when I retried the run, the tests worked properly, but I will investigate what might've been happening here a bit more. I don't think it's anything related to this specific PR

shravanhariharan2 (approved):
lgtm

PR #323, opened 2022-11-30
Implement user handles

### LINKED ISSUE #75: add custom profile URLs
Allow users to have custom profile URLs, e.g. `https://members.acmucsd.com/user/sumeetbansal`. Will probably require an additional `customUrl` column in the `Users` table and modifying the existing `GET /api/v1/users/:uuid` route.

shanekim28:
Closes #75


farisashai:
Completed functionality for user handles.

Necessary behaviors:
- Get user handle as a field in public profile to view your handle before editing it.
- Get user profile by handle.
- Patch handle field.
- Initialize default handle for all existing accounts and when new account is registered.

API Functionality
**Register Route Provides Valid Default Handle**
Screenshot 2023-02-16 at 9 43 33 PM

**Get My User Profile with Handle Field**
Screenshot 2023-02-16 at 9 44 02 PM

**Get User Profile by Handle**
Screenshot 2023-02-16 at 10 00 45 PM

**Patch User Handle**
Screenshot 2023-02-16 at 10 00 28 PM

farisashai (approved):
LGTM 🚀

tests/handles.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    // Persist to Database
+    await new PortalState().createUsers(member).write();
+
+    // Get API Controller
+    const userController = ControllerFactory.user(conn);
+    const handle = 'mycustomhandle';
+
+    // Get Updated User Profile
+    const patchedUserResponse = await userController.patchCurrentUser({ user: { handle } }, member);
```
shravanhariharan2, line 31: nit: for consistency with other tests, let's remove the comments here since all tests follow this structure, and let's make the test descriptions start with uppercase and every other word be lowercase instead of all uppercase

shravanhariharan2: also nit: 'Logged In User Can Successfully Set Their Own Handle' -> 'Logged-in user can *update* their own handle' so that it's clear that what's being tested is the update rather than the "set" which can happen in registerUser

tests/handles.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    expect(patchedUserResponse.error).toBeNull();
+
+    const patchedUser = patchedUserResponse.user;
+
+    expect(patchedUser).toEqual({ ...member.getFullUserProfile(), handle });
+  });
+
+  test('User Cannot Set Handle Using an Invalid Handle Format', async () => {
+    // TODO: I can't get this test to reject with an error because validators aren't run here for some reason
```
shravanhariharan2, line 41: Right, the decorators only run when the API is hit via. HTTP (my bad, I should probably add this to the testing guidelines). But what we can do is manually invoke the validator that we want to test, as done here: https://github.com/acmucsd/membership-portal/blob/master/tests/feedback.test.ts#L47-L56

tests/handles.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    // const userController = ControllerFactory.user(conn);
+
+    // await expect(userController.patchCurrentUser({ user: { handle: shortHandle } }, member)).rejects.toThrow('');
+    // await expect(userController.patchCurrentUser({ user: { handle: longHandle } }, member)).rejects.toThrow('');
+    // await expect(userController
+    //   .patchCurrentUser({ user: { handle: specialCharHandle } }, member)).rejects.toThrow('');
+  });
+
+  // TODO: Do we need this test if there is no UUID being passed in?
+  test('Logged In User Cannot Set a Handle for a Different User', async () => {
```
shravanhariharan2, line 60: I'm a bit unsure what this means, but it seems like this and the next test are essentially testing the same thing, so I don't think we need this test unless there's something I'm missing

tests/handles.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    const userController = ControllerFactory.user(conn);
+
+    const userResponse = await userController.getCurrentUser(member);
+
+    expect(userResponse.error).toBeNull();
+
+    const { user } = userResponse;
+
+    expect(user.handle).toBe(member.handle);
+  });
```
shravanhariharan2, line 108: nit: we don't need a newline between each line of code, we can just group them by their purpose (e.g. group 99 and 101 for making the controller call, and 103-107 for the assertions.

tests/handles.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  test("Logged In User Can Fetch Multiple Profiles by User's Handles", async () => {
```
shravanhariharan2, line 110: What is this testing that hasn't been tested already? If this is testing finding a user's profile by their handle that isn't the current user, then I think only 1 API call / assertion is necessary instead of looping over multiple users/API calls, since those are essentially testing the same thing.

tests/handles.test.ts
```diff
@@ -0,0 +1,126 @@
+import { ControllerFactory } from './controllers';
+import { DatabaseConnection, PortalState, UserFactory } from './data';
+
+beforeAll(async () => {
```
shravanhariharan2, line 4: nit: let's rename the file to `user.handles.test.ts` again to make the context clear of what "handles" refer to.

shravanhariharan2 (commented):
Functionality looks good, most of my comments are nitpicks around formatting, conventions, and testing. Let me know if there are any questions with anything

api/validators/GenericRequests.ts
```diff
@@ -18,6 +18,12 @@ export class UuidParam {
   uuid: Uuid;
 }
 
+export class HandleParam {
+  @IsAlphanumeric()
```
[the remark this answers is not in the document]
farisashai: So tbh I was initially planning to set the default value for user handles to `{firstName}-{lastName}-{hash}` as we discussed, but then I realized that I couldn't use the existing validators for `IsAlphanumeric` if I included a dash and would have to create a custom `IsValidHandle` validator. 

I did try the approach of making a custom validator but couldn't figure out how to return different default error messages depending on whether the length constraint or regex was broken. If you know how to handle an error message for invalid length and invalid characters in the same validator and think it's worth changing, I can switch the default to include the dash. 

I do think that it would improve readability of handles as a bonus.

farisashai:
How do we deal with upper/lower case handle values? Should they be treated as identical if the handle is "Faris-Ashai-123" and the API query param is "faris-ashai-123"?

farisashai:
Note: I want to test that the default value is exactly 32 characters for a name that exceeds 32 char length also

tests/auth.test.ts
```diff
@@ -97,6 +98,82 @@ describe('account registration', () => {
...
+    verify(emailService.sendEmailVerification(user.email, user.firstName, anyString()))
+      .never();
+  });
+
+  test('User can include an optional handle to be set', async () => {
+    const conn = await DatabaseConnection.get();
+    const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
+
+    await new PortalState()
+      .createUsers(admin)
```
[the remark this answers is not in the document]
farisashai: I followed the same syntax as the first test in this file `user can register a new account` where we use the admin account to query for the `getUser` to validate that the registration was successful and the profile can be viewed by an existing user. The account arguably doesn't need to be an `admin` in either of these test cases since we are fetching the public profile type. 

I just changed it to a be a standard user for the query with no functional difference - we may want to adjust the other test case also if this is a concern in any way.

api/validators/GenericRequests.ts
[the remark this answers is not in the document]
shravanhariharan2: We can split both these validations into two separate validators - the `Length(3,32)` decorator, and a custom `HasValidSymbols` (or some similarly named validator) decorator that checks the dashes/regex.

api/decorators/Validators.ts
```diff
@@ -47,6 +47,23 @@ export function IsValidName(validationOptions?: ValidationOptions) {
   return templatedValidationDecorator(NameValidator, validationOptions);
 }
 
+@ValidatorConstraint()
+class HandleValidator implements ValidatorConstraintInterface {
+  regex = /^[a-zA-Z0-9-]+$/;
```
shravanhariharan2, line 52: nit: let's add a small comment here on what characters get matched with this regex (just so that developers who might not know how to parse regex still understand what's happening here).

api/decorators/Validators.ts
```diff
@@ -47,6 +47,23 @@ export function IsValidName(validationOptions?: ValidationOptions) {
...
+
+  validate(handle: string): boolean {
+    return handle.search(this.regex) !== -1;
+  }
+
+  defaultMessage(): string {
+    return 'Your handle contains invalid characters.';
+  }
```
shravanhariharan2, line 60: Should we specify what are valid characters? Either here or on the frontend?

migrations/0035-add-handle-column-to-users-table.ts
```diff
@@ -0,0 +1,36 @@
...
+      isUnique: true,
+    }));
+
+    // Populate the "handle" column of "Users" table in every row with a default value
+    // formatted as {firstName}{lastName}{hash}
+    await queryRunner.query(`
+      UPDATE "${TABLE_NAME}"
+      SET handle = "firstName" || '-' || "lastName" || '-' || substr(md5(random()::text), 0, 7)
+    `);
+
```
shravanhariharan2, line 27: nit: let's update this comment to add the dashes

shravanhariharan2: Also, what if this default value is longer than 32 characters? Should we apply our length constraints here as well (similar to how we do in `UserAccountService`?)

services/UserAccountService.ts
```diff
@@ -33,6 +34,28 @@ export default class UserAccountService {
     return user;
   }
 
+  public async findByHandle(handle: string): Promise<UserModel> {
+    const user = await this.transactions.readOnly(async (txn) => Repositories
+      .user(txn)
+      .findByHandle(handle));
+    if (!user) throw new NotFoundError('User was not found');
+    return user;
```
shravanhariharan2, line 42:
```suggestion
      .findByHandle(handle));
    if (!user) throw new NotFoundError('No user associated with this handle was found');
    return user;
```

services/UserAccountService.ts
```diff
@@ -33,6 +34,28 @@ export default class UserAccountService {
...
+  }
+
+  /**
+   * Generate a default user handle in the format of "[firstName][lastName][6 digit has]"
+   * @param firstName User's first name
+   * @param lastName User's last name
+   * @returns default user handle
+   */
```
shravanhariharan2, line 50: nit: let's update the comment to add the dashes here as well. Also, I think the param and return doc comments aren't really necessary, and since we don't include them anywhere else, let's remove them.

services/UserAuthService.ts
```diff
@@ -28,10 +29,17 @@ export default class UserAuthService {
       const userRepository = Repositories.user(txn);
       const emailAlreadyUsed = !!(await userRepository.findByEmail(registration.email));
       if (emailAlreadyUsed) throw new BadRequestError('Email already in use');
+      if (registration.handle) {
+        const userHandleTaken = await userRepository.isHandleTaken(registration.handle);
+        if (userHandleTaken) throw new BadRequestError('This handle is already in use.');
+      }
+      const userHandle = registration?.handle
+         ?? UserAccountService.generateDefaultHandle(registration.firstName, registration.lastName);
```
shravanhariharan2, line 37: Do we need the optional chaining here? I'm not sure there's ever a case where `registration` is null or undefined. If we don't need it, let's remove it

tests/auth.test.ts
```diff
@@ -97,6 +98,110 @@ describe('account registration', () => {
...
+
+    const user = {
+      email: 'acm@ucsd.edu',
+      firstName: 'ACM',
+      lastName: 'UCSD',
+      password: 'password',
+      major: UserFactory.major(),
+      graduationYear: UserFactory.graduationYear(),
+      handle: existingMember.handle,
+    };
```
shravanhariharan2, line 118: nit: I think we should implement a `UserRegistrationFactory` to generate these fields for us, so that the tests are shorter and make it more clear which fields we are intentionally setting vs. which ones are arbitrary. E.g., every field in this object doesn't matter except handle, so we should have something instead like:
```ts
const user = UserRegistrationFactory.fake({ handle: existingMember.handle });
```

tests/auth.test.ts
```diff
@@ -97,6 +98,110 @@ describe('account registration', () => {
...
+
+    const emailService = mock(EmailService);
+    when(emailService.sendEmailVerification(user.email, user.firstName, anyString()))
+      .thenResolve();
+    const authController = ControllerFactory.auth(conn, instance(emailService));
+    const registerRequest = { user };
+    await expect(authController.register(registerRequest, FactoryUtils.randomHexString()))
+      .rejects.toThrow('This handle is already in use.');
```
shravanhariharan2, line 126: nit: it's a bit unclear at a glance what the purpose of `FactoryUtils.randomHexString()` is here. Let's initialize a properly named variable for it instead and pass that variable into here.

tests/auth.test.ts
```diff
@@ -97,6 +98,110 @@ describe('account registration', () => {
...
+
+    verify(emailService.sendEmailVerification(user.email, user.firstName, anyString()))
+      .never();
+  });
+
+  test('User registration with a long name truncates handle to exactly 32 characters', async () => {
```
shravanhariharan2, line 132:
```suggestion

  test('User registration with a full name longer than 32 characters truncates handle to exactly 32 characters', async () => {
```

tests/auth.test.ts
```diff
@@ -97,6 +98,110 @@ describe('account registration', () => {
...
+      .write();
+
+    const user = {
+      email: 'acm@ucsd.edu',
+      firstName: 'ACMACMACMACMACMACMACMACMACMACMACMACMACMACM',
+      lastName: 'UCSDUCSDUCSDUCSDUCSDUCSDUCSDUCSD',
+      password: 'password',
+      major: UserFactory.major(),
+      graduationYear: UserFactory.graduationYear(),
+    };
```
shravanhariharan2, line 147: nit: instead of hardcoding a long firstName and lastName here, let's instead generate a random string of a specified length so that it's clearer what length of the full name is. e.g.:
```ts
const user = UserRegistrationFactory.fake({
  firstName: faker.datatype.string(20),
  lastName: faker.datatype.string(20),
})
```

tests/auth.test.ts
```diff
@@ -97,6 +98,110 @@ describe('account registration', () => {
...
+      firstName: user.firstName,
+      lastName: user.lastName,
+      major: user.major,
+      graduationYear: user.graduationYear,
+      bio: null,
+      points: 0,
+      handle: user.handle,
+      uuid: registerResponse.user.uuid,
+      profilePicture: null,
+    });
```
shravanhariharan2, line 199: Since we only care about the handle in this test, we only need to assert that the handle is what we'd expect it to be, and not the other fields. So let's just assert the handles and omit the other fields.

tests/auth.test.ts
shravanhariharan2: This will also be useful for future test cases since we haven't fully written out our user registration test suite yet, so those tests would use also use this factory.

tests/user.handles.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  test('Update handle error: too short', async () => {
+    const MIN_LENGTH = 3;
+    const user = UserFactory.fake({ handle: 'a' });
+
+    const errors = await validate(plainToClass(UserPatches, { ...user,
+      passwordChange: {
+        password: 'mypassword',
+        newPassword: 'mypassword',
+        confirmPassword: 'mypassword',
+      } }));
```
shravanhariharan2, line 93: Let's have these password changes be stored in a separate variable for improved readability, and let's add a comment on why these need to be included (i.e. mentioning that class-validator still validates optional fields for some reason and that it would error out if these fields aren't included).

tests/user.handles.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    expect(errors).toBeDefined();
+    expect(errors).toHaveLength(1);
+    expect(errors[0].property).toEqual('handle');
+    expect(errors[0].constraints).toBeDefined();
+    expect(Object.keys(errors[0].constraints)).toHaveLength(1);
+    expect(errors[0].constraints.length).toBeDefined();
+    expect(errors[0].constraints.length)
+      .toBe(`handle must be longer than or equal to ${MIN_LENGTH} characters`);
```
shravanhariharan2, line 102: Excellent coverage here 👌🏾

tests/user.handles.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  });
+
+  test('Update handle error: too long', async () => {
+    const MAX_LEN = 32;
+    const user = UserFactory.fake({ handle: 'a'.repeat(2 * MAX_LEN) });
```
shravanhariharan2, line 107: nit: since we use faker conventionally anyways, let's just use faker here instead to match the other tests:
```suggestion
    const MAX_LEN = 32;
    const user = UserFactory.fake({ handle: faker.datatype.hexadecimal(2 * MAX_LENGTH) });
```
Introducing randomness like this is useful to make it clear that the only thing we care about controlling is the length, and not the actual string.

tests/user.handles.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    const errors = await validate(plainToClass(UserPatches, { ...user,
+      passwordChange: {
+        password: 'mypassword',
+        newPassword: 'mypassword',
+        confirmPassword: 'mypassword',
+      } }));
```
shravanhariharan2, line 114: nit: same as above with moving the password changes to a separate variable (e.g. `const requiredValidatorFields = {...}`)

tests/user.handles.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    expect(errors).toHaveLength(1);
+    expect(errors[0].property).toEqual('handle');
+    expect(errors[0].constraints).toBeDefined();
+    expect(Object.keys(errors[0].constraints)).toHaveLength(1);
+    expect(errors[0].constraints.HandleValidator).toBeDefined();
+    expect(errors[0].constraints.HandleValidator)
+      .toBe('Your handle contains invalid characters.');
+  });
+
+  test('Update handle error: too long and invalid characters', async () => {
```
shravanhariharan2, line 146:
```suggestion
  test('Update handle error: too short and invalid characters', async () => {
```

shravanhariharan2 (commented):
One main comment regarding the default handle values set by the migration. The rest are again mostly nitpicks around code quality and conventions

tests/auth.test.ts
farisashai: lowkey copied the code from the other auth test that's doing the same thing with the hex string so idk what it means

migrations/0035-add-handle-column-to-users-table.ts
```diff
@@ -0,0 +1,36 @@
...
+      columnNames: ['handle'],
+      isUnique: true,
+    }));
+
+    // Populate the "handle" column of "Users" table in every row with a default value
+    // formatted as {firstName}-{lastName}-{hash} truncated to 32 characters
+    await queryRunner.query(`
+      UPDATE "${TABLE_NAME}"
+      SET handle = substr("firstName" || '-' || "lastName", 0, 26) || '-' || substr(md5(random()::text), 0, 7)
+    `);
```
shravanhariharan2, line 26: Is this indexing math correct? From [online](https://w3resource.com/PostgreSQL/substr-function.php) it seems like the third argument of `substr` is the number of characters to extract, which would mean that `firstName-lastName` would be 26 chars, the next dash would be 1 char, and then the final `substr` would be 7 chars, totalling 34 chars instead of 32. Have you verified that it's not 34 characters that's the limit here?

tests/auth.test.ts
shravanhariharan2: Ok, let's just refactor that in a separate PR then. I think it's supposed to be the `accessCode` param.

shravanhariharan2 (approved):
Super great work 👍🏾 thanks for all the patience and for taking over the PR. One last just to make sure everything is correct, please address it before merging. Also, since this PR would cause a break without manual migration, let's make sure to coordinate that before merging.

shravanhariharan2 (changes requested):
Revoking approval because we haven't dealt with this case yet:
> How do we deal with upper/lower case handle values? Should they be treated as identical if the stored database handle is "Faris-Ashai-123" and the API query param is "faris-ashai-123"?

Let's disallow uppercase for handles, similar to how Instagram does. We should also make the default handle created by the migration do only lower case as well.

migrations/0035-add-handle-column-to-users-table.ts
```diff
@@ -0,0 +1,36 @@
...
+      columnNames: ['handle'],
+      isUnique: true,
+    }));
+
+    // Populate the "handle" column of "Users" table in every row with a default value
+    // formatted as {firstName}-{lastName}-{hash} truncated to 32 characters
+    await queryRunner.query(`
+      UPDATE "${TABLE_NAME}"
+      SET handle = lower(substr("firstName" || '-' || "lastName", 0, 26) || '-' || substr(md5(random()::text), 0, 7))
+    `);
```
shravanhariharan2, line 26: Is this setting the same length hash as the one that's generated on user signup? It looks like this one is of length 7, but the signup one is of length 6:
```ts
// Hexadecimals look like 0x1b9Dle so we have to truncate the fixed '0x'.
const hashValue = faker.datatype.hexaDecimal(6).slice(2);
```

farisashai: Tested the migration again to confirm that the math is right - not sure why the substr arguments are so weird but it seems like its not inclusive for the start or end index so length is 25 (handle) + 5 hash + 2 dashes = 32

Screenshot 2023-03-29 at 6 31 26 PM

shravanhariharan2: >not sure why the substr arguments are so weird but it seems like its not inclusive for the start or end index so length is 25 (handle) + 5 hash + 2 dashes = 32

It seems like the hashes here are of length 6 though? Looks like the handle including the first '-' is of length 25, the next dash is 1, and the hash is 6, resulting in 32 total. The reason why the start/end positioning might be weird is because I think Postgres does 1-based indexing (see example [here](https://www.postgresqltutorial.com/postgresql-string-functions/postgresql-substring/)), so starting at 0 may make it include an empty string at the start which includes an extra character. Otherwise, I'm unsure why specifying `26` makes the string length 25, even if the string is longer than 26 characters.

I hate to stall this PR even further, but can we quickly replace the 0-indexing with 1-indexing and see if this results in the same behavior? I.e. let's just replace the 0, 26 with 1, 25 and the 0, 7 with 1, 6 and verify the output. I think this will sharpen our understanding of Postgres and substring indexing, while also making the code less confusing. Appreciate the patience here

shravanhariharan2: Will approve once this is verified.

shravanhariharan2 (approved):

steets250 (approved):
Dibs on `@steets250`

PR #324, opened 2023-01-21
Add new devs to user seeds and remove old devs

shravanhariharan2:
Adds a few frontend devs to the user seeds for ease of testing (e.g. verifying emails were received in personal inbox). Also removes a few old devs who are no longer active on the development team.


meshachadoe (approved):
Looks good.

PR #325, opened 2023-01-25
Add required 'isResumeVisible' field to upload resume route

shravanhariharan2:
Adds a field to set resume visibility on resume upload. This is a requirement for the route per the design of resume uploads, but until now it has not existed in the API spec:

image


dowhep (approved):
LGTM rocket rocket cake

PR #327, opened 2023-02-11
Make facebook link optional

farisashai:
We shouldn't be throwing an error on missing event URL since the column is (correctly) nullable. 

Screenshot 2023-02-10 at 11 24 20 PM

Screenshot 2023-02-10 at 11 26 13 PM


shravanhariharan2 (approved):
Good catch, looks good to me. Can we remove the `.nvmrc` before merging and add it in a separate PR if needed?

PR #329, opened 2023-02-18
Add DELETE /resume route

dowhep:
Implemented a route to delete the resume that users uploaded. Need to add unit test atm


tests/resume.test.ts
```diff
@@ -220,3 +220,29 @@ describe('patch resume', () => {
...
+    const resumeFile = FileFactory.pdf(Config.file.MAX_RESUME_FILE_SIZE / 2);
+    const fileLocation = 'fake location';
+
+    const storageService = Mocks.storage(fileLocation);
+    const resumeController = ControllerFactory.resume(
+      conn,
+      instance(storageService),
+    );
+    const request = { isResumeVisible: true };
+    await resumeController.uploadResume(resumeFile, request, member);
```
shravanhariharan2, line 239: Instead of relying on another controller call in a test that should only be testing `deleteResume`, let's use the `ResumeFactory` to generate a resume and use `PortalState::createResumes` to write that resume to the DB. That way, we can avoid this unnecessary `uploadResume()` call and go to directly testing `deleteResume()`.

See my comment on Meshach's PR here for more details/context: https://github.com/acmucsd/membership-portal/pull/317#discussion_r1093813060

dowhep: Yeah that was my initial thought process, but the process of desperately debugging led me astray. I can fix it real quick.

tests/resume.test.ts
```diff
@@ -220,3 +220,29 @@ describe('patch resume', () => {
...
+
+    resumeController.deleteResume(member);
+
```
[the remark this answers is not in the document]
dowhep: That was the single most impactful nit I have ever gotten

api/controllers/ResumeController.ts
```diff
@@ -65,4 +67,12 @@ export class ResumeController {
...
     return { error: null, resumes: resumes.map((resume) => resume.getPublicResume()) };
   }
+
+  @Delete()
+  async deleteResume(@AuthenticatedUser() user: UserModel): Promise<DeleteResumeResponse> {
+    const resume = await this.resumeService.getUserResume(user);
+    if (resume) await this.storageService.deleteAtUrl(resume.url);
+    await this.resumeService.deleteResume(user);
+    return { error: null };
+  }
```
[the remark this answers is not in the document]
dowhep: Sounds good, I can quickly change that to uuid

services/ResumeService.ts
```diff
@@ -61,4 +61,15 @@ export default class ResumeService {
       .resume(txn)
       .findByUserUuid(user.uuid));
   }
+
+  public async deleteResume(uuid: string, user: UserModel): Promise<ResumeModel> {
+    return this.transactions.readWrite(async (txn) => {
+      const resumeRepository = Repositories.resume(txn);
+      const resume = await resumeRepository.findByUuid(uuid);
+      if (!resume) throw new NotFoundError('Cannot find a resume of given uuid');
```
shravanhariharan2, line 69: nit: to match other "not found" error messages in other services
```suggestion
      const resume = await resumeRepository.findByUuid(uuid);
      if (!resume) throw new NotFoundError('Resume not found');
```

services/ResumeService.ts
```diff
@@ -61,4 +61,15 @@ export default class ResumeService {
...
+      if (resume.user.uuid !== user.uuid) throw new ForbiddenError('Cannot delete a resume of another user');
```
shravanhariharan2, line 70:
```suggestion
      if (resume.user.uuid !== user.uuid) throw new ForbiddenError('Cannot delete a resume that belongs to another user');
```

tests/mocks/MockFactory.ts
```diff
@@ -1,8 +1,10 @@
 import { anything, mock, when } from 'ts-mockito';
 import StorageService from '../../services/StorageService';
 
+const DEFAULT_FILE_URL = 's3.amazonaws.com/default-file-url.pdf';
```
shravanhariharan2, line 4: nit: let's add the leading 'https' to match the actual format of s3 urls even more
```suggestion
const DEFAULT_FILE_URL = 'https://s3.amazonaws.com/default-file-url.pdf';
```

tests/resume.test.ts
```diff
@@ -82,7 +82,7 @@ describe('upload resume', () => {
       .write();
 
     const resume = FileFactory.pdf(Config.file.MAX_RESUME_FILE_SIZE / 2);
-    const fileLocation = 'fake location';
+    const fileLocation = 's3.amazon.com/upload-resume.pdf';
```
shravanhariharan2, line 85: nit: instead of defining and using the same `fileLocation` for each route, let's just create a static `FILE_LOCATION` variable at the top of this file set to the url `https://s3.amazonaws.com/upload-resume.pdf` and just use that same one in every test.

tests/resume.test.ts
```diff
@@ -220,3 +220,63 @@ describe('patch resume', () => {
...
+    const storageService = Mocks.storage();
+    const resumeController = ControllerFactory.resume(
+      conn,
+      instance(storageService),
+    );
+    await resumeController.deleteResume(params, member);
+
+    const repository = conn.getRepository(ResumeModel);
+    const resumesStored = await repository.find({ user: member });
+
```
shravanhariharan2, line 245: nit: instead of using the `ResumeRepository`, let's stick with our convention of using `conn.manager.findOne` as we do in other tests (e.g. [here](https://github.com/acmucsd/membership-portal/blob/master/tests/merchOrder.test.ts#L194-L199))

tests/resume.test.ts
```diff
@@ -220,3 +220,63 @@ describe('patch resume', () => {
...
+      .createUsers(member)
+      .createResumes(member, resume)
+      .write();
+
+    const params = { uuid: resume.uuid };
+    const resumeController = ControllerFactory.resume(conn);
+
+    await expect(
+      resumeController.deleteResume(params, anotherMember),
+    ).rejects.toThrowError(ForbiddenError);
```
shravanhariharan2, line 264: nit: let's assert the error message matches the exact error we'd expect to be thrown, instead of just the type. If we decide to throw a different kind of `ForbiddenError` in a future PR, then we would need to distinguish between the two errors in our tests, so we should do it right now instead to prevent future headaches.

tests/resume.test.ts
```diff
@@ -220,3 +220,63 @@ describe('patch resume', () => {
     ).rejects.toThrowError(ForbiddenError);
   });
 });
+
+describe('delete resume', () => {
+  test('delete resume route successfully deletes user resume', async () => {
+    const conn = await DatabaseConnection.get();
+    const member = UserFactory.fake();
+    const resume = ResumeFactory.fake({ isResumeVisible: false, user: member });
+    await new PortalState()
```
shravanhariharan2, line 229: nit: it doesn't really matter what the `isResumeFalse` attribute here is set to, so let's just exclude it from the factory call here. That way, it's clearer that the value doesn't matter since it's randomly generated, vs. specifying it and having the developer need to guess whether or not its value does matter.

tests/resume.test.ts
```diff
@@ -220,3 +220,63 @@ describe('patch resume', () => {
...
+      .createUsers(member)
+      .write();
+
+    const params = { uuid: resume.uuid };
+    const resumeController = ControllerFactory.resume(conn);
+
+    await expect(
+      resumeController.deleteResume(params, member),
+    ).rejects.toThrowError(NotFoundError);
+  });
```
shravanhariharan2, line 281: nit: same as above with specifying the error message.

shravanhariharan2 (commented):
Behavior looks good to me, pretty much all nitpicks on error message descriptions or testing.

tests/resume.test.ts
dowhep: I think it might be more reasonable to put it within the overall upload describe section since only that section use this specific url.

tests/resume.test.ts
dowhep: Sounds good, I think I should use find to be consistent with future for scalable amount of resumes.

tests/resume.test.ts
dowhep: I found a method "toThrowErrorMatchingInlineSnapshot" that seems to be working. Would that create a new file since it compares snapshot?

api/controllers/ResumeController.ts
```diff
@@ -65,4 +67,11 @@ export class ResumeController {
     const resumes = await this.resumeService.getVisibleResumes();
     return { error: null, resumes: resumes.map((resume) => resume.getPublicResume()) };
   }
+
+  @Delete()
+  async deleteResume(@Params() params: UuidParam, @AuthenticatedUser() user: UserModel): Promise<DeleteResumeResponse> {
+    const resume = await this.resumeService.deleteResume(params.uuid, user);
+    if (resume) await this.storageService.deleteAtUrl(resume.url);
+    return { error: null };
+  }
```
shravanhariharan2, line 76: Do we need this `if (resume)` check? When would there be a case where `resume` is falsy? It seems like we're already throwing an error if resume is falsy in the service function, so to me it doesn't seem like `resume` would ever be falsy on line 74.

tests/resume.test.ts
```diff
@@ -74,6 +74,8 @@ describe('resume fetching', () => {
 });
 
 describe('upload resume', () => {
+  const fileLocation = 's3.amazon.com/upload-resume.pdf';
```
shravanhariharan2, line 77:
```suggestion
  const fileLocation = 'https://s3.amazonaws.com/upload-resume.pdf';
```

tests/resume.test.ts
```diff
@@ -217,6 +216,69 @@ describe('patch resume', () => {
     const resumeController = ControllerFactory.resume(conn);
     await expect(
       resumeController.patchResume(params, request, anotherMember),
-    ).rejects.toThrowError(ForbiddenError);
+    ).rejects.toThrowErrorMatchingInlineSnapshot(
+      '"Cannot update a resume of another user"',
+    );
```
shravanhariharan2, line 221: What does this method do? What is an inline snapshot? Is this a better alternative to just using `toThrow('Cannot update a resume of another user')`?

shravanhariharan2: https://jestjs.io/docs/snapshot-testing#inline-snapshots
https://jestjs.io/docs/expect#tothrowerrormatchinginlinesnapshotinlinesnapshot

It doesn't seem to me that inline snapshots are too relevant to what we want to assert, is there a use case that is useful for us here?

shravanhariharan2 (commented):
Quick comments on testing and one API logic piece

tests/resume.test.ts
dowhep: I did not realize toThrow works with strings as well. I was certain vscode was shouting at me when I tried it initially. Inline snapshot, however, works here as it throws when I have a different string.

api/controllers/ResumeController.ts
dowhep: You are right, this check was not necessary.

tests/resume.test.ts
shravanhariharan2: I still think we should stray away from using this method because it is not relevant to the testing we are trying to do here, even though the method may provide the same expected results we are looking for. It looks like "inline snapshots" are a concept related to [Snapshot Testing](https://jestjs.io/docs/snapshot-testing#snapshot-testing-with-jest), which is meant specifically for UI testing, as explained in the docs.

Including this method here only causes more confusion because in every other test we assert exceptions with `toThrow`, which provides the expected behavior while also semantically making sense - we expect an error to be thrown with this specific error message, and nothing more.

shravanhariharan2 (approved):

dowhep:
LGTM

PR #333, opened 2023-03-10
Update CI to build code into ARM and to tag Docker images 

### LINKED ISSUE #332: Start versioning Docker image tags on deployment
Currently, when a new build is deployed to prod, a Docker image is tagged with the "latest" tag. When CapRover acklowledges the deployment and stores it in its history, it stores all the image versions with the "latest" tag:

image

This poses a problem when it comes to reverting changes - CapRover knows which image to revert to based on the image name/tag that is associated with that version (e.g., versions 89-93 are all associated with `membership-portal-api:latest`). This means that if we want to revert prod to a previous build, CapRover will revert the current build to the `membership-portal-api:latest` image, which _is_ the current image, so nothing gets reverted. This means that we are currently unable to revert to a previous build at all, which can be extremely problematic if there's a complex bug that lands into prod that would require a revert to perserve as much production availability as possible.

We should begin tagging our Docker images with incrementing versions according to [semver](https://semver.org/). Ideally this version should match our `package.json` versioning as well, meaning that we would also need to properly version every change for both Node and Docker.

shravanhariharan2:
Closes #332.

This PR introduces a bunch of changes to our CI to support the migration of the portal instance on AWS to a cheaper instance backed by AWS Graviton processors. Graviton processors are ARM-based processors built by AWS themselves that boast similar levels of performance to traditional x86 processors, but at a cheaper cost due to the fact that they are built in-house. This change is part of an overall effort by us to decrease our AWS costs - the full details and numbers can be found [on our Notion](https://www.notion.so/acmucsd/AWS-Cost-Reduction-d5cf7cadd1254726906942ee004e0ea6).

There are many changes (some subtle) to the CI and overall infrastructure, but here is a breakdown of the notable ones:
* Portal instance now lives on a t4g.small instance vs. a t3a.small instance, and is now running on ARM
* CI now compiles our Docker image into both x86 and ARM for backwards compatibility (and in case we need to switch off back to x86 for any reason). This is done with the help of [Docker's buildx utility](https://docs.docker.com/engine/reference/commandline/buildx/). There are comments in the CI code itself that hopefully explain what's happening to build the Docker image into multiple architectures.
* CI now deploys image onto a new Caprover instance which lives on the new t4g.small instance (which is pointed to by caprover.acmucsd.com, rather than the old app.acmucsd.com).
* Docker images are now tagged with the first 7 characters of the Git commit hash on master - this resolves the issue linked above and allows us to rollback images as needed (more details are in the issue itself).

On merge, this PR will tag the Docker image built with the new Git commit hash structure, so both the prod and testing servers will be running based off of these hash-tagged images, and not the `latest` tag.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,8 +84,6 @@ jobs:
 
           # See https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
           # for details
-
-          # TODO: tag both the git hash AND latest in the build command once prod ready
           command: |
             GIT_HASH=$(echo $CIRCLE_SHA1 | cut -c -7)
             echo "$DOCKER_PASS" | docker login --username $DOCKER_USER --password-stdin
@@ -96,6 +94,7 @@ jobs:
               --cache-from type=local,src=/caches/docker-buildx \
               --cache-to type=local,dest=/caches/docker-buildx,mode=max \
               --tag acmucsd/membership-portal-api:latest \
+              --tag acmucsd/membership-portal-api:$GIT_HASH \
               --push .
       - save_cache:
           key: v1-{{ .Branch }}-{{ epoch }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -55,49 +55,64 @@ jobs:
...
       - run:
           command: |
-            docker run caprover/cli-caprover:v2.1.1 caprover deploy --caproverUrl https://captain.app.acmucsd.com --caproverPassword $CAPROVER_PASS --caproverApp membership-portal-api --imageName acmucsd/membership-portal-api:latest
-            docker run caprover/cli-caprover:v2.1.1 caprover deploy --caproverUrl https://captain.app.acmucsd.com --caproverPassword $CAPROVER_PASS --caproverApp membership-portal-api-testing --imageName acmucsd/membership-portal-api:latest
+            GIT_HASH=$(echo $CIRCLE_SHA1 | cut -c -7)
+            docker run caprover/cli-caprover:v2.1.1 caprover deploy --caproverUrl https://captain.caprover.acmucsd.com --caproverPassword $CAPROVER_PASS --caproverApp membership-portal-api-testing --imageName acmucsd/membership-portal-api:$GIT_HASH
+            docker run caprover/cli-caprover:v2.1.1 caprover deploy --caproverUrl https://captain.caprover.acmucsd.com --caproverPassword $CAPROVER_PASS --caproverApp membership-portal-api --imageName acmucsd/membership-portal-api:$GIT_HASH
 
 workflows:
   version: 2
```
dowhep, line 118: Is the workflow version supposed to differ from the config version?

dowhep (commented):
read 3:13

.circleci/config.yml
shravanhariharan2: Good catch, seems like from [here](https://circleci.com/docs/configuration-reference/#workflow-version) the version field isn't required if the overall config version is 2.1, so I'll remove this

.circleci/config.yml
```diff
@@ -55,52 +55,66 @@ jobs:
   image:
     working_directory: /app
     docker:
-      - image: docker:20.10.2
+      # Image and code adapted from https://www.docker.com/blog/multi-arch-build-what-about-circleci/
+      - image: jdrouet/docker-with-buildx:stable
```
shravanhariharan2, line 59: Add more comments here about what this image does and how buildx works

dowhep (approved):
LGTM 🚀 !

PR #338, opened 2023-04-04
Fix DELETE /resume route and resume service patch resume

dowhep:
Closes #[haven't called outed issue].


dowhep:
The response for DELETE /resume now looks like 
image

dowhep:
And when PATCHing a resume with non-existent uuid, instead of seeing
image
we should now see
image

shravanhariharan2 (approved):

PR #344, opened 2023-04-07
Implement Express Check-in

### LINKED ISSUE #341: Express Checkin
If we could create a new route like `POST /checkin {email, checkinCode}`, we could add an Express Check-In feature for larger ACM events such as Kickoff to allow new users to temporarily bypass registration to get check-in points.

User Flow:
- Potential new ACM member comes to their first event and does not have a Membership Portal account yet.
  - (We require them to check in before entering and going through the full registration process would take too long and hold up the line.)
- Instead, they go to a page from the login page on the portal to submit a form with their UCSD email and checkin code for the event currently happening and get a success screen on the front-end that they have been checked in.
- On the backend, we save the event checkin data corresponding to some kind of account (not sure on the details)
  - Potential Middle Step: Email them **"Thanks for checking into your first ACM event. Here's what you can use your points for. Make sure to complete the registration process at members.acmucsd.com/register"**
- When the user later registers, they already have points on their portal account and the event attendance history.

shravanhariharan2:
Closes #341


shravanhariharan2:
TODO:
* Write tests for POST /attendance/expressCheckin
* Implement proper actions on POST /registration if a user has already done an express checkin
* Write tests for above
* Explain design decisions in PR comments

api/controllers/AttendanceController.ts
```diff
@@ -25,16 +29,26 @@ export class AttendanceController {
...
   }
 
+  @UseBefore(UserAuthentication)
   @Get()
   async getAttendancesForCurrentUser(@AuthenticatedUser() user: UserModel): Promise<GetAttendancesForUserResponse> {
     const attendances = await this.attendanceService.getAttendancesForUser(user);
     return { error: null, attendances };
   }
 
+  @UseBefore(UserAuthentication)
```
farisashai, line 39: Is this decorator related to the current PR or a bug fix?

models/EventModel.ts
```diff
@@ -59,6 +60,9 @@ export class EventModel extends BaseEntity {
   @OneToMany((type) => AttendanceModel, (attendance) => attendance.event, { cascade: true })
   attendances: AttendanceModel[];
 
+  @OneToMany((type) => ExpressCheckinModel, (expressCheckin) => expressCheckin.event, { cascade: true })
+  expressCheckins: AttendanceModel[];
```
farisashai, line 64: Does this add column need to be specified in a migration?

services/AttendanceService.ts
```diff
@@ -71,6 +76,36 @@ export default class AttendanceService {
...
 
+  public async attendViaExpressCheckin(attendanceCode: string, email: string): Promise<PublicExpressCheckin> {
+    return this.transactions.readWrite(async (txn) => {
+      const expressCheckinRepository = Repositories.expressCheckin(txn);
+      const userRepository = Repositories.user(txn);
+
+      const hasUserUsedExpressCheckin = await expressCheckinRepository.hasUserUsedExpressCheckin(email);
+      if (hasUserUsedExpressCheckin) {
+        throw new UserError(
+          'You have already done an express check-in before. '
```
farisashai, line 87: I mentioned this in the issue but minor tweak to this error message: If they used Express Checkin for **this specific event,** error to let them know they're all good with checking in already, otherwise send this message

templates/expressCheckinConfirmation.ejs
```diff
@@ -0,0 +1,154 @@
...
+                <!-- END FOOTER -->
+
+                <!-- END CENTERED WHITE CONTAINER -->
+            </div>
+        </td>
+        <td style="font-family: sans-serif; font-size: 14px; vertical-align: top;">&nbsp;</td>
+    </tr>
+</table>
+</body>
+</html>
```
farisashai, line 154: Is there any way for you to show a screenshot of what this email looks like? Also can I suggest tweaks if it is auto-generated to improve the experience? with this feature

farisashai (changes requested):
One comment on an error message and some questions. Code is incredibly clean - idk how you did that so fast but you don't get to skip testing 😈 I gotta get you to 50 comments somehow

api/controllers/AttendanceController.ts
shravanhariharan2: Current PR - previously we've had this decorator over the entire AttendanceController class, meaning that every route is authenticated. This changes in this PR because this new route isn't authenticated, so we need to remove the decorator over the whole class and decorate only the routes that are authenticated.

models/EventModel.ts
shravanhariharan2: Nope, this is purely a TypeORM construct. This field allows `EventModel` to join with `ExpressCheckinModel` and be accessed with `event.expressCheckins`, for ease of use. Internally in the database though, this relationship only needs to be one-way - i.e., `ExpressCheckins.event` references the `Events` table, but the `Events` table doesn't need to know about `ExpressCheckins`, unless it's specifically queried via. TypeORM.

More about this behavior here: https://www.tutorialspoint.com/typeorm/typeorm_relations.htm

templates/expressCheckinConfirmation.ejs
shravanhariharan2: Yes to both - I'll add the screenshot before this PR leaves the draft stage

tests/auth.test.ts
```diff
@@ -90,7 +99,7 @@ describe('account registration', () => {
       .never();
   });
 
-  test('User cannot register with a handle that is already in use', async () => {
+  test('user cannot register with a handle that is already in use', async () => {
```
shravanhariharan2, line 102: Just to stay consistent with the naming of other tests

shravanhariharan2, line 102: Just to stay consistent with the naming of other tests

shravanhariharan2:
Email confirmation example:
image

shravanhariharan2:
There's a few ways we could've gone about designing express checkins at the database level. One naive way that came to mind was as below:

When a user who isn't registered does an express checkin, we create a partial User object and insert only their email into the Users table, so that we can get a uuid generated for them. Then, we insert the Attendances entry and Activity entry into their respective tables that reference that new partial user's uuid. This makes it so that later, when a user registers, we just update the User table columns that need to be updated, and when the user finishes registration, the attendance and activity entries are already there.

While this may seem like a simple and intuitive approach, there are a few key downsides that make the approach both less efficient and less secure. They are the following:
* For the Users table to support a partial user (i.e. just their uuid and their email), we would need to make most other fields nullable (or have some arbitrary default value, like 0 for `graduationYear` or "None" for `major`). This isn't desired because removing the not-null constraints of several columns could lead to potential bugs. For example, if our API validation for user registration or profile editing doesn't properly check for not-null in *all* the fields, then a potentially null field could reach the database for a user that is meant to have all profile fields present. We haven't rigorously tested this not-null validation, and as such, I don't think it's wise to remove the not-null constraints from the Users table.
* Second and more importantly, having the express checkin route populate several tables greatly increases the blast radius for potential spam against this API route. This is one of our only unauthenticated API routes we have, in addition to being one that has a dependency on a third party service, SendGrid. If this route were to receive a barrage of requests from a malicious client, data would populate 3 of our most important tables - Users, Attendances, and Activities. It would require heavy amounts of effort to recover from such an event, so ideally there should be a different access pattern for our database than writing to 3 tables in a single unauthenticated API route.

The solution presented considers both of these flaws with the help of an ExpressCheckins table. When a user does an express checkin, data is only written to this table. Only when a user completes registration is when the rest of the relations are populated. This both limits the blast radius of the express checkin route, while also not modifying the constraints of the Users table. While theoretically the blast radius has only been pushed down to the registration layer, before our V2 Portal UI launch we plan on moving this logic to after email verification is complete. The only reason it's in the registration logic now is so that it can match our regular attendance logic where unverified users can attend events and earn points.

Before this route is in use, we should add some sort of rate limiting so that it cannot be spammed and potentially use up all of our SendGrid email credits.

dowhep (commented):
Looks good overall, but per the new route policy, please add the screenshots of the postman responses of the route to ensure the decorators passed the test.

farisashai (approved):
> Email confirmation example: image

I like this email content but some small tweaks to explain how things operate on the portal maybe. Good to merge afterwards besides this

"
Hi _Shravan_,

Thanks for attending your first ACM event **_Hack School: Intro to React!_**. You've just earned your very first _10_ membership points. Earn more points to climb the leaderboard, make amazing friends and create memories here with our community! You can spend the points you earned and collect at future events at members.acmucsd.com/store on shirts, sweaters, and more swag.

In order to finish becoming a full member and access these benefits, please register for an account on our Membership Portal. You will not be able to check in for future events until registering and verifying your email.

[Register Button] 
Visit members.acmucsd.com/register if the link above doesn't work.

Best Regards,
ACM at UCSD
"

shravanhariharan2:
Postman docs for new route: https://documenter.getpostman.com/view/12949536/TVRedVwQ#fc60a70e-abd0-4b43-94d1-c7efdf75c54a

shravanhariharan2:
Updated email content:
image

PR #350, opened 2023-04-11
Add missing awaits to expect(...).rejects tests

### LINKED ISSUE #328: Add missing 'await' to merch store test
One of the merch store tests is missing an `await` before an `expect.rejects` clause, which causes this seemingly unrelated and nondeterministic error when running the test:
image

Adding the await should fix the error: https://stackoverflow.com/questions/53638949/typeorm-connection-terminated

alexzhang1618:
Closes #328.

Add missing awaits to `expect(...).rejects.ToThrow` tests in merchStore.test.ts.
These previously would cause non-deterministic errors where the DB connection would terminate before the request is fulfilled and checked.


shravanhariharan2:
In this PR are we trying to add awaits to every `expect(...).rejects.toThrow` that is missing an await? Or only specific ones? If we're trying to add `await` to all of them, there's a handful more in `merchOrder.test.ts` (e.g. [here](https://github.com/acmucsd/membership-portal/blob/alex/store-test-await/tests/merchOrder.test.ts#L570), [here](https://github.com/acmucsd/membership-portal/blob/alex/store-test-await/tests/merchOrder.test.ts#L852), and [here](https://github.com/acmucsd/membership-portal/blob/alex/store-test-await/tests/merchOrder.test.ts#L1591)) that would need it too.

One thing to note is that we don't *actually* need to add an `await` for every single `toThrow` call, because async calls that are the last line of code in an async function will normally block. You can see we use this pattern in our repositories (e.g. [here](https://github.com/acmucsd/membership-portal/blob/master/repositories/EventRepository.ts#L14), where we don't `await` the return call because the interpreter will block regardless). So we need to decide whether it makes sense to `await` every single `toThrow` call, or just the ones that aren't the last line of code.

I think it makes more sense to just `await` every single `toThrow` call because that way it's clearer and more consistent to the developer that `await`s are a pattern to be used with `toThrow`, even if they aren't required in some calls. If the intention of the PR is to cover all `toThrow` calls, then let's add `await`s to all of them. Otherwise, we can `await` just the merch store test call that's causing the non-determinism.

alexzhang1618:
Oops the intention was to add `await`s to all `.toThrow` calls, I've updated this now

shravanhariharan2 (approved):
Thanks for taking care of this 👍🏾

PR #351, opened 2023-04-18
Create Multiple Gallery Store Images

### LINKED ISSUE #339: Support array of store merch images
We should restructure the store item images to allow for a list of images rather than a single image.

Requirements
- Upload multiple new images for one item (limit to X images)
- Get all store items with list of images in the JSON response.
- Patch or delete to remove some subset of the images (minimum 1 required)

dowhep:
Closes #339.


migrations/0036-add-merch-item-image-table.ts
```diff
@@ -0,0 +1,42 @@
+import {MigrationInterface, QueryRunner, Table, TableIndex } from "typeorm";
+
+const TABLE_NAME = 'MerchItemImages';
```
shravanhariharan2, line 3: nit: let's call this table `"MerchandiseItemPhotos"`. Since we use "Merchandise" instead of "Merch" for our other table names, we should stay consistent with that. And let's use "photos" instead of "images" since we are using the term "photos" for the environment variables to limit photo size:
image

dowhep:
Will add the postman requests soon

types/ApiRequests.ts
```diff
@@ -194,6 +194,9 @@ export interface EditMerchItemRequest {
 export interface CreateMerchItemOptionRequest {
   option: MerchItemOption;
 }
+export interface CreateMerchItemPhotoRequest {
+  // position: number;
+}
```
nik-dange, line 199: Is this request for the purpose of just having a request body since all the necessary info is passed through the uuid in the params? If so, is position:number necessary to be commented out or can we represent this better?

dowhep: It was a residue of old code - I originally created this in case I needed to include any parameters. Turns out I don't need any.

dowhep: Nice catch

dowhep:
Postman Request:
image

services/MerchStoreService.ts
```diff
@@ -208,6 +212,33 @@ export default class MerchStoreService {
...
+
+        const photoUpdatesByUuid = new Map(photos.map((photo) => [photo.uuid, photo]));
+
+        item.photos.map((currentPhoto) => {
+          if (!photoUpdatesByUuid.has(currentPhoto.uuid)) return;
+          const photoUpdate = photoUpdatesByUuid.get(currentPhoto.uuid);
+          // photo positions are 0-index
+          return MerchandiseItemPhotoModel.merge(currentPhoto, photoUpdate);
+        });
+
```
maxwn04, line 231: This is a little confusing to me. It just creates a map of uuid to new photos, then changes the old photo with that uuid to the new photo right?

maxwn04 (commented):
I read through the changes. I think it looks good, and I think I understand most of it. Some little things I'm unsure about. Not really sure how to improve it/what else needs to be in a review

services/StorageService.ts
```diff
@@ -68,6 +68,20 @@ export default class StorageService {
...
+    //                      probability of at least one collision.
+
+    let result = '';
+    for (let i = 0; i < stringLength; i += 1) {
+      result += chars.charAt(Math.floor(Math.random() * chars.length));
+    }
+
+    return result;
+  }
+
```
shravanhariharan2, line 84: Why do we need this whole implementation? Can we just return a random uuid instead? Does it need to be a length of 25?

api/controllers/MerchStoreController.ts
```diff
@@ -158,14 +159,27 @@ export class MerchStoreController {
...
+    @AuthenticatedUser() user: UserModel): Promise<CreateMerchPhotoResponse> {
     if (!PermissionsService.canEditMerchStore(user)) throw new ForbiddenError();
-    const picture = await this.storageService.upload(file, MediaType.MERCH_PHOTO, params.uuid);
-    const item = await this.merchStoreService.editItem(params.uuid, { picture });
-    return { error: null, item };
+
+    // generate a random string for the picture url
+    const randomID = StorageService.getRandomString();
+    const picture = await this.storageService.uploadToFolder(file, MediaType.MERCH_PHOTO, randomID, params.uuid);
+    const photo = await this.merchStoreService.createItemPhoto(params.uuid, { picture });
```
shravanhariharan2, line 171: See my comment above on if we actually need `getRandomString()`, but if we don't, you can just set `randomId` equal to `uuid.v4()` to generate a random uuid.

Also nit: `randomId` isn't too useful of a variable name especially in the context of what it's used for - something like `fileName` or `uniqueFileName` might make more sense

shravanhariharan2, line 171: nit: it's unclear by the variable naming what the difference is supposed to be between `picture` and `photo` - semantically they both mean essentially the same thing. How about something like `uploadedPhoto` for the former and `merchPhoto` for the latter?

api/controllers/MerchStoreController.ts
```diff
@@ -158,14 +159,27 @@ export class MerchStoreController {
...
+  }
+
+  @UseBefore(UserAuthentication)
+  @Delete('/item/picture/:uuid')
+  async deleteMerchItemPhoto(@Params() params: UuidParam, @AuthenticatedUser() user: UserModel):
+  Promise<DeleteMerchItemPhotoResponse> {
+    if (!PermissionsService.canEditMerchStore(user)) throw new ForbiddenError();
+    await this.merchStoreService.deleteItemPhoto(params.uuid);
+    return { error: null };
   }
```
shravanhariharan2, line 183: Shouldn't we also delete the file from S3?

api/controllers/MerchStoreController.ts
```diff
@@ -190,7 +204,7 @@ export class MerchStoreController {
     @AuthenticatedUser() user: UserModel): Promise<GetOneMerchOrderResponse> {
     if (!PermissionsService.canAccessMerchStore(user)) throw new ForbiddenError();
     const order = await this.merchStoreService.findOrderByUuid(params.uuid);
-    if (!PermissionsService.canSeeMerchOrder(user, order)) throw new NotFoundError();
+    if (!PermissionsService.canSeeMerchOrder(user, order.getPublicOrderWithItems())) throw new NotFoundError();
```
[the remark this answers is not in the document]
shravanhariharan2: +1, it doesn't seem like PermissionsService changed in this PR, why do we need this change?

services/MerchStoreService.ts
```diff
@@ -208,6 +212,33 @@ export default class MerchStoreService {
...
       }
 
+      if (photos) {
+        // error on duplicate photo uuids
+        const dupSet = new Set();
+        photos.forEach((photo) => {
+          if (dupSet.has(photo.uuid)) { throw new UserError(`Multiple edits is made to photo: ${photo.uuid}`); }
+          dupSet.add(photo.uuid);
+        });
+
```
shravanhariharan2, line 222: To make life much much simpler, I don't think we should support rearranging photos and their indexes for an item. See our Discord discussion [here](https://discord.com/channels/573028991527550986/791028414538186772/1145092442392105150) - needing to deal with rearranging photos and updating indexes (as you can see with this function's implementation) is not clean, not efficient, and not worth the gains we get from supporting this feature (because admins can instead just delete/add photos to modify the list as needed). So, let's remove this function's photo implementation entirely.

services/MerchStoreService.ts
```diff
@@ -279,6 +310,72 @@ export default class MerchStoreService {
     });
   }
 
+  /**
+   * TODO: add global variable
+   * Verify that items have valid options. An item with variants disabled cannot have multiple
+   * options, and an item with variants enabled cannot have multiple option types.
+   */
+  private static verifyItemHasValidPhotos(item: MerchItem | MerchandiseItemModel) {
```
shravanhariharan2, line 318: nit: outdated comment

shravanhariharan2: Also, let's add in the global variable here to avoid magic numbers.

services/MerchStoreService.ts
shravanhariharan2: In fact, let's not worry about photo position at all in the backend. Have a look at how we implemented MerchItemOptionMetadata - it follows the exact same pattern we're looking for here (can remove any option metadata, but can only add to the end), but doesn't have *any* logic around positions that takes place in the backend. This is intentional, so that the frontend can have full agency over how they want to implement positioning (e.g. if they want to allow pre-pending photos to the front, they can use negative positions). The backend's job is to just persist that representation.

services/MerchStoreService.ts
```diff
@@ -279,6 +310,72 @@ export default class MerchStoreService {
...
+   * @returns created item photo
+   */
+  public async createItemPhoto(item: Uuid, properties: MerchItemPhoto): Promise<PublicMerchItemPhoto> {
+    return this.transactions.readWrite(async (txn) => {
+      const merchItem = await Repositories.merchStoreItem(txn).findByUuid(item);
+      if (!merchItem) throw new NotFoundError('Merch item not found');
+
+      // add to the end
+      const position = merchItem.photos.length;
+
```
shravanhariharan2, line 338: Per my comment above on positioning, let's not have any logic around updating positions. We should just make sure that it's a field passed in to `properties` above so that the frontend can pass us the position and we just store it internally

services/MerchStoreService.ts
```diff
@@ -279,6 +310,72 @@ export default class MerchStoreService {
...
+      if (merchItem.photos.length === 1 && !merchItem.hidden) {
+        throw new UserError('Cannot delete the only photo for a visible merch item');
+      }
+      // decrement photo index for all photos after the inserted photo
+      const { position } = photo;
+      merchItem.photos.forEach((p) => {
+        if (p.position > position) {
+          merchStoreItemPhotoRepository.upsertMerchItemPhoto(p, { position: p.position - 1 });
+        }
+      });
```
shravanhariharan2, line 373: Per comment above on positions, no need for this logic - let the frontend handle positions and indexes as they like.

tests/Seeds.ts
```diff
@@ -318,14 +318,25 @@ async function seed(): Promise<void> {
...
+  });
+  const MERCH_ITEM_1_PHOTO_1 = MerchFactory.fakePhoto({
+    merchItem: MERCH_ITEM_1,
+    position: 1,
+  });
+  const MERCH_ITEM_1_PHOTO_2 = MerchFactory.fakePhoto({
+    merchItem: MERCH_ITEM_1,
+    picture: 'https://i.imgur.com/pSZ921P.png',
+    position: 2,
+  });
```
shravanhariharan2, line 339: nit: curious why only the last photo has a picture, but not the first two?

shravanhariharan2: And why photo_0 doesn't have a position?

tests/Seeds.ts
```diff
@@ -388,16 +399,29 @@ async function seed(): Promise<void> {
...
     hasVariantsEnabled: true,
   });
+  const MERCH_ITEM_2_PHOTO_0 = MerchFactory.fakePhoto({
+    merchItem: MERCH_ITEM_2,
+  });
+  const MERCH_ITEM_2_PHOTO_1 = MerchFactory.fakePhoto({
+    merchItem: MERCH_ITEM_2,
+    picture: 'https://i.imgur.com/pSZ921P.png',
+    position: 1,
+  });
```
shravanhariharan2, line 424: Should photo_0 have a position and a picture?

tests/Seeds.ts
```diff
@@ -445,47 +473,81 @@ async function seed(): Promise<void> {
...
     description: 'Guaranteed 2x return on Grailed.',
     monthlyLimit: 2,
     lifetimeLimit: 5,
     hidden: false,
     hasVariantsEnabled: false,
   });
+  const MERCH_ITEM_3_PHOTO_0 = MerchFactory.fakePhoto({
+    merchItem: MERCH_ITEM_3,
+    picture: 'https://i.imgur.com/QNdhfuO.png',
+  });
```
shravanhariharan2, line 485: Should this have a position?

tests/Seeds.ts
```diff
@@ -445,47 +473,81 @@ async function seed(): Promise<void> {
...
     description: 'Salt and pepper not included.',
     monthlyLimit: 3,
     lifetimeLimit: 10,
     hidden: false,
     hasVariantsEnabled: false,
   });
+  const MERCH_ITEM_4_PHOTO_0 = MerchFactory.fakePhoto({
+    merchItem: MERCH_ITEM_4,
+    picture: 'https://i.pinimg.com/originals/df/c5/72/dfc5729a0dea666c31c5f4daea851619.jpg',
+  });
```
shravanhariharan2, line 529: Position

tests/Seeds.ts
```diff
@@ -445,47 +473,81 @@ async function seed(): Promise<void> {
...
     description: 'Self-explanatory.',
     monthlyLimit: 1,
     lifetimeLimit: 2,
     hidden: false,
     hasVariantsEnabled: true,
   });
+  const MERCH_ITEM_5_PHOTO_0 = MerchFactory.fakePhoto({
+    merchItem: MERCH_ITEM_5,
+    picture: 'https://i.etsystatic.com/8812670/r/il/655afa/3440382093/il_340x270.3440382093_cbui.jpg',
+  });
```
shravanhariharan2, line 550: Position

tests/data/MerchFactory.ts
```diff
@@ -35,14 +36,13 @@ export class MerchFactory {
     const fake = MerchandiseItemModel.create({
       uuid: uuid(),
       itemName: faker.datatype.hexaDecimal(10),
-      picture: faker.image.cats(),
+      // picture: faker.image.cats(),
       description: faker.lorem.sentences(2),
```
shravanhariharan2, line 40: nit: if we don't need this code anymore, let's remove it instead of commenting it out

tests/data/MerchFactory.ts
```diff
@@ -51,9 +51,25 @@ export class MerchFactory {
         .createOptions(numOptions)
         .map((option) => MerchandiseItemOptionModel.merge(option, { item: fake }));
     }
+    if (!substitute?.photos) {
+      const numPhotos = FactoryUtils.getRandomNumber(1, 5);
+      fake.photos = MerchFactory
+        .createPhotos(numPhotos)
+        .map((photo, i) => MerchandiseItemPhotoModel.merge(photo, { position: i, merchItem: fake }));
+    }
```
shravanhariharan2, line 59: nit: createPhotos should have the responsibiilty of managing positions, since they're both at the same layer of abstraction (and managing positions is at a lower level of abstraction than creating a merch item). Let's move the positioning logic in there instead

api/validators/MerchStoreRequests.ts
```diff
@@ -129,6 +131,29 @@ export class MerchItemOptionEdit implements IMerchItemOptionEdit {
...
 
+// TODO: use global variable for limit
+export class MerchItemPhoto implements IMerchItemPhoto {
+  @Allow()
+  picture: string;
+
+  @Min(0)
+  @Max(4)
+  position: number;
+}
```
shravanhariharan2, line 142: Given our descoping of positioning logic, I don't think we need any constraints on this position number. I think it makes sense to let position be any number possible, and have the frontend manage whatever number they want them to be, since the only thing that matters is their relative order to each other. e.g., 0, 1, 2, 3 conveys the same thing as -12091, -234, 394, 9023840932.

shravanhariharan2 (changes requested):
Great to see the progress here especially given that it's a fairly daunting feature which requires some nuance to its implementation and scope. Let me know if there's any questions about any of the comments I left.

api/controllers/MerchStoreController.ts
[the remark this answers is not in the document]
dowhep: Seeing this now it does seem unnecessary, I think I was trying to do some refactoring there but wasn't being successful.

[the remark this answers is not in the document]
dowhep: Will remove

services/MerchStoreService.ts
dowhep: Specifically, this part handles updating the positions of the pictures. The merge will propagate this change to the database.

.vscode/settings.json
```diff
@@ -10,5 +10,11 @@
...
   "prettier.singleQuote": true,
   "prettier.trailingComma": "all",
-  "files.trimTrailingWhitespace": true
+  "files.trimTrailingWhitespace": true,
+  "[javascriptreact]": {
+    "editor.tabSize": 2
+  },
+  "[typescriptreact]": {
+    "editor.tabSize": 2
+  }
```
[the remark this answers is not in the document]
dowhep: That is odd. Will remove the tabs settings.

I supposed trimming trailing whitespace is helpful.

services/StorageService.ts
dowhep: I tried to find a random UUID implementation natively. I think nowadays [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) is the standard but it is only supported for node version 19+

api/controllers/MerchStoreController.ts
dowhep: Sounds good.

api/controllers/MerchStoreController.ts
dowhep: wait a sec, I don't think the photo was deleted from the original code base

services/MerchStoreService.ts
dowhep: Sure I will make it consistent

tests/Seeds.ts
dowhep: Fake photo actually already generates a fake url for the object, as well as the default position. I kinda just assign different values when necessary.

dowhep: Perhaps for consistency sake I should make everything explicit.

tests/Seeds.ts
dowhep: Similar to the comment above, I left it blank as it assume default values

tests/data/MerchFactory.ts
dowhep: tbh I left it there because I thought fake.image.cats() is funny.

dowhep: will remove

tests/data/MerchFactory.ts
dowhep: Changed. What do you think about moving merchItem to createPhotos as an optional field?

api/controllers/MerchStoreController.ts
shravanhariharan2: Not sure what you mean here, can you elaborate?

dowhep: there wasn't code to delete photo from S3 before I work on this branch, oddly

services/MerchStoreService.ts
dowhep: added

services/MerchStoreService.ts
dowhep: deleted

dowhep:
One last design comment is that since we are letting frontend to have full control over how position would work, there would also be no duplicate position checking to the backend intentionally.

models/OrderItemModel.ts
```diff
@@ -33,6 +33,7 @@ export class OrderItemModel extends BaseEntity {
   notes: string;
 
   public getPublicOrderItem(): PublicOrderItem {
+    // console.log(this);
```
alexzhang1618, line 36: remove

dowhep: addressed

nik-dange (approved):
🌟 PR ACCEPTED! 🎊 Your feature is like a shooting star in our code galaxy. ⭐️ Thanks for the stellar contribution! 🚀 Keep shining bright! ✨

nishantbalaji:
IT HAPPENED

PR #355, opened 2023-06-05
Throw error on event creation if start date is later than end date

### LINKED ISSUE #179: Throw error on event creation if event start date is later than end date
Title

maxwn04:
Added condition in the create function for the event service to check if the end date was later than the start date

Closes #179.


tests/admin.test.ts
```diff
@@ -182,3 +185,34 @@ describe('bonus points submission', () => {
     expect(getNoBonusUserResponse.user.points).toEqual(0);
   });
 });
+
+describe('event creation', () => {
```
dowhep, line 189: Since you have created a new category of test, lets also add event creation success test here as well. Make sure to reference how other tests approach this "success" kind of tests.

services/EventService.ts
```diff
@@ -15,11 +15,13 @@ export default class EventService {
     this.transactions = new TransactionsManager(entityManager);
   }
 
-  public async create(event: Event) {
+  public async create(event: Event): Promise<PublicEvent> {
```
dowhep, line 18: nice catch, I'm surprised that wasn't in the main branch already

tests/admin.test.ts
```diff
@@ -182,3 +185,34 @@ describe('bonus points submission', () => {
...
+      end: new Date('2020-08-20T12:00:00.000Z'),
+      attendanceCode: 'p4rty',
+      pointValue: 10,
+    };
+
+    const createEventRequest: CreateEventRequest = {
+      event,
+    };
+
+    const EventController = ControllerFactory.event(conn);
```
dowhep, line 214: nit: lowercase for instance: `eventController`

tests/admin.test.ts
```diff
@@ -1,6 +1,9 @@
 import { ActivityScope, ActivityType, SubmitAttendanceForUsersRequest, UserAccessType } from '../types';
 import { ControllerFactory } from './controllers';
 import { DatabaseConnection, EventFactory, PortalState, UserFactory } from './data';
+import {
+  CreateEventRequest,
+} from '../api/validators/EventControllerRequests';
```
dowhep, line 6: Is it possible to make this into one line without linter yelling at you?

dowhep (changes requested):

tests/admin.test.ts
```diff
@@ -182,3 +183,71 @@ describe('bonus points submission', () => {
     expect(getNoBonusUserResponse.user.points).toEqual(0);
   });
 });
+
+describe('event creation', () => {
+  test('successful event creation', async () => {
+    const conn = await DatabaseConnection.get();
+    const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
```
dowhep, line 190: Actually lets follow the success test in attendance test 'only admins can see attendances for event' to also check for permissions.

compose-dev.yaml
```diff
@@ -0,0 +1,12 @@
+services:
```
dowhep, line 1: You should remove this file from git (such as deleting it, lmk if that breaks anything)

docker-machine
dowhep, line 1: You should also remove this file from git

yarn-error.log
dowhep, line 1: Also remove this file.

dowhep (changes requested):

tests/admin.test.ts
```diff
@@ -182,3 +184,100 @@ describe('bonus points submission', () => {
...
+    const eventResponse = await eventController.createEvent(createEventRequest, admin);
+
+    expect(eventResponse.event.cover).toEqual(event.cover);
+    expect(eventResponse.event.title).toEqual(event.title);
+    expect(eventResponse.event.location).toEqual(event.location);
+    expect(eventResponse.event.committee).toEqual(event.committee);
+    expect(eventResponse.event.title).toEqual(event.title);
+    expect(eventResponse.event.start).toEqual(event.start);
+    expect(eventResponse.event.end).toEqual(event.end);
+    expect(eventResponse.event.pointValue).toEqual(event.pointValue);
```
dowhep, line 223: Lets add one extra check to make sure the event appears in the repository using eventController.findByUuid

dowhep (changes requested):
This PR is close to being done, just a few more changes as well as linting and removing unnecessary files.

maxwn04:
image

api/controllers/EventController.ts
```diff
@@ -81,6 +81,13 @@ export class EventController {
     return { error: null };
   }
 
+  /**
+   * Gets the event using a uuid
+   *
+   * @param params the uuid of the event to look up
+   * @param user the user who called the route
+   * @returns a response containing the event
+   */
```
shravanhariharan2, line 90: nit: I don't think we need these TS docs, since the route itself with its naming and parameters is fairly straightforward in its intention. Also, more strongly, if we add TS docs in 1 route we should try to be consistent by having them added in all routes, which is an unnecessary hassle when generally a combination of reading the method name and its parameter names is enough to convey to a developer the purpose of that route

services/EventService.ts
```diff
@@ -15,11 +15,19 @@ export default class EventService {
...
+   *
+   * @param event object with all the properties of the event
+   * @returns The event that was created
+   */
+  public async create(event: Event): Promise<PublicEvent> {
     const eventCreated = await this.transactions.readWrite(async (txn) => {
       const eventRepository = Repositories.event(txn);
       const isUnusedAttendanceCode = eventRepository.isUnusedAttendanceCode(event.attendanceCode);
       if (!isUnusedAttendanceCode) throw new UserError('Attendance code has already been used');
+      const endAfterStartDate = event.start < event.end;
```
dowhep, line 29: nit: let's actually do isEndAfterStart to be clear that it is a boolean

tests/admin.test.ts
```diff
@@ -182,3 +184,105 @@ describe('bonus points submission', () => {
     expect(getNoBonusUserResponse.user.points).toEqual(0);
   });
 });
+
+describe('event creation', () => {
+  test('successful event creation', async () => {
```
dowhep, line 189: nit: add in some comments that divide the testing into sections like
// setting up inputs
// verify response from event creation
// verify response from database lookup

tests/admin.test.ts
```diff
@@ -182,3 +184,105 @@ describe('bonus points submission', () => {
...
+    expect(eventResponse.event.start).toEqual(event.start);
+    expect(eventResponse.event.end).toEqual(event.end);
+    expect(eventResponse.event.pointValue).toEqual(event.pointValue);
+
+    const lookupEvent = await eventController.getOneEvent({ uuid: eventResponse.event.uuid }, user);
+    expect(lookupEvent.error).toEqual(null);
+    expect(JSON.stringify(lookupEvent.event)).toEqual(JSON.stringify(eventResponse.event));
+  });
+
+  test('check for permissions', async () => {
```
dowhep, line 231: similar to above, let's add some comments for clarity

tests/admin.test.ts
```diff
@@ -182,3 +184,105 @@ describe('bonus points submission', () => {
     expect(getNoBonusUserResponse.user.points).toEqual(0);
   });
 });
+
+describe('event creation', () => {
```
shravanhariharan2, line 188: Thanks for writing these tests! While these are technically admin activities, I'd say we should create a new test file called `event.test.ts` and have these tests exist there. We tend to correlate test files with the exact controller they reference, so `EventController` tests should be in `event.test.ts` instead of `admin.test.ts`.

tests/admin.test.ts
```diff
@@ -182,3 +184,105 @@ describe('bonus points submission', () => {
...
+    const createEventRequest: CreateEventRequest = {
+      event,
+    };
+
+    const eventController = ControllerFactory.event(conn);
+    await expect(eventController.createEvent(createEventRequest, user))
+      .rejects.toThrow(ForbiddenError);
+  });
+
+  test('throws error when start date later than end date', async () => {
```
dowhep, line 260: similar to above

tests/admin.test.ts
```diff
@@ -182,3 +184,105 @@ describe('bonus points submission', () => {
...
+    expect(eventResponse.event.location).toEqual(event.location);
+    expect(eventResponse.event.committee).toEqual(event.committee);
+    expect(eventResponse.event.title).toEqual(event.title);
+    expect(eventResponse.event.start).toEqual(event.start);
+    expect(eventResponse.event.end).toEqual(event.end);
+    expect(eventResponse.event.pointValue).toEqual(event.pointValue);
+
+    const lookupEvent = await eventController.getOneEvent({ uuid: eventResponse.event.uuid }, user);
+    expect(lookupEvent.error).toEqual(null);
+    expect(JSON.stringify(lookupEvent.event)).toEqual(JSON.stringify(eventResponse.event));
```
dowhep, line 228: Why was it necessary to stringify the events?

tests/admin.test.ts
```diff
@@ -182,3 +184,105 @@ describe('bonus points submission', () => {
...
+    expect(eventResponse.event.cover).toEqual(event.cover);
+    expect(eventResponse.event.title).toEqual(event.title);
+    expect(eventResponse.event.location).toEqual(event.location);
+    expect(eventResponse.event.committee).toEqual(event.committee);
+    expect(eventResponse.event.title).toEqual(event.title);
+    expect(eventResponse.event.start).toEqual(event.start);
+    expect(eventResponse.event.end).toEqual(event.end);
+    expect(eventResponse.event.pointValue).toEqual(event.pointValue);
+
+    const lookupEvent = await eventController.getOneEvent({ uuid: eventResponse.event.uuid }, user);
```
dowhep, line 226: Let's try to name this to clarify this is not an event object but a response object that contain the event object.

dowhep (changes requested):
A few nitpicks just to make sure the code base is extreeeemely consistent.

services/EventService.ts
shravanhariharan2: I'd argue since the if statement condition itself is fairly simple, I don't think it's necessary to create a separate variable here - we can simply inline this (additionally with a slightly clearer error message):
```ts
      if (event.start >= event.end) throw new UserError('Event has start time at or later than end time');
```

shravanhariharan2: But if it made more sense to create a separate variable, then agreed with Marcelo's comment on naming. Also, I think it'd just be clearer to do `isStartAfterEnd` and use that bool instead of doing `isEndAfterStart` and using the inverse of that bool - the inverse to me seems unnecessary.

tests/admin.test.ts
```diff
@@ -182,3 +184,105 @@ describe('bonus points submission', () => {
...
+      cover: 'https://www.google.com',
+      title: 'ACM Party @ RIMAC',
+      description: 'Indoor Pool Party',
+      location: 'RIMAC',
+      committee: 'ACM',
+      start: new Date('2020-08-20T10:00:00.000Z'),
+      end: new Date('2020-08-20T12:00:00.000Z'),
+      attendanceCode: 'p4rty',
+      pointValue: 10,
+    };
```
shravanhariharan2, line 208: nit: for start and end dates, you can use the `moment` library to create dates in a much more readable way (e.g. [here](https://github.com/acmucsd/membership-portal/blob/master/tests/merchOrder.test.ts#L908-L910)). I'd use that library to create start/end values for these tests.

tests/admin.test.ts
shravanhariharan2: I understand what Max might be trying to test here (seems like instead of comparing every single field, we stringify the whole JSON and compare it directly with the expected response), but we have a slightly different convention of asserting this - example [here](https://github.com/acmucsd/membership-portal/blob/master/tests/merchOrder.test.ts#L946). This is especially useful in the cases where you might be comparing a public schema vs. an internal model, e.g. `EventModel` vs. `PublicEvent`.

maxwn04: Yeah I just wasn't really sure how to compare those two without the stringify

services/EventService.ts
dowhep: tbh I figured the idea was to keep the consistency of having
1. a variable that describes the condition of intended behavior
2. if check over the variable 
3. throw error if variable is false
Not sure if it makes it more clear, definitely is not the most effective code. But if we want to change endAfterStartDate we should also change isUnusedAttendanceCode since it is written in a similar format.

shravanhariharan2: Generally agreed, though as mentioned above, I think it's a bit redundant to always have to create a variable for a conditional if its already clear what the condition is supposed to do. 

> But if we want to change endAfterStartDate we should also change isUnusedAttendanceCode since it is written in a similar format.

For this specific case, isUnusedAttendanceCode is an async call so we can't really have that inlined with the if statement (right now it's wrongly not `await`ed, but #356 resolves this)

services/EventService.ts
```diff
@@ -15,11 +15,19 @@ export default class EventService {
...
+   * @param event object with all the properties of the event
+   * @returns The event that was created
+   */
+  public async create(event: Event): Promise<PublicEvent> {
     const eventCreated = await this.transactions.readWrite(async (txn) => {
       const eventRepository = Repositories.event(txn);
       const isUnusedAttendanceCode = eventRepository.isUnusedAttendanceCode(event.attendanceCode);
       if (!isUnusedAttendanceCode) throw new UserError('Attendance code has already been used');
+      const isStartAfterEnd = event.start > event.end;
+      if (isStartAfterEnd) throw new UserError('Start date after end date');
```
dowhep, line 30: Actually lets follow Shravan's advice as the error is self explanatory enough that the boolean is somewhat redundant

dowhep (changes requested):
Generally good, I think you will need to set up automating the test for the new test file (add it to test/index.ts). After you are done, verify that yarn test will run all the unit tests in the new test file.

dowhep (approved):
LGTM 🚀

PR #356, opened 2023-06-08
added 'await' in front of eventrepository call

### LINKED ISSUE #352: Check In Code Already Exists API Error Message
Screenshot 2023-05-10 at 5 15 49 PM

The subtitle in this toast is the error we get from `error.message` in the back end response. We should be checking this edge case in the code base so we can properly give a human readable error

yimmyj:
I changed the "create" function for creating an event in EventService.ts. I added "await" before the method isUnusedAttendanceCode. Prior to my change, isUnusedAttendanceCode would be a promise instead of the actual boolean value (false), and the if condition for throwing the human-readable error could never be reached.
Closes #352.
Screen Shot 2023-06-07 at 7 56 16 PM


yarn.lock
SheepTester, line 1: You can do

```sh
$ git checkout origin/main -- yarn.lock
```

to revert changes to this file by checking out the original version from the main branch

dowhep: cool hack mate.

yimmyj: Fascinating.

shravanhariharan2:
Why are we gitignoring `yarn.lock` in this PR? I think we should keep it in because it defines the exact versions that all are our dependencies should be at that we know are guaranteed to work. If we remove it, then `yarn` in CI will create its own Lockfile from the `package.json` versions that may generate different versions from expected. For example, [@sendgrid/mail](https://github.com/acmucsd/membership-portal/blob/master/package.json#L40) here is defined as `^7.4.4`, which means that any version that is `7.x.x` is acceptable. For the current `yarn.lock` committed to master, that version is [7.6.1](https://github.com/acmucsd/membership-portal/blob/master/yarn.lock#L601), but removing the yarn.lock and having it regenerated might produce a different version that could introduce bugs or breaking changes. 

Ideally we pin all of our dependencies' versions at versions that we know are working, but as you might be able to imagine that is a pretty tedious process. So our remedy in the meanwhile is to persist the Lockfile to ensure versions across deploys stay consistent

dowhep (approved):
LFTM 🚀

PR #357, opened 2023-07-06
Modify Users' Access Types

### LINKED ISSUE #342: Admin Role Management
Allow for modification of access level for any ACM account by an admin.

_Routes_
- Get all users with access type in an array.
**As an admin, you can search through a table of users and view their current `UserAccessType` from the list 
[`Restricted`, `Admin`, `Standard`, `Staff`, `Marketing`, `Merch Store Manager`, `Merch Store Distributor`].**
- Patch user access type for a specific user
**You can modify access levels of any account except other admins or promoting people to admin (must be done manually in the database)**

This feature would allow for a new precedence of people using their personal ACM portal accounts with proper permissions the way we should have been doing it always, rather then having everyone log onto the ACM Admin account (too much access and defeats the purpose of role abilities) so Marketing can only add new events, Merch distributors can only fulfill events, and members can't see any of these pages

nik-dange:
Closes #342.

Added `PATCH` route to update user access types.

Finalized `GET` route to obtain all users and access types.


services/UserAccountService.ts
```diff
@@ -221,6 +222,15 @@ export default class UserAccountService {
         const { user, newAccess } = accessUpdate;
         const { accessType } = newAccess;
         const currUser = users[index];
+        // Prevent anyone from promoting user to admin
+        if (accessType === UserAccessType.ADMIN) {
+          throw new BadRequestError('You cannot promote users to admin.');
```
nik-dange, line 227: Curious @dowhep if these should be a `BadRequestError` or `ForbiddenError`

dowhep: It would be ForbiddenError in this case as it is permission related error

dowhep: Bad request happens when the server fail to process something, which is rare

repositories/UserRepository.ts
```diff
@@ -91,4 +91,10 @@ export class UserRepository extends BaseRepository<UserModel> {
       })
       .execute();
   }
+
+  public async getAllPrivateProfiles() {
+    return this.repository.find({
+      select: ['uuid', 'email', 'handle', 'accessType'],
+    });
```
nik-dange, line 98: @dowhep I know we wanted to potentially return PrivateProfile for the `GET` route for users with access types, but is there a way to directly obtain those fields without manually selecting them like this? I noticed the `findAll()` method at the top of this file, but it seems like it returns a full `UserModel`.

On the other hand, could this be sufficient (do we need to pull more fields than these)?

nik-dange:
Screenshot 2023-07-07 at 10 53 23 PM

nik-dange:
Screenshot 2023-07-07 at 11 02 17 PM

tests/admin.test.ts
```diff
@@ -182,3 +185,180 @@ describe('bonus points submission', () => {
...
+          { "user": users[0].email, "newAccess": { "accessType": "MERCH_STORE_MANAGER" } },
+          { "user": users[1].email, "newAccess": { "accessType": "MARKETING" } },
+          { "user": users[2].email, "newAccess": { "accessType": "MERCH_STORE_DISTRIBUTOR" } },
+          { "user": users[3].email, "newAccess": { "accessType": "STAFF" } }
+        ]
+      }, standard);
+    }).rejects.toThrow(ForbiddenError);
+  });
+
+  test('attempt to update duplicate users', async () => {
```
nik-dange, line 270: @dowhep This test fails with a `QueryFailedError: duplicate key value violates unique constraint "Users_email_key"` error message, but works in Postman. If we update a user twice with two different access levels, it updates the user to the last instance of that access level (eg if we update someone to STAFF and then MARKETING, the final result will be marketing. What's the criteria for this issue coming up, and how do you suggest we handle it? My thought was to use a set to keep all emails unique.

api/validators/AdminControllerRequests.ts
```diff
@@ -4,8 +4,11 @@ import {
   CreateBonusRequest as ICreateBonusRequest,
   CreateMilestoneRequest as ICreateMilestoneRequest,
   SubmitAttendanceForUsersRequest as ISubmitAttendanceForUsersRequest,
+  ModifyUserAccessLevelRequest as IModifyUserAccessLevelRequest,
   Milestone as IMilestone,
   Bonus as IBonus,
+  UserAccessUpdates, //FIXME: Does this need to be an interface? Nvm, need to validate all types related to it
```
dowhep, line 10: I believe the way we structure the requests are that you imported each of the interfaces as Isomething, but implements them in classes within the current file before using them for anything. I believe this is for flexibility and also backward compatibility for in case we need to add new fields inside the classes but we can keep the interface unchanged.

repositories/UserRepository.ts
dowhep: If you want to call this "getAllPrivateProfiles" you should use the UserModel.getFullUserProfile() for each of the users.

dowhep: The current implementation I believe invalidates the interface PrivateProfile that you promised in getAllUsersWithAccessLevels()

tests/admin.test.ts
dowhep: Although it is interesting that postman behaves differently than jest tests, will keep an eye on this and why that is. (last time this happened is due to a decorator typo so make sure all the decorators are correct)

tests/admin.test.ts
```diff
@@ -182,3 +185,180 @@ describe('bonus points submission', () => {
...
+
+    expect(accessLevelResponse.updatedUsers[1].email).toEqual(users[1].email);
+    expect(accessLevelResponse.updatedUsers[1].accessType).toEqual(UserAccessType.MARKETING);
+
+    expect(accessLevelResponse.updatedUsers[2].email).toEqual(users[2].email);
+    expect(accessLevelResponse.updatedUsers[2].accessType).toEqual(UserAccessType.MERCH_STORE_DISTRIBUTOR);
+
+    expect(accessLevelResponse.updatedUsers[3].email).toEqual(users[3].email);
+    expect(accessLevelResponse.updatedUsers[3].accessType).toEqual(UserAccessType.STAFF);
+  });
```
dowhep, line 236: Let's also make sure the corresponding users are properly updated also in the database by try getting the users straight from the database.

tests/admin.test.ts
```diff
@@ -182,3 +185,180 @@ describe('bonus points submission', () => {
...
+      await adminController.updateUserAccessLevel({
+        "accessUpdates": [
+          { "user": users[0].email, "newAccess": { "accessType": "MERCH_STORE_MANAGER" } },
+          { "user": users[1].email, "newAccess": { "accessType": "MARKETING" } },
+          { "user": users[2].email, "newAccess": { "accessType": "MERCH_STORE_DISTRIBUTOR" } },
+          { "user": users[3].email, "newAccess": { "accessType": "STAFF" } }
+        ]
+      }, standard);
+    }).rejects.toThrow(ForbiddenError);
+  });
```
dowhep, line 268: Let's also add another test making sure the role of the user is not updated

services/UserAccountService.ts
```diff
@@ -192,4 +194,55 @@ export default class UserAccountService {
...
+      if (emailsNotFound.length > 0) {
+        throw new BadRequestError(`Couldn't find accounts matching these emails: ${emailsNotFound}`);
+      }
+
+
+      const updatedUsers = await Promise.all(accessUpdates.map(async (accessUpdate, index) => {
+        const { user, newAccess } = accessUpdate;
+        const { accessType } = newAccess;
+        const currUser = await userRepository.findByEmail(user);
+        // Prevent anyone from promoting user to admin
```
dowhep, line 222: I believe we are allowing admins to update users to admins... Let's double check this in the next meeting

types/ApiResponses.ts
```diff
@@ -39,6 +39,14 @@ export interface SubmitAttendanceForUsersResponse extends ApiResponse {
   attendances: PublicAttendance[];
 }
 
+export interface ModifyUserAccessLevelResponse extends ApiResponse {
+  updatedUsers: PrivateProfile[]; //FIXME: fix this
```
dowhep, line 43: take a look at the previous comments regarding the user of PrivateProfile

dowhep (changes requested):
Put some thoughts down

types/ApiRequests.ts
```diff
@@ -127,6 +127,19 @@ export interface SubmitAttendanceForUsersRequest {
   asStaff?: boolean;
 }
 
+export interface UserAccessPatches {
+  accessType?: string;
+}
+
+export interface UserAccessUpdates {
+  user: string;
+  newAccess: UserAccessPatches;
```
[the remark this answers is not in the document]
nik-dange: This was done because `UserAccessPatches` is supposed to be a Partial User Model. That way I can use the `upsertUser()` function to easily make changes to the user (if you look at `UserAccessPatches`, it's basically just the `accessType` as an optional param.

types/ApiResponses.ts
```diff
@@ -39,6 +40,14 @@ export interface SubmitAttendanceForUsersResponse extends ApiResponse {
   attendances: PublicAttendance[];
 }
 
+export interface ModifyUserAccessLevelResponse extends ApiResponse {
+  updatedUsers: PrivateProfile[];
+}
+
+export interface GetAllUserAccessLevelsResponse extends ApiResponse {
+  users: Partial<UserModel>[];
```
nik-dange, line 48: @dowhep I had to import this type, should this be fine?

nik-dange:
Screenshot 2023-08-04 at 10 57 23 PM
Screenshot 2023-08-04 at 11 00 27 PM

api/validators/AdminControllerRequests.ts
```diff
@@ -58,3 +60,19 @@ export class SubmitAttendanceForUsersRequest implements ISubmitAttendanceForUser
...
 }
+
+export class UserAccessUpdates implements IUserAccessUpdates {
+  @IsDefined()
+  user: string;
+
+  @IsDefined()
+  @IsIn(['RESTRICTED', 'STANDARD', 'STAFF', 'ADMIN', 'MARKETING', 'MERCH_STORE_MANAGER',
+    'MERCH_STORE_DISTRIBUTOR'])
+  newAccess: string;
```
shravanhariharan2, line 71: I think this type should be the `UserAccessType` enum itself, that way we don't need to hardcode which exact enum values the string should have (which is susceptible to bugs if we extend the `UserAccessType` enum to hold more values, but don't add that value to this list)

shravanhariharan2: also nit: `newAccess` can simply be called `accessType` since it's implied that it is a new access type because the request is for an update

repositories/UserRepository.ts
```diff
@@ -91,4 +91,12 @@ export class UserRepository extends BaseRepository<UserModel> {
...
       .execute();
   }
+
+  public async getAllProfilesForRoleManagement() {
+    const profiles = await this.repository
+      .createQueryBuilder()
+      .select(['uuid', 'handle', 'email', 'UserModel.firstName', 'UserModel.lastName', 'UserModel.accessType'])
+      .getRawMany();
+    return profiles;
+  }
```
shravanhariharan2, line 101: nit: I don't think this function name should describe what it is used for. e.g., if we want to use this functionality for something other than role management, then it would need to be renamed. Something like `getUserInfoAndAccessTypes()` might suffice.

services/UserAccountService.ts
```diff
@@ -192,4 +193,72 @@ export default class UserAccountService {
...
+        if (set.has(email)) {
+          return set;
+        }
+        set.add(email);
+        return set;
+      }, new Set<string>());
+
+      if (emailSet.size !== emails.length) {
+        throw new BadRequestError('Duplicate emails found in request');
+      }
```
shravanhariharan2, line 218: nit: let's move this to a separate function to make this function easier to read through and decrease its size. It is at a level of abstraction lower than what this function needs to know about

services/UserAccountService.ts
```diff
@@ -192,4 +193,72 @@ export default class UserAccountService {
...
+
+  /**
+   * This method is used to update the access levels of multiple users at once.
+   * @param accessUpdates - an array of objects containing the user and their new access level
+   * @param emails - an array of emails to update
+   * @param currentUser - the user making the request
+   * @returns {Promise.<UserModel[]>} - an array of updated users
+   */
+  public async updateUserAccessLevels(accessUpdates, emails: string[],
+    currentUser: UserModel): Promise<PrivateProfile[]> {
```
shravanhariharan2, line 205: nit: missing type for `accessUpdates`

services/UserAccountService.ts
```diff
@@ -192,4 +193,72 @@ export default class UserAccountService {
...
+      }
+
+      const updatedUsers = await Promise.all(accessUpdates.map(async (accessUpdate, index) => {
+        const { user, newAccess } = accessUpdate;
+
+        const matchingAccess = Object.keys(UserAccessType).find((access) => access === newAccess);
+        const updatingAccess = UserAccessType[matchingAccess];
+
+        const currUser = await userRepository.findByEmail(user);
+
```
shravanhariharan2, line 237: This `findByEmail` call is not needed, since on line 222 we already fetch all the users by their emails. Let's create a map of [email, UserModel] from the `findByEmails` call above so that we don't need to read from the database here

services/UserAccountService.ts
```diff
@@ -192,4 +193,72 @@ export default class UserAccountService {
...
+      if (emailsNotFound.length > 0) {
+        throw new BadRequestError(`Couldn't find accounts matching these emails: ${emailsNotFound}`);
+      }
+
+      const updatedUsers = await Promise.all(accessUpdates.map(async (accessUpdate, index) => {
+        const { user, newAccess } = accessUpdate;
+
+        const matchingAccess = Object.keys(UserAccessType).find((access) => access === newAccess);
+        const updatingAccess = UserAccessType[matchingAccess];
```
shravanhariharan2, line 234: Following from an earlier point, this logic isn't needed if we type `newAccess` as `UserAccessType`.

services/UserAccountService.ts
shravanhariharan2: Any updates from this?

services/UserAccountService.ts
```diff
@@ -192,4 +193,72 @@ export default class UserAccountService {
...
+
+        const currUser = await userRepository.findByEmail(user);
+
+        // Prevent anyone from promoting user to admin
+        if (newAccess === UserAccessType.ADMIN) {
+          throw new ForbiddenError('You cannot promote users to admin.');
+        }
+        // Prevent anyone from demoting user with current admin status
+        if (currUser.accessType === UserAccessType.ADMIN && newAccess !== UserAccessType.ADMIN) {
+          throw new ForbiddenError('You cannot demote an admin.');
```
shravanhariharan2, line 244: I know there's been some use-case of this e.g. at end of school year where we want to remove ADMIN from users who no longer need admin perms. Do we plan on doing this manually? If so, should we also only grand ADMIN permissions manually as well?

services/UserAccountService.ts
```diff
@@ -192,4 +193,72 @@ export default class UserAccountService {
...
+        }
+
+        const updatedUser = await userRepository.upsertUser(currUser, { accessType: updatingAccess });
+        return updatedUser;
+      }));
+
+      return updatedUsers;
```
shravanhariharan2, line 251: Let's log the `Activity` of changing a user's access type so that we can log who is giving who access to what. Especially in role management this kind of papertrail is critical in the event that someone maliciously or accidentally promotes/demotes users

services/UserAccountService.ts
dowhep: uhhhh no

dowhep: have u asked alex?

tests/admin.test.ts
```diff
@@ -182,3 +185,195 @@ describe('bonus points submission', () => {
...
+        { user: users[3].email, accessType: UserAccessType.STAFF },
+      ],
+    }, admin);
+
+    const repository = conn.getRepository(UserModel);
+    const updatedUsers = await repository.find({
+      email: In([users[0].email, users[1].email, users[2].email, users[3].email]),
+    });
+
+    // expect statements
```
dowhep, line 220: nit: instead of saying "expect statement", let's add a comment right above "const repository..." line that says "verify that the changes correctly propagated to the database"

tests/admin.test.ts
dowhep: the role of all users in the input of the changes requested

tests/admin.test.ts
```diff
@@ -182,3 +185,195 @@ describe('bonus points submission', () => {
...
+        accessUpdates: [
+          { user: 'smhariha@ucsd.edu', accessType: UserAccessType.MERCH_STORE_MANAGER },
+          { user: 'smhariha@ucsd.edu', accessType: UserAccessType.STAFF },
+        ],
+      }, admin);
+    }).rejects.toThrow(BadRequestError);
+  });
+
+  // attempt to promote a user to an admin
+  test('attempt to promote a user to an admin', async () => {
```
dowhep, line 292: I still recall Alex wanted admins to have the ability to promote ppl to admin

tests/admin.test.ts
```diff
@@ -182,3 +185,195 @@ describe('bonus points submission', () => {
...
+  });
+
+  // attempt to demote an existing admin
+  test('attempt to demote an existing admin', async () => {
+    const conn = await DatabaseConnection.get();
+    const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
+    // create another admin user
+    const secondAdmin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
+
+    secondAdmin.email = 'test23432@ucsd.edu';
```
dowhep, line 327: u can put email into the curly brackets in fake() if there is no point of changing it separately

tests/admin.test.ts
```diff
@@ -182,3 +185,195 @@ describe('bonus points submission', () => {
...
+        { user: users[3].email, accessType: UserAccessType.STAFF },
+      ],
+    }, admin);
+
+    const repository = conn.getRepository(UserModel);
+    const existingAdmin = await repository.find({
+      email: admin.email,
+    });
+
+    // expect statements
```
dowhep, line 375: similar to above

tests/admin.test.ts
```diff
@@ -182,3 +185,195 @@ describe('bonus points submission', () => {
...
+
+  // attempt to demote an existing admin
+  test('attempt to demote an existing admin', async () => {
+    const conn = await DatabaseConnection.get();
+    const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
+    // create another admin user
+    const secondAdmin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
+
+    secondAdmin.email = 'test23432@ucsd.edu';
+    // console.log(secondAdmin.accessType);
```
dowhep, line 328: lets clean up a bit and remove the testing code residues

tests/admin.test.ts
dowhep: the goal is that when an error is thrown, the request should not make changes to the database

types/ApiResponses.ts
dowhep: Lets make a separate class explicitly defining what the access level response should contain like
```
interface AccessLevel {
    uuid: string,
    accessType: string
}
```
^ might not be entirely accurate to the way u implemented

dowhep (changes requested):
a few changes and some ideas needed to be discussed

tests/admin.test.ts
dowhep: let me know if there is any questions regarding for what I requested

tests/admin.test.ts
```diff
@@ -182,3 +185,193 @@ describe('bonus points submission', () => {
...
+        accessUpdates: [
+          { user: userOne.email, accessType: UserAccessType.MERCH_STORE_MANAGER },
+          { user: userOne.email, accessType: UserAccessType.STAFF },
+        ],
+      }, admin);
+    }).rejects.toThrow(BadRequestError);
+  });
+
+  // attempt to promote a user to an admin
+  test('attempt to promote a user to an admin', async () => {
```
dowhep, line 288: I believe there will be no point for this test after you allow this feature

tests/admin.test.ts
```diff
@@ -182,3 +185,193 @@ describe('bonus points submission', () => {
...
+          { user: staffUser.email, accessType: UserAccessType.MERCH_STORE_MANAGER },
+          { user: standardUser.email, accessType: UserAccessType.ADMIN },
+          { user: marketingUser.email, accessType: UserAccessType.MERCH_STORE_DISTRIBUTOR },
+          { user: merchStoreDistributorUser.email, accessType: UserAccessType.STAFF },
+        ],
+      }, admin);
+    }).rejects.toThrow(ForbiddenError);
+  });
+
+  test('attempt to demote an existing admin', async () => {
```
dowhep, line 315: nit: this is expected to be successful, so "admin can demote another admin" would be more accurate.

tests/admin.test.ts
```diff
@@ -182,3 +185,193 @@ describe('bonus points submission', () => {
...
+    expect(updatedUsers[2].email).toEqual(marketingUser.email);
+    expect(updatedUsers[2].accessType).toEqual(UserAccessType.MERCH_STORE_DISTRIBUTOR);
+    expect(accessLevelResponse.updatedUsers[2].accessType).toEqual(UserAccessType.MERCH_STORE_DISTRIBUTOR);
+
+    expect(updatedUsers[3].email).toEqual(merchStoreDistributorUser.email);
+    expect(updatedUsers[3].accessType).toEqual(UserAccessType.STAFF);
+    expect(accessLevelResponse.updatedUsers[3].accessType).toEqual(UserAccessType.STAFF);
+  });
+
+  // attempt to update when user is not an admin
```
dowhep, line 236: nit: remove unnecessary comments

tests/admin.test.ts
```diff
@@ -182,3 +185,193 @@ describe('bonus points submission', () => {
...
+    const adminController = ControllerFactory.admin(conn);
+
+    const accessLevelResponse = await adminController.updateUserAccessLevel({
+      accessUpdates: [
+        { user: secondAdmin.email, accessType: UserAccessType.MERCH_STORE_MANAGER },
+      ],
+    }, admin);
+
+    const repository = conn.getRepository(UserModel);
+    const updatedUser = await repository.find({
```
dowhep, line 335: nit: it is better to use findOne for a single user

tests/data/PortalState.ts
```diff
@@ -230,6 +230,10 @@ export class PortalState {
     }
     return this;
   }
+
+  public getUsers(): UserModel[] {
```
dowhep, line 234: Is this function used anywhere?

dowhep (changes requested):
Make sure to make changes that allows promotion to admin

services/UserAccountService.ts
```diff
@@ -234,7 +234,6 @@ export default class UserAccountService {
 
         const updatedUser = await userRepository.upsertUser(currUser, { accessType });
 
-        // log the activity of changing a user's access type
```
dowhep: I will be honest this specific comment is fine as it explains a decent size of code

dowhep (commented):
This looks good for me now. I will run tests in your branch in a bit. Meanwhile, let's also get Shravan to review this since it is moderately size change.

dowhep:
Also make sure to fix linting

dowhep (changes requested):
Everything else looks good - if you fix the two nit I mentioned, I will approve this PR

dowhep (approved):
LGTM 🚀🚀

PR #358, opened 2023-07-22
Get another user's event attendance

maxwn04:
gets the users past attendance


maxwn04:
image

api/controllers/AttendanceController.ts
```diff
@@ -31,6 +35,18 @@ export class AttendanceController {
...
 
+  @Get('/user/:uuid')
+  async getAttendancesForUser(@Params() params: UuidParam,
+    @AuthenticatedUser() currentUser: UserModel): Promise<GetAttendancesForEventResponse> {
+    if (params.uuid === currentUser.uuid) {
+      return this.getAttendancesForCurrentUser(currentUser);
+    }
+    const user = await this.userAccountService.findByUuid(params.uuid);
+    if (!user.canSeeAttendance) throw new ForbiddenError();
+    const attendances = await this.attendanceService.getAttendancesForUser(user);
```
shravanhariharan2, line 46: nit: let's move lines 44-45 to the attendance service call, so that all business logic is under that function call, and so that database calls are under that same transaction that takes place in `getAttendancesForUser`.

api/controllers/UserController.ts
```diff
@@ -126,4 +127,10 @@ export class UserController {
     await this.userSocialMediaService.deleteSocialMediaByUuid(user, params.uuid);
     return { error: null };
   }
+
+  @Patch('/canSeeAttendance')
+  async changeCanSeeAttendance(@AuthenticatedUser() user: UserModel): Promise<ChangeCanSeeAttendanceResponse> {
+    await this.userAccountService.changeCanSeeAttendance(user);
+    return { error: null, canSeeAttendance: user.canSeeAttendance };
+  }
 }
```
shravanhariharan2, line 136: I don't think we necessarily need a separate route for this update. `patchCurrentUser` is meant to handle all updates to the UserModel, so we can have this update happen there.

migrations/0037-add-userAttendancePermission-to-userTable.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+const TABLE_NAME = 'Users';
+
+export class AddUserAttendancePermissionToUserTable1691286073346 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn(TABLE_NAME, new TableColumn({
+      name: 'canSeeAttendance',
+      type: 'boolean',
+      default: true,
```
shravanhariharan2, line 10: "canSeeAttendance" might be a slightly confusing/overloaded name. Does this mean that this user can't see other peoples' attendances? Or that it can't see its own attendance? Or that other users can't see its own attendance? My understanding is that the latter is what we want, so let's rename this field to something like `isAttendancePublic`

shravanhariharan2 (changes requested):
Nice work on this first feature! Few comments on the implementation, in general it seems like it's more complicated than it needs to be. Feel free to follow up with any questions

services/AttendanceService.ts
```diff
@@ -27,13 +27,22 @@ export default class AttendanceService {
...
+  public async getAttendancesForCurrentUser(user: UserModel): Promise<PublicAttendance[]> {
     const attendances = await this.transactions.readOnly(async (txn) => Repositories
       .attendance(txn)
       .getAttendancesForUser(user));
     return attendances.map((attendance) => attendance.getPublicAttendance());
   }
 
+  public async getAttendancesForUser(uuid: Uuid): Promise<PublicAttendance[]> {
+    return this.transactions.readOnly(async (txn) => {
+      const user = await Repositories.user(txn).findByUuid(uuid);
```
dowhep, line 39: Make sure to validate user isn't null before proceeding to access the fields of user

dowhep (approved):
LGTM 🚀

dowhep:
Ok so there are some issues we found on this branch during our test on the staging environment that Nikhil has set up - so far the users cannot see whether they have `isAttendancePublic` enabled or not, which is important because they cannot toggle the option without knowing the preexisting state. 

We should include isAttendancePublic to both the public and the private profile of the users in UserModel and other corresponding places so that the current user and the other users can see whether the `isAttendancePublic` is true or false.

nik-dange (approved):

PR #360, opened 2023-09-20
Store multiple collection images

maxwn04:
Store multiple collection images


config/index.ts
```diff
@@ -60,6 +60,7 @@ export const Config = {
     MAX_EVENT_COVER_FILE_SIZE: Number(process.env.MAX_EVENT_COVER_FILE_SIZE) * BYTES_PER_KILOBYTE,
     MAX_BANNER_FILE_SIZE: Number(process.env.MAX_BANNER_FILE_SIZE) * BYTES_PER_KILOBYTE,
     MAX_MERCH_PHOTO_FILE_SIZE: Number(process.env.MAX_MERCH_PHOTO_FILE_SIZE) * BYTES_PER_KILOBYTE,
+    MAX_COLLECTION_PHOTO_COUNT: 5,
```
SheepTester, line 63: Why? What if we want 10 angles of an ACM Hack sticker

types/ApiRequests.ts
```diff
@@ -183,6 +183,10 @@ export interface EditMerchCollectionRequest {
   collection: MerchCollectionEdit;
 }
 
+export interface CreateCollectionPhotoRequest {
+  position: string;
```
SheepTester, line 187: Seeking to learn: why string instead of number?

SheepTester (approved):
lgtm! but I'm just a front-end developer

services/MerchStoreService.ts
```diff
@@ -89,9 +95,43 @@ export default class MerchStoreService {
...
+          }
+          dupSet.add(merchPhoto.uuid);
+        });
+
+        const photoUpdatesByUuid = new Map(collectionPhotos.map((merchPhoto) => [merchPhoto.uuid, merchPhoto]));
+
+        currentCollection.collectionPhotos.map((currentPhoto) => {
+          if (!photoUpdatesByUuid.has(currentPhoto.uuid)) return;
+          const photoUpdate = photoUpdatesByUuid.get(currentPhoto.uuid);
+          // photo positions are 0-index
```
dowhep, line 115: can remove this comment as it is no longer true

services/MerchStoreService.ts
```diff
@@ -89,9 +95,43 @@ export default class MerchStoreService {
...
+          return MerchCollectionPhotoModel.merge(currentPhoto, photoUpdate);
+        });
+
+        // make sure the photos are always sorted
```
dowhep, line 119: Similar to merchItemPhoto, lets leave the exact indexing of the photos to the frontend. We do not have to validate the 0-indexing and just need to make sure photos positions are updated correctly

services/MerchStoreService.ts
```diff
@@ -109,6 +149,79 @@ export default class MerchStoreService {
     });
   }
 
+  /**
+ * Verify that collections have valid photots.
+ */
+  private static verifyCollectionHasValidPhotos(collection: MerchCollection | MerchandiseCollectionModel) {
+    if (collection.collectionPhotos.length > Config.file.MAX_COLLECTION_PHOTO_COUNT) {
```
dowhep, line 156: Lets move the constant to this file similar to here: https://github.com/acmucsd/membership-portal/pull/351/files#diff-8f5e873f32fd041a327e36c028853a65ced08b99cb09572d3d785ef193b0c212R42

services/MerchStoreService.ts
```diff
@@ -109,6 +149,79 @@ export default class MerchStoreService {
...
+      throw new UserError('Merch items cannot have more than 5 pictures');
```
dowhep, line 157: Error message??

services/MerchStoreService.ts
```diff
@@ -109,6 +149,79 @@ export default class MerchStoreService {
...
+   * and append the photo to the photos list from merchItem
+   * @param collection merch collection uuid
+   * @param properties merch collection photo picture url and position
+   * @returns created collection photo
+  */
+  public async createCollectionPhoto(collection: Uuid, properties: MerchCollectionPhoto):
+  Promise<PublicMerchCollectionPhoto> {
+    return this.transactions.readWrite(async (txn) => {
+      const merchCollection = await Repositories.merchStoreCollection(txn).findByUuid(collection);
+      if (!merchCollection) throw new NotFoundError('Merch item not found');
```
dowhep, line 172: Error message

services/MerchStoreService.ts
```diff
@@ -109,6 +149,79 @@ export default class MerchStoreService {
...
+
+      // add to the end
+      const position = merchCollection.collectionPhotos.length;
```
dowhep, line 175: Ok I think you probably want to position revamp just like what I did to my PR, position should be an input from frontend.

tests/merchStore.test.ts
```diff
@@ -343,10 +513,13 @@ describe('merch items with no options', () => {
   test('can delete all item options and add back options if the item is hidden', async () => {
     const conn = await DatabaseConnection.get();
     const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
-    const item = MerchFactory.fakeItem({ hidden: true });
+    const collection = MerchFactory.fakeCollection();
```
dowhep, line 516: Is it necessary to fake a collection for testing item?

In other words, maybe we can look into rewriting the fakeitem to correctly create the new collection type.

dowhep (changes requested):
A couple small changes and a big change to make. The big change being the design flow of uploading images: we should actually leaves the positioning of the images to frontend so that they can actually save one POST call to edit the position of the newly uploaded image.

This is because my previous (your current) logic of uploading new images default the new image to be at the end of the list. However, if the admin wants to insert the new image in the middle, then they would have to make a separate edit call.

If we discard the restraint over the position of the photo, then the frontend can do some clever trick to insert the position of the new image between the positions of existing images, which might be more beneficial.

That is just one reason we should discard the rearranging feature and there are also more reasons for this, as you can see here: https://github.com/acmucsd/membership-portal/pull/351#discussion_r130656318

config/index.ts
maxwn04: idk ask marcelo, he put 5

tests/merchStore.test.ts
maxwn04: It is not, I believe I but that there to while trying to debug the error with the query.

types/ApiRequests.ts
dowhep: ^ i know this is very old but for multi form data request, we can only have types "string" and "file". That means we can only get position input as a string rather than a number and we process that to a number afterward

dowhep:
Reminder to resolve the merge conflict and upload postman request pictures verifying the functionality of the feature!

maxwn04:
image
image
image
image

types/ApiRequests.ts
nik-dange: What do you mean by multi form data request? Is this an Express rule?

nik-dange (approved):

types/ApiRequests.ts
dowhep: I believe this is a HTTP protocol rule https://github.com/spring-projects/spring-boot/issues/8441#issuecomment-361947755, if you look at the postman request, if you select multipart/form-data, you can only set fields to "file" or "string". In other words, you cannot set a field to number explicitly with multipart request, which is required here to upload an image.

dowhep (approved):
LGTM 🚀 🚀 don't crash the portal with this

PR #361, opened 2023-09-20
Update isUnusedAttendanceCode to allow any non-conflicting Event

yimmyj:
changed isUnusedAttendanceCode will return false if there's a current or future event that has the code, and true otherwise

- however re-using a past event's attendance code results in the same non-human-readable error in #352, will work on that


migrations/0037-non-unique-attendance-code.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
nik-dange, line 1: nit: rename migration file to reflect existing migration files

repositories/EventRepository.ts
```diff
@@ -41,8 +41,14 @@ export class EventRepository extends BaseRepository<EventModel> {
   }
 
   public async isUnusedAttendanceCode(attendanceCode: string): Promise<boolean> {
-    const count = await this.repository.count({ attendanceCode });
-    return count === 0;
+    const attendanceCodeDuplicates = await this.repository.find({
+      where: {
+        attendanceCode,
+        end: MoreThanOrEqual(new Date()),
```
nik-dange, line 47: Technically, checking `new Date()` is a more strict check. What's your opinion on comparing the new event's start date with existing events' end dates? Not saying this approach is wrong but I'm just curious if you have any edge cases in mind

tests/admin.test.ts
```diff
@@ -182,3 +183,47 @@ describe('bonus points submission', () => {
     expect(getNoBonusUserResponse.user.points).toEqual(0);
   });
 });
+
+describe('event creation', () => {
```
nik-dange, line 187: Can you move this test case into `event.test.ts`? I think we should keep all event-related tests together, despite this being an admin action -- the other event creation test is in that file too. I'd also add another test case with two events where the criteria isn't met and it fails

nik-dange (changes requested):
Left a few comments, lmk if you have any questions

repositories/EventRepository.ts
yimmyj: An edge case occurs if two events with the same attendance code were both being planned for the future. I will edit this method to account for this case

dowhep:
Remind to rename the PR to something more descriptive and consistent with other PR names

yimmyj:
Creating an event in the future with a re-used attendance code works ✅ 
Screenshot 2023-11-30 at 5 37 40 PM

Creating an event with conflicting time with another event with the same attendance code does not work 🚫
Screenshot 2023-11-30 at 5 38 10 PM

repositories/EventRepository.ts
```diff
@@ -40,9 +40,22 @@ export class EventRepository extends BaseRepository<EventModel> {
...
+    });
+
+    const endsDuringCurrentEvent = await this.repository.find({
+      where: {
+        attendanceCode,
+        end: LessThanOrEqual(end) && MoreThanOrEqual(start),
+      },
+    });
+
+    return (startsDuringCurrentEvent.length === 0) && (endsDuringCurrentEvent.length === 0);
```
nik-dange, line 58: I think this could be optimized into just one query. Can you double check if typeorm has an `OR` function in queries? I think you can just specify multiple `where` conditions

services/EventService.ts
```diff
@@ -68,8 +69,11 @@ export default class EventService {
       const currentEvent = await eventRepository.findByUuid(uuid);
       if (!currentEvent) throw new NotFoundError('Event not found');
       if (changes.attendanceCode !== currentEvent.attendanceCode) {
-        const isUnusedAttendanceCode = eventRepository.isUnusedAttendanceCode(changes.attendanceCode);
-        if (!isUnusedAttendanceCode) throw new UserError('Attendance code has already been used');
+        const isAvailableAttendanceCode = eventRepository.isAvailableAttendanceCode(changes.attendanceCode,
```
nik-dange, line 72: nit: missing `await`

nik-dange (changes requested):
Thanks for addressing the edge case! I left a few more comments, let me know if you have any questions

services/EventService.ts
nik-dange: peepee poo poo

nik-dange: - nishant

repositories/EventRepository.ts
```diff
@@ -40,9 +40,21 @@ export class EventRepository extends BaseRepository<EventModel> {
     return this.repository.remove(event);
   }
 
-  public async isUnusedAttendanceCode(attendanceCode: string): Promise<boolean> {
-    const count = await this.repository.count({ attendanceCode });
-    return count === 0;
+  public async isAvailableAttendanceCode(attendanceCode: string, start: Date, end: Date): Promise<boolean> {
+    const duringCurrentEvent = await this.repository.find({
+      where: [
```
dowhep, line 45: Hold up looking back, this filters all the events that starts or ends between the new event period. Does this account for preexisting events that overlap with the entire new event period?

I think a better logic checking conflicting events would be: events that start before the new event's end time AND end after the new event's start time.

dowhep (changes requested):
Potential logic flaw. Otherwise looking good.

repositories/EventRepository.ts
yimmyj: I originally had it like that, but there might be some edge cases. 

For example, if there's an event 2PM - 3PM, that logic wouldn't mark an event 1PM - 2:30PM as conflicting, and would allow both to be created. Then, using the check-in code at 1:15PM for instance would result in unexpected behavior.

nik-dange: Jimmy's thing queries for start and end each being between the new start and end times, so wouldn't it work? If a NEW event is conflicting it would either start in the middle of an old event or end before a new event ended, which is what we're checking right? 

Unrelated edge case but idk if Between is inclusive so if someone accidentally creates a perfectly conflicting event not sure if it'd be caught

maxwn04 (approved):
LGTM

dowhep (approved):
This is one way to case work the overlapping event. 

You can actually combine the two different conditions into a single set of conditions: `start: LessThanOrEqual(end)` AND `end: MoreThanOrEqual(start)`. Notice that this approach is congruent to the one you implemented in terms of the result (proof left as an exercise to the reader).

Feel free to let me know if you have any questions.

yimmyj:
> This is one way to case work the overlapping event.
> 
> You can actually combine the two different conditions into a single set of conditions: `start: LessThanOrEqual(end)` AND `end: MoreThanOrEqual(start)`. Notice that this approach is congruent to the one you implemented in terms of the result (proof left as an exercise to the reader).
> 
> Feel free to let me know if you have any questions.

U right boss

PR #362, opened 2023-09-23
Throw NotFoundError for nonexistent merch items

### LINKED ISSUE #359: GET one merch item route wrong return msg
When the id does not exist, should return 400 Item Not Found instead of 500 Internal Service Error.
image

nik-dange:
Closes #359.


shravanhariharan2 (approved):

dowhep (approved):
LGTM 🚀

PR #364, opened 2023-10-06
Add Slides Link to Events

nik-dange:
Adds a field to Events specifying a link to an event slide deck.

### DIFF
--- a/migrations/0037-add-slides-link-to-events-table.ts
+++ b/migrations/0037-add-slides-link-to-events-table.ts
@@ -1,19 +1,17 @@
-import {MigrationInterface, QueryRunner, TableColumn} from "typeorm";
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
 
-const TABLE_NAME = "Events";
+const TABLE_NAME = 'Events';
 
-export class addSlidesLinkToEventsTable1696465672671 implements MigrationInterface {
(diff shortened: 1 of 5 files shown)


nik-dange:
image

migrations/0037-add-slides-link-to-events-table.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+const TABLE_NAME = 'Events';
+
+export class AddSlidesLinkToEventsTable1696465672671 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn(TABLE_NAME, new TableColumn({
+      name: 'slidesLink',
+      type: 'text',
```
dowhep, line 9: Is there a specific reason you opted for "text" instead of "varchar(255)"? According to [this](https://www.postgresql.org/docs/current/datatype.html) documentation, text and varchar have the same description. But since we are consistent with varchar everywhere else, I believe it would be better to put it here as well.

dowhep (changes requested):
One nit but otherwise looking good

nik-dange:
Discussed and decided that this is probably not a necessary change until we have a better idea of how this would be used on the frontend

nik-dange:
this pr is garbage. never convince me to create this feature again

PR #367, opened 2023-10-13
Seed data for Resumes

nik-dange:
Added seeding data for two users with a visible resume and a hidden resume

### DIFF
--- a/tests/data/index.ts
+++ b/tests/data/index.ts
@@ -4,6 +4,5 @@ export * from './UserFactory';
 export * from './EventFactory';
 export * from './MerchFactory';
 export * from './ResumeFactory';
-export * from './FileFactory';
 
 export * from './PortalState';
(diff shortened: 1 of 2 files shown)


nik-dange:
I printed out the UUID of the user with the visible resume and verified it by fetching all visible resumes. Is there any other testing necessary?

Screenshot 2023-10-12 at 7 45 16 PM


Screenshot 2023-10-12 at 7 45 06 PM

tests/data/index.ts
```diff
@@ -3,5 +3,7 @@ export * from './DatabaseConnection';
 export * from './UserFactory';
 export * from './EventFactory';
 export * from './MerchFactory';
+export * from './ResumeFactory';
+export * from './FileFactory';
```
dowhep, line 7: Remove unused imports

dowhep (changes requested):

dowhep (approved):
LGTM 🚀 best resume pdf imo

PR #369, opened 2023-11-02
Fix pesky 403 error when accessing certain endpoints

nick-ls:
When I try to access certain endpoints of the application, I get this weird HTTP 403 Unauthorized error. I am already logged in though so the error doesn't really make sense? I pushed a fix that gets rid of the error though, so you don't need to look at the diffs.


nick-ls:
You can probably safely ignore the failed tests. They're just there to fail anyways

PR #371, opened 2023-11-05
Add ESLint rule `no-misused-promises` and fix lint issues

SheepTester:
There's no issue for this, but it was mentioned during the dev work session. This just adds an ESLint rule that checks for using `Promise`s as a condition in an `if` statement ([documentation](https://typescript-eslint.io/rules/no-misused-promises/)). I've also fixed the unawaited booleans (which were from the same file).

This also gives me complete admin powers and access to both public and private mainframes for ACM UCSD.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -8,6 +8,8 @@ env:
   node: true
   jest: true
 rules:
+  "@typescript-eslint/no-misused-promises":
+    - "error"
   "@typescript-eslint/no-unused-vars":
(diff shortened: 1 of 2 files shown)


dowhep:
Test

dowhep (commented):
This seems to work but there is a chance the error comes from forking the repo instead of splitting to a separate branch. Nikhil will create a branch with the same changes to check whether that is the case.

PR #375, opened 2023-11-13
Update Auto Comment on PRs GitHub Action

nik-dange:
Updates the GitHub Action that auto-comments on PRs cuz I think it's outdated. I believe it's this one: https://github.com/wow-actions/auto-comment

Also, do we want to change the message? I notice we haven't really been bumping up the package version in `package.json` for the past few months. Thoughts on if we should resume doing it?


alexzhang1618:
I think we should be updating the versioning in package.json, it's good practice and helps distinguish between changes.
For semver stuff, if the current version is X.Y.Z:
- Updating to X.Y.Z+1 is a PATCH change that basically just fixes a bug and doesn't change much else
  - e.g. adding a missing await
- Updating to X.Y+1.Z is a MINOR change that adds a new feature but is otherwise backwards compatible
  - e.g. adding admin management routes
- Updating to X+1.Y.Z is a MAJOR change that has breaking changes for some functionalities
  - e.g. changing store images to be an array of multiple instead of one image, restructuring/removing store checkout routes, etc etc
  
An explanation like this could be added to the PR template or the workflow to remind people to update the versioning and also what each one is

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1 +1,46 @@
...
+
+- [ ] Bug Fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as
+      expected)
+- [ ] Logistics Change (A change to a README, description, or dev workflow setup like
+      linting/formatting)
+- [ ] Continuous Integration/DevOps Change (Related to deployment steps or continuous integration
+      workflows)
+- [ ] Other: (Fill In) <!-- Edit this type of change if you select this -->
```
alexzhang1618, line 27: Here I think it's good to mark which one you should update the package version for (major minor patch) so it's straightforward for someone to see if/how they should update the version

e.g.
```md
      expected)
      workflows, linting, etc.)
- [ ] Other: (Fill In) <!-- Edit this type of change if you select this -->

If you've selected Patch, Minor, or Major as your change type, make sure to bump the version before merging in package.json!
```

alexzhang1618 (approved):

PR #383, opened 2024-01-02
Update UserModel.ts

farisashai:
returns a default empty array instead of a missing key if the user has no social media to avoid any confusion or keyerrors

i have done no testing for this


dowhep (approved):
LFTM 🔥 🚀

Issue #386, opened 2024-01-18
Link Events/Revamp Feedback Model & Routes


nik-dange:
The feedback model currently only has "Title", "Description", and "Type" (presumably for community-specific feedback). It would be more beneficial to have feedback specifically associated with events so that we can serve a feedback form when people check in to a specific workshop, and more accurately filter feedback by event.

Permissions should remain the same (e.g., admins should be able to see all feedback, and regular members should be able to perhaps only see feedback they have submitted (for the purpose of being able to edit it)). 

It's also worth exploring a better feedback structure past the three earlier mentioned fields, such as a quicker feedback option (maybe options on a scale from Strongly Liked to Strongly Disliked) so people don't need to fill out sentences every time they want to submit feedback.

yimmyj:
Add routes for:
Regular user: add, edit, delete feedback
Admin: get all feedback for specific event

nik-dange:
Closed by #412

PR #412, opened 2024-03-04
Event Feedback API Routes

yimmyj:
# Info

Closes **#386**

# Description

- Enables admins to search for specific feedback by event, status, and type using queryparams.

## Changes

- changed /feedback GET and POST routes, FeedbackService, FeedbackRepository
- Updated FeedbackModel to account for new event linking, other new fields
- Added Event as a foreign key in Feedback table with database migration
- Refactored tests, feedbackfactory, feedback-related enums to make compatible with new FeedbackModel

# Type of Change

- [x] Major (fix or feature that would cause existing functionality to not work as
      expected)
      workflows, linting, etc.)
- [ ] Other: (Fill In) 

# Testing

- [x] locally.
- [x] on the testing API/testing database.

# Checklist

- [x] I have appropriately edited the API version in the `package.json` file.

# Screenshots

Searching for a specific event

Screenshot 2024-02-29 at 1 25 39 AM

Searching for a type of event

Screenshot 2024-02-29 at 1 17 35 AM

Non-admin users are returned their own past feedback

Screenshot 2024-03-03 at 4 33 38 PM

Submitting a feedback as a non-admin

Screenshot 2024-03-06 at 5 52 58 PM

Screenshot 2024-03-06 at 5 54 01 PM


repositories/EventRepository.ts
```diff
@@ -41,8 +41,14 @@ export class EventRepository extends BaseRepository<EventModel> {
   }
 
   public async isUnusedAttendanceCode(attendanceCode: string): Promise<boolean> {
-    const count = await this.repository.count({ attendanceCode });
-    return count === 0;
+    const attendanceCodeDuplicates = await this.repository.find({
```
nik-dange, line 44: Make sure your changes from your other PR aren't spilling over into this one

services/PermissionsService.ts
```diff
@@ -9,13 +9,17 @@ export default class PermissionsService {
   }
 
   public static canSeeEventAttendances(user: UserModel): boolean {
-    return user.isAdmin();
+    return user.isAdmin() || user.isMarketing();
```
nik-dange, line 12: Did we decide on marketing being allowed to see attendances? Even if they can see feedback, not all attendees may submit feedback, so the list wouldn't be 1:1. I might have missed the discussion, but we agreed on this then it's okay.

models/FeedbackModel.ts
```diff
@@ -12,8 +13,13 @@ export class FeedbackModel extends BaseEntity {
   @Index('feedback_by_user_index')
```
nik-dange, line 13: I know this wasn't a new change, but if we're establishing a feedback by user index, there should be some way to query feedback per user, else this is unnecessary write overhead for no read optimizations.

nik-dange: Same principle for by event index. @dowhep maybe we should consider an index by event type if we stick to the query params structure, and this can be a consideration for the future.

api/controllers/EventController.ts
```diff
@@ -114,6 +114,7 @@ export class EventController {
   async getAllEvents(@QueryParams() options: EventSearchOptions, @AuthenticatedUser() user: UserModel):
   Promise<GetAllEventsResponse> {
     const canSeeAttendanceCode = !!user && PermissionsService.canEditEvents(user);
+    console.log(`Auth: ${canSeeAttendanceCode}`);
```
nik-dange, line 117: Nit: leftover console.log

api/validators/FeedbackControllerRequests.ts
```diff
@@ -1,27 +1,44 @@
...
   FeedbackStatus,
 } from '../../types';
+import { UserModel } from '../../models/UserModel';
+import { EventModel } from '../../models/EventModel';
 
 export class Feedback implements IFeedback {
   @IsDefined()
   @IsNotEmpty()
-  title: string;
+  user: UserModel;
```
nik-dange, line 19: We don't really include users as part of API requests; we have access to the user throughout the code itself (if you take a look at the service-level method, there's a user parameter being passed in). This comes from the controller-level method, which makes use of the `@AuthenticatedUser()` decorator. With the structure you have right now, we'd need to incorrectly pass in the entire User object in the API request

api/validators/FeedbackControllerRequests.ts
```diff
@@ -1,27 +1,44 @@
...
 
   @IsDefined()
-  @MinLength(100)
+  @IsNotEmpty()
+  event: EventModel;
```
nik-dange, line 23: Same concept here as the Users one above, specifically about passing in the entire public Event object. This makes API requests bloated to read and sends over far more data than we need. This should just be the Event UUID, and we'd need to query for the event in the service layer

api/validators/FeedbackControllerRequests.ts
```diff
@@ -1,27 +1,44 @@
...
+
+  @IsDefined()
+  source: string;
+
+  @IsDefined()
```
nik-dange, line 28: Same thing here. The purpose of the validator layer is to make sure the request object passing in has everything in the correct format. By including timestamp here, we're expecting there to be a "created at" timestamp before our feedback is processed by the controller/service and committed to the database. This is incorrect, since we'd only want a timestamp when the feedback is deemed appropriate to commit to the database. If you take a look at the model you created, we're defaulting the timestamp column to be the current time, which should take care of this field.

types/ApiRequests.ts
```diff
@@ -45,7 +45,7 @@ export interface RegistrationRequest {
 // USER
 
 export interface Feedback {
```
nik-dange, line 47: Event should also be a property here

nik-dange (changes requested):
I only took a look at the POST feedback route for now–let's try to fix that first and then move on to the fetching feedback route. 

There's a couple major things to refactor about the API request format, and I left a few comments detailing why they need to be changed. The biggest thing to remember is only passing in fields that are necessary from the frontend. I couldn't leave a comment since the code wasn't edited in this portion, but the event uuid should be passed in to the service method, and we should have a check to ensure the event exists/is valid before creating any feedback (users shouldn't be able to submit feedback for future events either; would love to hear your thoughts on older events as well). You can take a look at some of the attendance routes and services to see similar examples.

types/ApiRequests.ts
```diff
@@ -1,5 +1,5 @@
-import { FeedbackStatus, FeedbackType, SocialMediaType, UserAccessType } from './Enums';
 import { Uuid } from '.';
+import { FeedbackStatus, FeedbackType, SocialMediaType, UserAccessType } from './Enums';
```
nik-dange, line 2: remove feedback status from here when you remove it from the request

migrations/0039-add-feedback-event-field.ts
```diff
@@ -0,0 +1,55 @@
+import { MigrationInterface, QueryRunner, TableColumn, TableIndex, TableForeignKey } from 'typeorm';
+
+const TABLE_NAME = 'Feedback';
+const OLD_NAME = 'title';
+const NEW_NAME = 'source';
+
+export class AddFeedbackEventField1709112961573 implements MigrationInterface {
```
nik-dange, line 7: Nit: instead of `field`, use `column` in the name of the migration

api/validators/FeedbackControllerRequests.ts
```diff
@@ -1,27 +1,36 @@
...
+  @MinLength(20)
   description: string;
 
   @IsDefined()
   @IsValidFeedbackType()
   type: FeedbackType;
+
+  @IsDefined()
```
nik-dange, line 31: Remove status

repositories/EventRepository.ts
nik-dange: Bump

api/decorators/Validators.ts
```diff
@@ -145,7 +145,7 @@ export function IsValidFeedbackType(validationOptions?: ValidationOptions) {
 @ValidatorConstraint()
 class FeedbackStatusValidator implements ValidatorConstraintInterface {
   validate(feedbackStatus: FeedbackStatus): boolean {
-    return Object.values(FeedbackStatus).includes(feedbackStatus) && feedbackStatus !== FeedbackStatus.SUBMITTED;
+    return Object.values(FeedbackStatus).includes(feedbackStatus);
```
nik-dange, line 148: Not sure if we still need this, but run it and double check on your side

api/controllers/FeedbackController.ts
```diff
@@ -21,9 +23,10 @@ export class FeedbackController {
   }
 
   @Get()
-  async getFeedback(@AuthenticatedUser() user: UserModel): Promise<GetFeedbackResponse> {
+  async getFeedback(@QueryParams() options: FeedbackSearchOptions,
+    @AuthenticatedUser() user: UserModel): Promise<GetFeedbackResponse> {
     const canSeeAllFeedback = PermissionsService.canRespondToFeedback(user);
```
nik-dange, line 28: I feel as though this is a bit inconsistent (technically, a standard user would result in the `canSeeAllFeedback` variable being false. However, they can still view their own feedback, just not the feedback of others. `canRespondToFeedback` is a bit too generalizing for readability in my opinion, but I'm curious what you think makes sense here.

Also, a potential bug/inconsistency. As a standard user, using query params to see my own feedback doesn't work since it just default queries all of the user's feedback (I think?) :

image

models/FeedbackModel.ts
nik-dange: Honestly I feel like rather than/in addition to the query params structure we have a standard GET route for getting all feedback by event uuid, just seems a bit cleaner to me. Having that additional index may be fine with just the query params structure

tests/feedback.test.ts
```diff
@@ -22,30 +34,63 @@ afterAll(async () => {
 
 describe('feedback submission', () => {
   test('properly persists on successful submission', async () => {
+    const event = EventFactory.fake({
```
nik-dange, line 37: Can you see if there's any examples/any ways to have this as a global event, since it's created a couple different times in this test suite?

tests/Seeds.ts
```diff
@@ -669,6 +678,41 @@ async function seed(): Promise<void> {
     orderLimit: 10,
   });
 
+  // FEEDBACK SEEDING
+
+  // Event with multiple feedbacks: PAST_AI_WORKSHOP_1
+  const FEEDBACK_SAME_EVENT_1 = FeedbackFactory.fake({ user: USER_FEEDBACK_1,
+    event: PAST_AI_WORKSHOP_1,
+    description: 'Man this shit sucks',
```
nik-dange, line 686: no profanity pls :( Max would be sad

nik-dange (changes requested):
A few more small changes to make + some questions for you

types/ApiRequests.ts
yimmyj: There's a UpdateFeedbackStatusRequest in that file still, may need to keep it

api/controllers/FeedbackController.ts
yimmyj: It works fine on my end--maybe you didn't change the access token? For me, upon running the API, hitting the GET route as a standard user returns empty json, and all after submitting a few feedbacks, all of the user's feedbacks are returned properly regardless of the queryparams. 
Screenshot 2024-03-28 at 1 22 13 AM
Screenshot 2024-03-28 at 1 23 06 AM

tests/feedback.test.ts
yimmyj: Addressed with a function :)

api/controllers/FeedbackController.ts
yimmyj: I agree with canRespondToFeedback not being the most descriptive variable name, especially since the method itself simply checks if the user is an admin. This was in the previous version of the function and I didn't change it. I've refactored it to canQueryAllFeedbacks, let me know if that's better or if I should rename it to something else.

api/decorators/Validators.ts
yimmyj: Seems to work fine without it -- good catch

migrations/0039-add-feedback-event-field.ts
yimmyj: Addressed

types/ApiRequests.ts
```diff
@@ -45,9 +45,16 @@ export interface RegistrationRequest {
...
+
+  event: Uuid;
+
+  source: string;
+
   description: string;
+
   type: FeedbackType;
+
+  status: FeedbackStatus;
```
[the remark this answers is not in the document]
yimmyj: Addressed

api/validators/FeedbackControllerRequests.ts
yimmyj: Addressed

migrations/0039-add-feedback-event-column.ts
```diff
@@ -0,0 +1,55 @@
...
+        columnNames: ['event'],
+        referencedTableName: 'Events',
+        referencedColumnNames: ['uuid'],
+        onDelete: 'CASCADE',
+      }),
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    // Is this correct? I followed example in migration #29
```
nik-dange, line 41: Yeah, looks ok to me. you can remove this comment

Also I didn't know this but according to ChatGPT apparently creating the index on a column before the foreign key can speed up foreign key creation so good catch

migrations/0039-add-feedback-event-column.ts
```diff
@@ -0,0 +1,55 @@
...
+    await queryRunner.dropForeignKey(TABLE_NAME, new TableForeignKey({
+      columnNames: ['event'],
+      referencedTableName: 'Events',
+      referencedColumnNames: ['uuid'],
+      onDelete: 'CASCADE',
+    }));
+
+    await queryRunner.dropIndex('Feedback', 'feedback_by_event_index');
+    await queryRunner.dropColumn('Feedback', 'event');
```
nik-dange, line 50: Nit but if you declare table name above use that variable here

tests/Seeds.ts
```diff
@@ -179,6 +185,9 @@ async function seed(): Promise<void> {
   const USER_HIDDEN_RESUME = UserFactory.fake();
   const RESUME_2 = ResumeFactory.fake({ user: USER_HIDDEN_RESUME, isResumeVisible: false, url: RESUME_URL });
 
+  // const USER_FEEDBACK_1 = UserFactory.fake();
```
nik-dange, line 188: Nit: cleanup

nik-dange (commented):
Looks much better and cleaner to me, great work! One small nit I left a comment on. 

Also, I noticed that you can technically send the same feedback multiple times for a single event. IMO, we should disallow this because someone can spam this button (even if a rate limiter is in place, they can clutter the feedback for this event with random stuff). This is a big PR so I'll spend tomorrow rigorously testing it some more in case small things like this pop up.

services/PermissionsService.ts
```diff
@@ -9,14 +9,18 @@ export default class PermissionsService {
...
   }
 
   public static canSubmitFeedback(user: UserModel): boolean {
     return user.state === UserState.ACTIVE;
   }
```
SheepTester, line 17: btw, currently users stuck in the `PASSWORD_RESET` state can't submit feedback

```suggestion
  public static canSubmitFeedback(user: UserModel): boolean {
    return user.state === UserState.ACTIVE || user.state === UserState.PASSWORD_RESET;
  }
```

types/ApiResponses.ts
SheepTester, line 1: it looks like in EventModel.ts you've added a `feedback` array to events, but I don't see that change reflected here. I'm not a backend person so not sure if that's intentional

dowhep (approved):
LGTM 🚀 🚀

nik-dange (approved):
truly inspirational

services/FeedbackService.ts
```diff
@@ -17,28 +16,41 @@ export default class FeedbackService {
...
-    const addedFeedback = await this.transactions.readWrite(async (txn) => {
+    return this.transactions.readWrite(async (txn) => {
+      const event = await Repositories.event(txn).findByUuid(feedback.event);
+      if (!event) throw new NotFoundError('Event not found!');
+
+      const feedbackRepository = Repositories.feedback(txn);
+
+      const hasAlreadySubmittedFeedback = await feedbackRepository.hasUserSubmittedFeedback(user, event);
+      if (hasAlreadySubmittedFeedback) throw new UserError('You have already submitted feedback for this event!');
+
```
nik-dange, line 42: I just wrapped everything in this top level transaction so that if anything inside fails, the entire transaction fails. I know we're really only relying on reads and then one write inside, but I think this is a better pattern to follow in general

tests/feedback.test.ts
```diff
@@ -73,127 +114,353 @@ describe('feedback submission', () => {
   });
 
   test('admins can view feedback from any member', async () => {
+    const event = EventFactory.fake();
```
nik-dange, line 117: I removed your other function that was creating the event–we have `EventFactory.fake()` to do that for us.

PR #419, opened 2024-03-22
Fix null linkedEvent for some Order routes

nik-dange:
## Changes

- Added a left join in the get single order and get many orders query

# Type of Change

      expected)
      workflows, linting, etc.)
- [ ] Other: (Fill In) 

# Testing

- [x] locally.

# Checklist

# Screenshots

image
Screenshot 2024-03-21 at 10 44 26 PM

### DIFF
--- a/repositories/MerchOrderRepository.ts
+++ b/repositories/MerchOrderRepository.ts
@@ -11,7 +11,7 @@ import { MerchandiseItemModel } from '../models/MerchandiseItemModel';
 export class MerchOrderRepository extends BaseRepository<OrderModel> {
   /**
    * Gets a single order. Returns the order joined with ordered items,
-   * user, pickup event, the ordered items' merch options,
+   * user, pickup event, the pickup event's linked event, the ordered items' merch options,
    * and those merch options' merch items.
    *
    * This is the same set of joins that gets executed for OrderPickupEventRepository::findByUuid()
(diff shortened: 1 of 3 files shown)


nik-dange:
Don't merge I haven't added this query everywhere it needs to go yet

repositories/MerchOrderRepository.ts
```diff
@@ -48,7 +49,18 @@ export class MerchOrderRepository extends BaseRepository<OrderModel> {
    * Gets all orders for a given user. Returns the order joined with its pickup event and user.
    */
   public async getAllOrdersForUser(user: UserModel): Promise<OrderModel[]> {
-    return this.repository.find({ user });
```
nik-dange, line 51: @dowhep The comment says this query joins with pickup event and user but I was confused where it does this since the join isn't explicit. I rewrote the query to match the query on line 58 (`getAllOrderswithItemsForUser`), but I'm just wondering where the join was done in the first place

alexzhang1618, line 51: Curious if this was here before previously instead of the full query so we don't go over the max payload size when sending all orders for a user?

alexzhang1618 (approved):
LGTM 🚢

repositories/MerchOrderRepository.ts
nik-dange: Oh yea that's probably why, also I overlooked that this query is only orders and excludes order items. I guess the question now becomes–do we always want linkedEvent info coming back with every single order? I assume yes since it seems like the frontend wants to show the linked event for order history. I can remove the `orderItem` joins from the query, but if there's a way to bake the linkedEvent join into the `find` that would be cleanest.

nik-dange: Realized that the `find` does this magical join because the joins are specified in the OrderModel, as explicitly stated [here](https://github.com/acmucsd/membership-portal/blob/5c1bd1b076ffda633c35cfbb97cfe25df4411949/models/OrderModel.ts#L22), and with the `eager:true` option. I don't think we establish any relations between orders and linked events, so I adjusted this query at the repository layer instead

Issue #289, opened 2022-02-15
Update deprecated CI image


sumeet-bansal:
![image](https://user-images.githubusercontent.com/29763383/154136476-58aab3ba-1af7-455f-b20e-b56d7f691db7.png)

sumeet-bansal:
@shravanhariharan2 @michl1001

PR #420, opened 2024-03-26
Update Deprecated CI Docker Images

nik-dange:
# Info

Partially closes  **#289**

# Description

The Docker images that we use on CircleCI are deprecated, and we should switch to their next-gen images. More info here: https://circleci.com/docs/next-gen-migration-guide/, but this change should be more reliable/maintained/ideally speed up deployment times.

## Changes

- Switched Node images from circleci to cimg

# Type of Change

      expected)
- [x] Continuous Integration/DevOps Change (Related to deployment steps, continuous integration
      workflows, linting, etc.)
- [ ] Other: (Fill In) 

# Testing

- [ ] locally.
- [ ] with appropriate Postman routes. Screenshots are included below.

# Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have followed the style guidelines of this project.
- [ ] My changes produce no new warnings.

# Screenshots


nik-dange:
I was able to find a matching next-gen image for our Node v14.17.6 ones from here (https://circleci.com/developer/images/image/cimg/node), but I was unable to find a 1:1 match for our Postgres image (which uses 11.5), see https://circleci.com/developer/images/image/cimg/postgres. Thoughts on moving forward from this? Using the closest minor version vs. overall upgrading our Postgres version? Would need to look into going about the latter but it might not be a bad idea

shravanhariharan2 (approved):
Worth looking into upgrading both our Node and Postgres versions at some point, though I don't think it needs to be scoped into this PR. The current Node LTS is [v20.12.X](https://nodejs.org/en), so we're fairly far behind and are at some level of risk to vulnerabilities that could be patched in more recent updates - and the same goes for Postgres.

If the deprecated images is a blocker (i.e. CCI is going to remove the images entirely from their old-gen repository), then it's worth bumping Postgres to a supported minor version on 11.X. Then, we can later look into upgrading the major version, using the testing instance and our unit tests to heavily validate that nothing's broken.

Approving this PR to unblock, but I'd make sure this behaves as intended when deploying onto the testing instance first before merging to master.

PR #426, opened 2024-04-04
Update isUnusedAttendanceCode to allow any non-conflicting Event

yimmyj:


dowhep (approved):
LGTM 🚀

nik-dange (changes requested):
make sure to change the package version + migration number once again, since we merged in Eric's PR before yours

nik-dange (changes requested):
Quick thing, did you test checking in to an event when there are multiple events with the same checkin code? I misremembered that the checkin route uses find by checkin code, but double check this functionality either way

services/AttendanceService.ts
```diff
@@ -59,6 +59,7 @@ export default class AttendanceService {
   }
 
   private validateEventToAttend(event: EventModel) {
+    console.log(event);
```
echang594, line 62: Delete this

echang594 (changes requested):

services/AttendanceService.ts
echang594: Other than that, LGTM

repositories/EventRepository.ts
```diff
@@ -33,16 +33,51 @@ export class EventRepository extends BaseRepository<EventModel> {
   }
 
   public async findByAttendanceCode(attendanceCode: string): Promise<EventModel> {
-    return this.repository.findOne({ attendanceCode });
+    const allEvents = await this.repository.find({ attendanceCode });
```
nik-dange, line 36: nit: rename to `matchingEvents`

repositories/EventRepository.ts
```diff
@@ -33,16 +33,51 @@ export class EventRepository extends BaseRepository<EventModel> {
...
+
+    // Find all events with the given attendance code
+    const eligible = allEvents.filter((event) => !event.isTooEarlyToAttendEvent() && !event.isTooLateToAttendEvent());
```
nik-dange, line 39: nit: rename to `validEvents` or something along those lines

tests/event.test.ts
```diff
@@ -126,6 +126,96 @@ describe('event creation', () => {
...
+    };
+
+    const createEventRequest: CreateEventRequest = {
+      event,
+    };
+
+    const eventController = ControllerFactory.event(conn);
+    await eventController.createEvent(createEventRequest, admin);
+
+    event = {
```
nik-dange, line 200: Try to use the built-in methods to generate fake events rather than hardcoding them in like this (ie EventFactory's fake function)

tests/event.test.ts
```diff
@@ -126,6 +126,96 @@ describe('event creation', () => {
...
+  test('test non-conflicting event creation with re-used past attendance code', async () => {
+    const conn = await DatabaseConnection.get();
+    const admin = UserFactory.fake({ accessType: UserAccessType.ADMIN });
+    const user = UserFactory.fake();
+
+    await new PortalState()
+      .createUsers(admin, user)
+      .write();
+
+    const event = {
```
nik-dange, line 139: See comment regarding using fake function

nik-dange (changes requested):
Couple small nits. I think we'll need to stress test this with a few different situations to make sure it works as intended just to make sure there's no strange edge cases. Can you validate this w/ Postman testing?

repositories/EventRepository.ts
```diff
@@ -33,16 +33,51 @@ export class EventRepository extends BaseRepository<EventModel> {
...
   public async findByAttendanceCode(attendanceCode: string): Promise<EventModel> {
-    return this.repository.findOne({ attendanceCode });
+    const matchingEvents = await this.repository.find({ attendanceCode });
+
+    // Find all events with the given attendance code
+    const validEvents = matchingEvents.filter((event) => !event.isTooEarlyToAttendEvent() && !event.isTooLateToAttendEvent());
+
+    // If there are eligible events, return the first one
+    if (validEvents.length > 0) {
+      return validEvents[0];
```
nik-dange, line 43: What defines the first event? Is it sorted by start time?

repositories/EventRepository.ts
```diff
@@ -33,16 +33,51 @@ export class EventRepository extends BaseRepository<EventModel> {
...
+      const timeDifference = Math.abs(eventStartTime.getTime() - currentTime.getTime());
+
+      // Update closest event if necessary
+      if (timeDifference < closestTimeDifference) {
+        closestEvent = event;
+        closestTimeDifference = timeDifference;
+      }
+    });
+
+    return closestEvent;
```
nik-dange, line 62: Just curious, what's the edge case that warrants this code? Shouldn't all potentially valid events be caught in the above code? I know we're validating it one more time in the AttendanceService, just curious what the reasoning for adding it was

nik-dange: And side note, if there is a use case that warrants this code, does it make more sense to use validEvents instead of matchingEvents?

nik-dange (commented):
Left two questions

repositories/EventRepository.ts
yimmyj: The use case is the small pop-up that announces whether an event "hasn't occurred yet / has already occurred / doesn't exist" when the user enters an attendance code that isn't matched to an event currently going on, in AttendanceService's validateEventToAttend method.

 The way it used to work when attendance codes were unique is that the validate method would simply check if the one event with the matching attendance code hasn't started, has already ended, or doesn't exist in the schema. 

However with repeated attendance codes, it is possible for events sharing the same code to be simultaneously "too early" AND "too late," so I chose to return the closest event to the current date as a "guess" as to what the user meant to check in to when they entered a code. For example, if code "ABC" was used both for an event a year ago and two days in the future, I'm assuming that the user meant to check in to the one two days in the future, and will show the "hasn't occurred yet" message. Let me know if this implementation needs more consideration.

repositories/EventRepository.ts
yimmyj: In theory there should never be more than one eligible event; that would mean that there are two events occurring at the same time sharing an attendance code which wouldn't make any sense. This is handled by isAvailableAttendanceCode and tested in the attached postman screenshots. 

However, since the check-in period is defined as 30 minutes before and after an event, two events happening one after another with the same attendance code could cause an issue. 

The solution to this could be to simply disallow this case, which involves changing isAvailableAttendanceCode to check for conflicting events in this "check-in period" rather than the duration of the event itself. Let me know your thoughts on this change!

dowhep: I agree with the "check-in period" approach. However, realistically, we can allow more padding room cases like this, such as not allowing events with the same checkin code 3 days before and after the event. This would avoid the edge case of having a time precisely in the "check-in period"s of two close events with the same checkin code.

repositories/EventRepository.ts
dowhep: I believe we shouldn't have the "hasn't occurred yet" message at all as it give away the checkin code for a future event, nullifying the secrecy of the checkin code.

dowhep (changes requested):
Just my 2 cents

Issue #441, opened 2024-05-11
Email ics file for pickup events


echang594:
- Attach calendar info to the order confirmation email

Original issue here:
https://github.com/acmucsd/membership-portal-ui-v2/issues/249

PR #443, opened 2024-05-15
Email ics file for pickup events

echang594:
# Info

Closes **#441**.

# Description

What changes did you make? List all distinct problems that this PR addresses. Explain any relevant
motivation or context.

- Added calendar info for initial order confirmation and on rescheduled orders using ics package

## Changes

# Type of Change

      expected)
      workflows, linting, etc.)
- [ ] Other: (Fill In) 

# Testing

- [x] locally.

# Checklist

# Screenshots

Screenshot 2024-05-15 at 2 54 15 PM
Screenshot 2024-05-15 at 2 54 31 PM

### DIFF
--- a/services/EmailService.ts
+++ b/services/EmailService.ts
@@ -1,4 +1,5 @@
 import { MailService, MailDataRequired } from '@sendgrid/mail';
+import { createEvent } from 'ics';
 import * as ejs from 'ejs';
 import * as fs from 'fs';
 import * as path from 'path';
@@ -75,7 +76,8 @@ export default class EmailService {
     }
   }
 
-  public async sendOrderConfirmation(email: string, firstName: string, order: OrderInfo): Promise<void> {
+  public async sendOrderConfirmation(email: string, firstName: string, order: OrderInfo,
+    calendarInfo: OrderPickupEventCalendarInfo): Promise<void> {
     try {
       const data = {
         to: email,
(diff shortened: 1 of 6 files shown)


services/MerchStoreService.ts
```diff
@@ -903,6 +905,15 @@ export default class MerchStoreService {
     };
   }
 
+  private static toEventCalendarInfo(pickupEvent: OrderPickupEventModel): OrderPickupEventCalendarInfo {
+    return {
+      start: pickupEvent.start.getTime(),
```
nik-dange, line 910: What are your thoughts on adding location as a field to OrderPickupEvents?

services/EmailService.ts
```diff
@@ -88,6 +90,14 @@ export default class EmailService {
           pickupEvent: order.pickupEvent,
           link: `${Config.client}/store/orders`,
         }),
+        attachments: [
+          {
+            content: EmailService.createCalendarFile(calendarInfo),
+            filename: 'invite.ics',
+            type: 'text/calendar',
```
nik-dange, line 97: Seeking to learn: Where does this type come from? Is this just how ics files are meant to be defined (is this a Sendgrid/EJS/ics thing?)

nik-dange (changes requested):
Looks great to me, a much appreciated feature! What are your thoughts on adding location as a column to `OrderPickupEventModel`? We now support linking regular events to pickup events, so maybe on the frontend if it's linked they can just fill in that location but if it's unlinked the user can input a location. This way we can show it in the calendar invite as well.

services/EmailService.ts
echang594: This is the mime type, it is used to identify the file format. For example we use "application/json" to send out the responses from the backend in json format

services/MerchStoreService.ts
echang594: After discussing, we've decided to add it and copy over the location from the linkedEvent if none is provided. One potential issue is that if the linkedEvent location updates, this field will not, however, we do not think this will be a big problem as the user will be notified through discord.

PR #452, opened 2024-10-10
Moving most public types to API layer

### LINKED ISSUE #326: Move mapping of private to public types from service layer to API layer
As part of our TypeScript rewrite a few years ago, the codebase introduced the idea of "private" types and "public" types - private types being types used exclusively within the codebase and not exposed to any client, and public types being types exposed to a client (most commonly via. an API response). For example, `FeedbackModel` would be a private type, and `PublicFeedback` would be its corresponding public type. We have this distinction of private vs. public to give the developer control over what data should be visible to the client via. API response, and in the future, to have public data types exclusively be exported via. a types package for the frontend codebase to use.

During our TypeScript rewrite, we had the mapping of private types to public types located in the service layer of the codebase (e.g., [here in EventService](https://github.com/acmucsd/membership-portal/blob/master/services/EventService.ts#L25)). This issue proposes to move this mapping from the service layer to the API layer. This pattern is more ideal since the public type is a concept specific to the API layer (since it defines the structure of the responses of the API to the client), so it should be used only in the API layer. An initial effort has been made to have new services keep the mapping away from it and within the API layer (e.g., [here in ResumeService](https://github.com/acmucsd/membership-portal/blob/master/services/ResumeService.ts#L22)), but since we do still have remnants of services returning the public type (as linked earlier), the mix of patterns can be confusing and something that should be solidified to reduce future confusion.

jennymar:
# Info

Closes #326 

# Description

What changes did you make? List all distinct problems that this PR addresses. Explain any relevant
motivation or context.

Changes were made to almost all the controller classes in the API layer, as well as their corresponding service classes in the service layer. The mapping of private to public types was moved from the service layer to the API layer. 

## Changes

**API Layer**
- `AdminController.ts`
- `AttendanceController.ts`
- `EventController.ts`
- `FeedbackController.ts`
- `LeaderboardController.ts`
- `MerchStoreController.ts`
- `UserController.ts`

**Service Layer**
- `AttendanceService.ts`
- `EventService.ts`
- `FeedbackService.ts`
- `MerchStoreService.ts`
- `UserAccountService.ts`

# Type of Change

- [x] Patch (non-breaking change/bugfix)
      expected)
      workflows, linting, etc.)
- [ ] Other: (Fill In) 

# Testing

- [ ] locally.
- [ ] with appropriate Postman routes. Screenshots are included below.

# Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have followed the style guidelines of this project.
- [ ] My changes produce no new warnings.

# Screenshots

### DIFF
--- a/api/controllers/AdminController.ts
+++ b/api/controllers/AdminController.ts
@@ -97,6 +97,6 @@ export class AdminController {
   async getAllUsersWithAccessLevels(@AuthenticatedUser() user: UserModel): Promise<GetAllUserAccessLevelsResponse> {
     if (!PermissionsService.canSeeAllUserAccessLevels(user)) throw new ForbiddenError();
     const users = await this.userAccountService.getAllFullUserProfiles();
-    return { error: null, users };
+    return { error: null, users: users.map((user) => user.getFullUserProfile()) };
   }
 }
--- a/api/controllers/AttendanceController.ts
+++ b/api/controllers/AttendanceController.ts
@@ -26,14 +26,14 @@ export class AttendanceController {
     @AuthenticatedUser() user: UserModel): Promise<GetAttendancesForEventResponse> {
     if (!PermissionsService.canSeeEventAttendances(user)) throw new ForbiddenError();
     const attendances = await this.attendanceService.getAttendancesForEvent(params.uuid);
-    return { error: null, attendances };
+    return { error: null, attendances: attendances.map((attendance) => attendance.getPublicAttendance())};
   }
 
   @UseBefore(UserAuthentication)
   @Get()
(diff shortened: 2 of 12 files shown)


maxwn04:
Are we keeping the user profile routes in UserAccountService the same? I think we can still add resumes/social media to the UserModel. Maybe we can change the query as well but that might affect other routes.
https://github.com/acmucsd/membership-portal/blob/2d73efb1f194c76dd21f13f184f828a10758fdec/services/UserAccountService.ts#L195

maxwn04:
Are we keeping the user profile routes in UserAccountService the same? I think we can still add resumes/social media to the UserModel. Maybe we can change the query as well but that might affect other routes.
https://github.com/acmucsd/membership-portal/blob/2d73efb1f194c76dd21f13f184f828a10758fdec/services/UserAccountService.ts#L195

